PR #6, opened 2016-04-22
Attach to property

orangy:
This proposal enables instances to which properties are delegated using `by` keyword to receive "attach" call so that they can do some useful work before getValue/setValue calls will start happening.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ The proposals themselves are colloquially referred to as KEEPs.
 | Coroutines | repo: [kotlin-coroutines](https://github.com/Kotlin/kotlin-coroutines) | [Issues](https://github.com/Kotlin/kotlin-coroutines/issues)
 | Type aliases | [type-aliases.md](proposals/type-aliases.md) | [Issue #4](https://github.com/Kotlin/KEEP/issues/4)
 | Bound callable references | [bound-callable-references.md](proposals/bound-callable-references.md) | [Issue #5](https://github.com/Kotlin/KEEP/issues/5)
+| Attach to property | [attach-to-property.md](proposals/attach-to-property.md) |
 
 ## How to give feedback
 
--- a/proposals/attach-to-property.md
+++ b/proposals/attach-to-property.md
@@ -1,0 +1,142 @@
+# Attach to delegate property object
+
+* **Type**: Design proposal
+* **Author**: Ilya Ryzhenkov
+* **Contributors**:
+* **Status**:
+* **Prototype**: Was implemented and then disabled
+
+## Summary
+
+Recognize special operator function available for delegate object and call it right after it has been stored
+into a field for delegation.
+
+## Glossary
+
+- "delegate instance": instance on which property's get/set accessors execute respective getValue/setValue calls.
+- "host instance": instance containing delegated properties as members, or being extended with extension property.
+- "attachTo": placeholder for the name of the operator function
+- "constructor" means either primary constructor, or each of secondary constructor if there is no primary constructor
+
+## Motivation / use cases
+
+Make delegate instance aware of host instance and property it was used with.
+
+- Register host instance in external table of managed entities, e.g. database schema, component container.
+- Register delegate properties in host instance with callback for "metadata", e.g. for database columns.
+- Call back into host instance with additional information, e.g. property metadata, delegate instance.
+  Can be used to communicate delegate instance back to the host instance for later use, by declaring member
+  extension `attachTo` function with receiver of delegate type.
+- Execute custom code before first access to property, e.g. `async` property that is being calculated
+  in the background, and blocks as needed when trying to use the value. With coroutines, we can suspend instead
+  of blocking.
+
+## Description
+
+Client code:
+```
+class HostType {
+   val name by DelegateType()
+}
+```
+
+Attach operator:
+```
+operator fun DelegateType.attachTo(instance: HostType, property: KProperty<*>) : Unit`
+```
+
+- No extra instances are created, all information is already available
+
+## Alternative names
+
+- bindToInstance
+- delegatedFrom
+
+## Implementation
+
+### Syntax
+
+No new syntax is required.
+
+### Resolution
+
+- resolving the `attachTo` function happens similar to resolving `getValue`/`setValue` functions.
+- if there are more than one `attachTo` functions available, overload resolution selects best candidate.
+- `attachTo` that is not accessible is ignored.
+- `attachTo` can be inline.
+- `attachTo` can be generic, inference happens normally based on the type of host instance.
+- Member extension `attachTo` on matching delegate type is preferred.
+- No implicit `attachTo` operator from Java
+
+### Diagnostics
+
+Declaration site:
+- `attachTo` should have Unit return type
+- `attachTo` should have two parameters, second one should have `KProperty<*>` type
+
+Use site (`by` keyword):
+-
+
+### Code generation (JVM)
+
+At the end of the primary constructor, or each of secondary constructors:
+For each property with discovered `attachTo` operator:
+Call to `attachTo` is generated passing `this` and property metadata.
+
+Member `attachTo`
+```
+name$delegate.attachTo(this, $$delegatedProperties[0])
+```
+
+Extension `attachTo`
+```
+FileKt.attachTo(name$delegate, this, $$delegatedProperties[0])
+```
+
+Member Extension `attachTo`
+```
+this.attachTo(name$delegate, this, $$delegatedProperties[0])
+```
+
+Notes:
+- inline `attachTo` function is inlined normally in place of call
+
+### Code generation (JS)
+
+Similar to JVM.
+
+### Debug information
+
+- Breakpoint on a property should be triggered on `attachTo` call.
+- Step over on a property should go to next property with `attachTo` generated
+- Step into should go into respective `attachTo` method
+- Step over at the end of constructor should jump to first property with `attachTo` generated
+- Stacktrace navigation should show end of constructor as a place where `attachTo` was called
+
+## Questions and Notes
+
+- No special exception safety is required here. If `attachTo` throws, it aborts instance creation and propagates
+  further out of constructor.
+- Call happens at the end of a constructor, escaped `this` is not a big problem there since instance is initialised,
+  but still can be cause for "virtual call in constructor" issue if further inherited types are yet to be constructed.
+  However, if host instance is stored in a location accessible externally by other cores, it may also be problematic.
+- Stacktrace navigation is ambiguous at the point of `attachTo` call, it can be calling constructor
+  or respective property. Consider creating private function `attachTo$properties` that is called at the end of
+  constructor, which in turn calls all required `attachTo` functions.
+- Accessible operator function `attachTo` available on delegate instance cannot be called with host instance,
+  but since entire call is optional, should it be warning, error or just silently ignoring it?
+  Inaccessible `attachTo` functions are silently ignored without any diagnostics, which is not good as well.
+  Also it cannot be invokable property, object or functional value, because those cannot be operators.
+
+
+## Future evolution
+
+Enable `attachTo` variant for interface delegation:
+```
+class Foo : List<String> by mutableListOf() {
+    private lateinit var collection : MutableList<String>
+    private operator fun MutableList<String>.attachTo(instance: Foo, type: KClass<*>) {
+        collection = this
+    }
+}
+```
--- a/proposals/attach-to-property.md
+++ b/proposals/attach-to-property.md
@@ -1,22 +1,23 @@
-# Attach to delegate property object
+# Interception of delegated property binding
 
 * **Type**: Design proposal
 * **Author**: Ilya Ryzhenkov
-* **Contributors**:
+* **Contributors**: Zalim Bashorov
 * **Status**:
 * **Prototype**: Was implemented and then disabled
 
 ## Summary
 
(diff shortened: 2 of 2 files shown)


cypressious:
Part of the proposed functionality can already be accomplished using extension factory functions, e.g.:

```kotlin
class Foo {
    val delegated by factory()

    fun bindToLifecycle(listener: Listener) { ... }
}

fun Foo.factory() = Delegate(this)

class Delegate(foo: Foo) : Listener {
    init {
        foo.bindToLifecycle(this)
    }

    operator fun getValue(...)
}
```

Not sure if this is an argument against it. Of course you can't access the `KProperty` instance.

orangy:
@cypressious yes, but this doesn't work if you want to delegate to already existing instance, e.g. passed to constructor or otherwise available. You can wrap it with `attach(instance)` call which will return instance and execute attachment code, but it's more and more boilerplate. And, often, you still need `KProperty`.

cypressious:
I think in most cases, you will never delegate two properties to the same delegate, but maybe I'm wrong.

Anyway, I think making this mechanism available to extension functions is the real benefit here because you can delegate to a library class you don't control but can still handle stuff like lifecycle.

README.md
```diff
@@ -11,6 +11,7 @@ The proposals themselves are colloquially referred to as KEEPs.
 | Coroutines | repo: [kotlin-coroutines](https://github.com/Kotlin/kotlin-coroutines) | [Issues](https://github.com/Kotlin/kotlin-coroutines/issues)
 | Type aliases | [type-aliases.md](proposals/type-aliases.md) | [Issue #4](https://github.com/Kotlin/KEEP/issues/4)
 | Bound callable references | [bound-callable-references.md](proposals/bound-callable-references.md) | [Issue #5](https://github.com/Kotlin/KEEP/issues/5)
+| Attach to property | [attach-to-property.md](proposals/attach-to-property.md) |
```
abreslav: We'd still need an issue for discussion

proposals/attach-to-property.md
```diff
@@ -0,0 +1,142 @@
+# Attach to delegate property object
```
abreslav: I think we should work some more on the title. Options off the top of my head:
- Attachment handlers for delegated properties
- Creation handlers for property delegates

proposals/attach-to-property.md
```diff
@@ -0,0 +1,142 @@
...
+   val name by DelegateType()
+}
+```
+
+Attach operator:
+```
+operator fun DelegateType.attachTo(instance: HostType, property: KProperty<*>) : Unit`
+```
+
+- No extra instances are created, all information is already available
```
abreslav: We could allow returning a (potentially different) instance of `DelegateType`, to facilitate customizing delegate object upon creation

proposals/attach-to-property.md
```diff
@@ -0,0 +1,142 @@
...
+## Implementation
+
+### Syntax
+
+No new syntax is required.
+
+### Resolution
+
+- resolving the `attachTo` function happens similar to resolving `getValue`/`setValue` functions.
+- if there are more than one `attachTo` functions available, overload resolution selects best candidate.
```
abreslav: wrt to type of `instance` being the _default type_ of the containing class.

What if the property is a not a member? We must say explicitly that `null` as passed for the first parameter

proposals/attach-to-property.md
```diff
@@ -0,0 +1,142 @@
...
+- `attachTo` that is not accessible is ignored.
+- `attachTo` can be inline.
+- `attachTo` can be generic, inference happens normally based on the type of host instance.
+- Member extension `attachTo` on matching delegate type is preferred.
```
abreslav: This rule is surprising and needs justification/clarification

proposals/attach-to-property.md
```diff
@@ -0,0 +1,142 @@
...
+- No implicit `attachTo` operator from Java
```
abreslav: I was able to decrypt what this phrase implies, but most readers won't

proposals/attach-to-property.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Diagnostics
+
+Declaration site:
+- `attachTo` should have Unit return type
+- `attachTo` should have two parameters, second one should have `KProperty<*>` type
```
abreslav: No default parameters, no varargs

proposals/attach-to-property.md
```diff
@@ -0,0 +1,142 @@
...
+- inline `attachTo` function is inlined normally in place of call
+
+### Code generation (JS)
+
+Similar to JVM.
+
+### Debug information
+
+- Breakpoint on a property should be triggered on `attachTo` call.
+- Step over on a property should go to next property with `attachTo` generated
```
abreslav: What if the next property has no `attachTo`, but has a non-trivial initializer that the debugger has to stop on, or next follows an `init` block? I believe, this like should be removed.

proposals/attach-to-property.md
```diff
@@ -0,0 +1,142 @@
...
+Notes:
+- inline `attachTo` function is inlined normally in place of call
+
+### Code generation (JS)
+
+Similar to JVM.
+
+### Debug information
+
+- Breakpoint on a property should be triggered on `attachTo` call.
```
abreslav: It will be triggered on delegate expression first

proposals/attach-to-property.md
```diff
@@ -0,0 +1,142 @@
...
+Declaration site:
+- `attachTo` should have Unit return type
+- `attachTo` should have two parameters, second one should have `KProperty<*>` type
+
+Use site (`by` keyword):
+-
+
+### Code generation (JVM)
+
+At the end of the primary constructor, or each of secondary constructors:
```
abreslav: > or each of secondary constructors
Is this for the case when no primary constructor is defined?

proposals/attach-to-property.md
```diff
@@ -0,0 +1,142 @@
...
+### Code generation (JS)
+
+Similar to JVM.
+
+### Debug information
+
+- Breakpoint on a property should be triggered on `attachTo` call.
+- Step over on a property should go to next property with `attachTo` generated
+- Step into should go into respective `attachTo` method
+- Step over at the end of constructor should jump to first property with `attachTo` generated
```
abreslav: Ok, now I understand what you assumed above. It means that at some properties we will stop twice: first for initializer/delegate, then for `attachTo`. What a breakpoint should mean is not entirely clear

proposals/attach-to-property.md
```diff
@@ -0,0 +1,142 @@
...
+- Stacktrace navigation should show end of constructor as a place where `attachTo` was called
```
abreslav: Why not the property declaration?

proposals/attach-to-property.md
```diff
@@ -0,0 +1,142 @@
...
+## Questions and Notes
+
+- No special exception safety is required here. If `attachTo` throws, it aborts instance creation and propagates
+  further out of constructor.
+- Call happens at the end of a constructor, escaped `this` is not a big problem there since instance is initialised,
+  but still can be cause for "virtual call in constructor" issue if further inherited types are yet to be constructed.
+  However, if host instance is stored in a location accessible externally by other cores, it may also be problematic.
+- Stacktrace navigation is ambiguous at the point of `attachTo` call, it can be calling constructor
+  or respective property. Consider creating private function `attachTo$properties` that is called at the end of
+  constructor, which in turn calls all required `attachTo` functions.
```
abreslav: How does it matter, if this function is not in the source code anyways?

proposals/attach-to-property.md
```diff
@@ -0,0 +1,142 @@
...
+- Accessible operator function `attachTo` available on delegate instance cannot be called with host instance,
```
abreslav: Why?

abreslav:
Added some comments to the diff

README.md
orangy: My impression was that we discuss PRs in PRs, which is also an issue, no?

proposals/attach-to-property.md
orangy: I was thinking about it, since we have "unused" return type of the function, but thinking about various consequences I discarded this idea. I will add some text about why this is not an option.

proposals/attach-to-property.md
orangy: Ok

proposals/attach-to-property.md
orangy: Looks like this is already the case, at least for getValue/setValue, it works as expected. So this line can be removed altogether.

proposals/attach-to-property.md
orangy: Will elaborate.

proposals/attach-to-property.md
orangy: Exactly. It was expanded in "glossary", which was copied from this place. So, either we need to say just "constructor" here, or expand it in every place.

bashor:
Can we slightly rethink "delegation by" semantic?

Let's say that `val foo by expression` is equivalent to:

private val foo$delegate = expression.getDelegate(this, ::foo)
val foo: T
	get() = foo$delegate.getValue(this, ::foo)



And add to stdlib:
```
operator inline fun <D> D.getDelegate(instance: Any?, property: KProperty<*>): D = this
```

IMHO it slightly more elegant, and allow to avoid mutable state in delegate class.

proposals/attach-to-property.md
orangy: Should it be triggered on both? Debug experience is moot here, so I'd appreciate any help in formalising it.

proposals/attach-to-property.md
orangy: Since these calls are generated at the end of constructor, there is nothing that follows these calls. Until another feature generates something at the end of constructor, of course. Alternative would be generate `attachTo` calls next to PUTFIELD, but that means attachment handlers will see incomplete host instance.

proposals/attach-to-property.md
orangy: Attachment handler call relates to two things: 
- constructor, where it is being called according to stacktrace (in case of multiple secondary ctors)
- property, which is associated with particular handler being called

hence, ambiguity.

proposals/attach-to-property.md
orangy: Incomplete sentence, will rephrase. Missing "when cannot be called", e.g. when `attachTo` is accessible, receiver matches, but type of `instance` parameter doesn't match.

orangy:
@bashor that would eliminate possibility to optimise delegate's storage in the future. We were considering optimising delegate fields when possible, e.g. when delegate is an `object` or when it's pretty primitive expression like another property value or constant value. Since we cannot reason about what `getDelegate` would return, that pretty much means we have to have a field for a property, even when it's actually not needed.

bashor:
Good point! 👍
For the record:
With `getDelegate` we still can optimize some cases:
* all your cases when it's our simple version
* when return type of getDelegate is `object`

orangy:
@bashor yep, I'll add it to alternatives.

abreslav:
This YouTrack issue should be mentioned somewhere in the proposal: [KT-7925](KT-7925 Notify property delegate if it's attached to an object) *Notify property delegate if it's attached to an object*

Groostav:
Can I ask what the expected criteria for inheritance in the implementation of `provideDelegate` are?

I've got:
- a delegate factory `prop(): PropInterface` (for idiomatic use `val whatever by prop()`), 
- the type `PropInterface: ReadOnlyProperty` 
- a `PropImpl: PropInterface`, 
- a method `provideDelegate` method on `PropImpl` 

this setup results in `provideDelegate` not being called --I need to write an extension function for `PropInterface`, in order to get it called. It would seem there is no vtable lookup on the resulting object, only a compiler-inserted (static) check for the `provideDelegate` method.

Is this by design? Should it not be documented in the docs: https://kotlinlang.org/docs/reference/whatsnew11.html#interception-of-delegated-property-binding

abreslav:
@Groostav A vtable lookup can only happen when some statically known type has a corresponding member declared. In Kotlin all conventions are resolved statically, and your static types don't have `provideDelegate`, so the behavior you describe looks legitimate.

I don't think this needs documenting, because this is the way all convention work: if the operator function is not present at compile time, it can't be called at runtime. The only difference here is that the compiler does not complain about the absent operator function, but simply skips the call to `provideDelegate` instead. This is unfortunate, but I see no way around it ATM

Groostav:
> the only difference here is that the compiler does not complain about the absent operator function

_nor does it complain about an extra operator function_ --that is, it doesn't complain when you have too few `provideDelegate` implementations and it doesnt complain when you have too many `provideDelegate` functions. This is the kind of wiggle-room that madness is born from. 

Similarly visibility is important: if the `provideDelegate` operator-function is `internal` or `private` it is not called, but no warnings about its declaration are given. 

Perhaps then an IDEA "suspicious" check: If IDEA can find a `provideDelegate` function on the inheritence tree for the returned property type, but not one that is statically available, it would give you a suspiciousness warning. In my particular case it could be even more restrictive, if the inferred return type is more specific than the declared return type, and the inferred return type has a `provideDelegate` method but the declared one does not, then IDEA would raise a suspicion: 

```
fun prop(): PropInterface = PropImpl()
```

results in 
> warning: the type of the returned instance has a `provideDelegate` method but the declared return type does not. 

---

ps: I love that IDEA is suspicious of my code.

elizarov:
@orangy Shall we close it now?

orangy:
Obsolete

PR #7, opened 2016-04-22
Update proposal for type aliases

dnpetrov:
- type aliases and declaration conflicts
- nested types & overrides
- minor fixes and clarifications
- type projections in arguments of generic type aliases
- metadata

### DIFF
--- a/proposals/type-aliases.md
+++ b/proposals/type-aliases.md
@@ -2,39 +2,71 @@
 
 * **Type**: Design proposal
 * **Author**: Dmitry Petrov
-* **Contributors**: Andrey Breslav, Stanislav Erokhin
+* **Contributors**: Andrey Breslav, Stanislav Erokhin, Vladimir Reshetnikov
 * **Status**: Under consideration
 * **Prototype**: In progress
-
-## Feedback 
 
 Discussion of this proposal is held in [this issue](https://github.com/Kotlin/KEEP/issues/4).
 
 ## Use cases
 
-* Avoid repetition of types in source code without a cost of introducing extra classes.
-    * Function types
-    * Collections (and other complex generics)
-    * Nested classes
-    * ...
+Type aliases provide alternative names for existing types, for example:
+* Function types
+```
+typealias MyHandler = (Int, String, Any) -> Unit
 
-## Restrictions
+typealias HtmlBuilderAction = HtmlBuilder.() -> Unit
 
-Java has no notion of "type aliases" and can't see them in class member signatures.
+typealias Predicate<T> = (T) -> Boolean
+```
+* Collections (and other generics)
+```
+typealias NodeSet = Set<Network.Node>
 
-For the same reason we can't enforce "newtype"
-(type alias with limited assignment-compatibility with its underlying type)
-in Java.
-However, this would be possible with value types.
+typealias FilesTable<K> = MutableMap<K, MutableList<File>>
+```
+* Nested classes
+```
+class Outer {
+  class Nested {
+    inner class Inner
+  }
+}
+
+typealias Something = Outer.Nested.Inner
+```
+and so on.
+
+Type aliases do not introduce new types.
+Instead, they are fully expanded, and are equivalent to the corresponding underlying types.
+
+Use **value classes** to define new types (which are not assignment-compatible with the corresponding underlying types,
+but do not introduce overhead related to additional heap allocations),
+Value classes will be supported closer to [Project Valhalla](http://openjdk.java.net/projects/valhalla/) release.
+
+**NB** Java has no concept of "type aliases" and can't see them in class member signatures.
+
+## Type aliases and tooling
+
+IDE and compiler should be fully aware of type aliases:
+* Diagnostic messages
+* Descriptor rendering in IDE (completion, structure view, etc)
+* ...
 
 ## Type alias declarations
 
 Type aliases are declared using `typealias` keyword:
 ```
 typeAlias
-    : modifiers 'typealias' SimpleName (typeParameters typeConstraints)? '='
-        type
+    : modifiers 'typealias' SimpleName (typeParameters)? '=' type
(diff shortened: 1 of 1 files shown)


proposals/type-aliases.md
```diff
@@ -2,27 +2,54 @@
...
 
-## Feedback 
+## Use cases
+
+Type aliases provide alternative names for existing types, for example:
+* Function types
+```
+typealias MyHandler = (Int, String, Any) -> Unit
 
-Discussion of this proposal is held in [this issue](https://github.com/Kotlin/KEEP/issues/4).
```
abreslav: Why is this removed? The link to the relevant issue is needed to direct the feedback there.

proposals/type-aliases.md
```diff
@@ -2,27 +2,54 @@
...
+* Diagnostic messages
+* Descriptor rendering in IDE (completion, structure view, etc)
+* ...
 
 ## Restrictions
 
 Java has no notion of "type aliases" and can't see them in class member signatures.
 
-For the same reason we can't enforce "newtype"
+For the same reason we can't enforce "new type"
```
abreslav: the public discussion has show that this is reason is rather weak, and provokes unnecessary discussion. We can use proper reasoning to justify lack of support for "new type":

1. Assignment-compatible type aliases will definitely be supported, because they are needed in popular use cases (naming function types and long generic types)
2. Value classes will be supported (at some point, closer to Valhalla release), because they enable many important use cases
3. There's no point in having two ways of expressing assignment-incompatible type aliases (newtype and value classes), so we don't want to support newtype.

proposals/type-aliases.md
```diff
@@ -32,7 +59,7 @@ However, this would be possible with value types.
 Type aliases are declared using `typealias` keyword:
 ```
 typeAlias
-    : modifiers 'typealias' SimpleName (typeParameters typeConstraints)? '='
```
abreslav: I think we should mention explicitly in the text that constraints and variance are not allowed

proposals/type-aliases.md
```diff
@@ -131,13 +159,46 @@ typealias Encoded<E> = ByteArray         // Warning: generic type parameter E is
...
-> Since it is not an error for generic classes,
+> NB this is not an error for generic classes.
+
+Type arguments of a generic type alias should be well-formed types (type projections).
+```
+interface NN<T : Any>
+typealias Predicate<T> = (T) -> Boolean
+
+typealias E1 = Predicate<NN<Int, Int>>          // Error: wrong number of type arguments
+typealias E2 = Predicate<NN<Int?>>              // Error: upper bound violated
```
abreslav: There's a caveat here: 

```
typalias A<T> = NN<T> // is this type well-formed?
```

We need special rules for well-formedness wrt parameters of type aliases, because we need to assume that they satisfy any constraints/variance requirements

proposals/type-aliases.md
```diff
@@ -131,13 +159,46 @@ typealias Encoded<E> = ByteArray         // Warning: generic type parameter E is
...
+>
+>```
+> typealias TMap<T> = Map<T, T>
+>
+> fun <T> foo(m: TMap<T>) {}         // fun <T> (m: Map<T, T>)
+>
+> fun bar(m: TMap<*>) {              // fun     (m: Map<*, *>)
+>     foo(m)                         // Error: cannot infer type parameter T
+> }
+>```
```
abreslav: I think we should add an example about `Array2D<String>` being not assignable to `Array2D<*>`

proposals/type-aliases.md
```diff
@@ -221,6 +289,7 @@ fun throws(): Empty = ...       // Error: return type Nothing should be specifie
 ```
 
 ### Type aliases as values
```
abreslav: I'd say "expression" and not "value"

proposals/type-aliases.md
```diff
@@ -332,12 +400,16 @@ interface Derived : Base {
...
-* Descriptor rendering in IDE (completion, structure view, etc)
-* ...
+TODO
+
+> Abbreviated types should be present in the serialized signatures.
+>
+> Note that additional check is required: if a type alias expansion gives a different type,
+> we can't use corresponding abbreviated form (due to incompatible changes).
+> We would still be able to compile against such binaries, just without abbreviated types in diagnostics.
+> Probably this should be a warning.
```
abreslav: An example needed, IMO

abreslav:
Review finished, some changes requested

proposals/type-aliases.md
dnpetrov: Fixed.

proposals/type-aliases.md
dnpetrov: Done.

proposals/type-aliases.md
dnpetrov: Within the current concept, no additional rules are required: this will be reported as an error at the use site with detailed information on type alias expansion context).
Fixed.

proposals/type-aliases.md
dnpetrov: Why? This is the terminology we use in resolve.
Probably "Type alias companion object" would be more clear.

proposals/type-aliases.md
dnpetrov: I find `Array2D<Int>` and `Array2D<out Number>` even more spectacular.
Done.

proposals/type-aliases.md
abreslav: Well, maybe, but the rules of checking types in the RHS of a typealias are dfferent from those of normal types

proposals/type-aliases.md
abreslav: What wording we use in the code is hardly a justification for saying something ambiguous in public docs :) "Value" is something rather unclear in this context, and, unlike the code, "expression" here means only one thing

PR #8, opened 2016-05-11
Add more info, examples and further questions to the bound callable references proposal

udalov:


proposals/bound-callable-references.md
```diff
@@ -27,10 +27,50 @@ Support "bound callable references" and "bound class literals".
 - Drop the "callable reference to object/companion member" error. Leave other diagnostics for now.
 - Support the following syntax: `<expression>::class` (see semantics below).
 
+### Parsing
+
+Now the LHS of a callable reference expression or a class literal expression (or *double colon expressions*) can mean either an expression or a type, and it's impossible to figure out what it is by looking only at the AST. So we always parse an expression, with optional question marks after it to support the case of nullable types.
```
abreslav: What about `Foo<Bar>::baz` — is the LHS parseable as an expression?

proposals/bound-callable-references.md
```diff
@@ -27,10 +27,50 @@ Support "bound callable references" and "bound class literals".
...
+
+```
+callableReferenceExpression
+    : (expression "?"*)? "::" SimpleName typeArguments?
+    ;
```
abreslav: This looks like a left-recursive rule to me :) Do we really parse it like this?

proposals/bound-callable-references.md
```diff
@@ -84,6 +124,42 @@ fun test(c: C?) {
 }
 ```
 
+#### Nullable references
+
+According to this proposal, in the following syntax
+```
+Foo?::bar
+```
+the left-hand side can only be interpreted as a type because `Foo?` is not a valid expression. Therefore this would necessarily be a reference to an extension function named `bar` to the type `Foo?`.
```
abreslav: This `?` business looks scary: what if we introduce such expression at some point (such proposals pop up every now and them)? Can we do something future-proof here?

proposals/bound-callable-references.md
udalov: Yes, `Foo<Bar>` is parsed as an expression. (It's a call expression without value arguments.)

proposals/bound-callable-references.md
udalov: No, we're parsing it as a slightly special-cased postfix expression. This rule is intentionally simplified here for better readability. Also I'm not sure what productions to include here to explain precisely what's happening...

proposals/bound-callable-references.md
udalov: I'll answer anyway: if such expression is introduced, we'll have to prefer nullable types over such expressions in the LHS of `::`. I don't see what else we can do to support both nullable types and such expressions.

proposals/bound-callable-references.md
abreslav: Maybe we could just have a note that this is a simplified rule

proposals/bound-callable-references.md
abreslav: But the discussion below implies that we'll prohibit `?` in the case when whatever precedes it is resolvable as an expression, no?

proposals/bound-callable-references.md
```diff
@@ -84,6 +124,42 @@ fun test(c: C?) {
...
+
+However, at some point we may want to give it the semantics of a reference to an expression `Foo` which is null when the result of that expression is null. In other words, it would be equivalent to
+```
+Foo?.let { it::bar }
+```
+
+To be able to introduce this later without breaking source compatibility, we should **prohibit** double colon expressions of the form `Foo?::bar` where `Foo` may be resolved as an *expression* (which usually means, there's a variable or a property named `Foo` in the scope). Otherwise the behavior would change:
```
abreslav: Does this apply for qualified LHS? E.g. `Foo.Bar?::baz`

proposals/bound-callable-references.md
```diff
@@ -128,10 +228,31 @@ We could try resolve type in LHS first, and only then try expression. This has t
 
 ## Possible future advancements
```
abreslav: No text between two headers is generally a bad sign :)

I'd explain briefly here what the meaning of this section is: we are aware of these possibilities, but probably won't implement them in 1.1

proposals/bound-callable-references.md
```diff
@@ -128,10 +228,31 @@ We could try resolve type in LHS first, and only then try expression. This has t
...
+
+  ```
+  class Foo {
+      fun bar() {}
+
+      fun test() = ::bar  // equivalent to "this::bar"
+  }
+  ```
+
+  This can be safely introduced later without breaking source compatibility.
```
abreslav: comment on top-level functions?

proposals/bound-callable-references.md
udalov: Ah yes, this will allow to implement either `expr?` or `expr?::foo` in the future

udalov:
I've addressed the comments, please review once again

PR #12, opened 2016-05-17
Proposal for sliding window extensions

cy6erGn0m:


hastebrot:
[Rendered](https://github.com/Kotlin/KEEP/blob/cy/window-sliding/proposals/stdlib/window-sliding.md)

The Rust people have the convention to provide a link in the description to the rendered proposals (well, at least in 90% of the cases). They use branches and pull requests for their RFCs, too [1].

Thought this could be useful for earlier impressions and feedback.

[1] https://github.com/rust-lang/rfcs/pulls

PR #14, opened 2016-05-23
Add proposal for `@JsName`

bashor:


proposals/js/JsName.md
```diff
@@ -0,0 +1,76 @@
...
+- `FIELD` (Field, including property's backing field)
+- `LOCAL_VARIABLE`
+- `TYPE` (Type usage)
+- `EXPRESSION`
+- `FILE`
+
+**Questionable targets:**
++ `VALUE_PARAMETER` (parameter of a function or a constructor)
+<br/>
+Can be useful to provide better interop with some frameworks, e.g. angularjs 1.x, where name of parameters used for injection and some frequently used dependencies contains `$`.
```
konsoletyper: AFAIK, Angular does not force you to name parameters that way. It's just a convention.

proposals/js/JsName.md
```diff
@@ -0,0 +1,76 @@
...
+<br/>
+**Notes:**
+    - this technic (using name of function parameters) is not so popular in JS world, at least I know only about angularjs 1.x where it used;
+    - it can be simply break by minifiers;
+    - angular 2 uses another way for DI;
+    - but it's simple to implement.
+
++ `PROPERTY_GETTER` and `PROPERTY_SETTER`
+<br/>
+Can be useful for native declarations to provide better api in Kotlin.
```
konsoletyper: Does `@JsName` applied to property accessors makes them functions instead of JS property accessors? You should clarify this.

bashor: See update.

konsoletyper:
There is `name` property in `@native` annotation which does the same for native JS declarations. Should we eliminate it in favor of newly introduced `@JsName`?

bashor:
Since we think about replace `native` annotation with `external` modifier it's not matter.  But such replacing depends on this proposal.

proposals/js/JsName.md
```diff
@@ -0,0 +1,88 @@
...
+Additionally it can be used to force using functions for accessors to avoid problems with some minifiers
```
konsoletyper: Another example are extension properties which can only be translated to functions.

proposals/js/JsName.md
```diff
@@ -0,0 +1,88 @@
...
+    - *doesn't generate JS property for them;*
+    - *translates accessors as functions using name from the annotation;*
+    - *translates an accessing to such property as call to accessor function using name from the annotation.*
+
+
+## Common
+
+*__Frontend:__ the annotation can not be combined with overload.*
+
+*__Frontend:__ consider new names when process clashes.*
```
konsoletyper: Also, intention in IDE

proposals/js/JsName.md
```diff
@@ -0,0 +1,88 @@
...
+    *__Backend:__ properties whose accessors has this annotation no longer are treated as JS properties,
+    instead they are interpreted as bunch of accessors (like in JVM), so backend:*
+    - *doesn't generate JS property for them;*
+    - *translates accessors as functions using name from the annotation;*
+    - *translates an accessing to such property as call to accessor function using name from the annotation.*
+
+
+## Common
+
+*__Frontend:__ the annotation can not be combined with overload.*
```
konsoletyper: We could enable it and generate bridges for JsName overloads

proposals/js/JsName.md
bashor: Intention for what?

proposals/js/JsName.md
bashor: Sorry, I made a typo here: "overload" -> "override"

proposals/js/JsName.md
```diff
@@ -28,6 +28,8 @@ Use **binary retention** since it should be available from deserialized descript
 + `FUNCTION` (constructors are not included)
 
 + `PROPERTY`
+<br/>
+New name provided by the annotation should be used for generating name of extension properties too which generated as functions.
```
konsoletyper: Currently, we generate `get/set_$propertyName_$mangleHash` identifier. What is the proposed pattern for accessor functions when there's `@JsName` on a property declaration?

proposals/js/JsName.md
```diff
@@ -81,6 +88,108 @@ Additionally it can be used to force using functions for accessors to avoid prob
...
+
+interface B {
+    @JsName("boo2")
+    fun baz2() {}
+}
+
+class C : A, B
+```
+
+Proposed solution: allow name clashes on native declarations and prohibit on non-native declarations.
```
konsoletyper: Please, clarify which declarations you mean: all of A, B, C can be either native and non-native independently.

proposals/js/JsName.md
```diff
@@ -81,6 +88,108 @@ Additionally it can be used to force using functions for accessors to avoid prob
...
+And doesn't make sense to prohibit to use this annotation on non-final declarations.
+
+So propose to **prohibit using JsName on overrides**, but **report about name clashes** on non-native declarations.
+
+
+**Case 1: declarations with different names from different parents with the same requested name (by JsName)**
+    _(map many names to the one)_
+
+
+_Note: A, B, C can be native._
```
konsoletyper: What about none of A, B, C are native, but overridden method(s) is/are native. What about fake override? What about `bar` having implementation in one of (both of) A and B?

proposals/js/JsName.md
bashor: Names clashed only in `C`

konsoletyper: No matter whether A and B are native or not?

bashor: Should we allow or prohibit declare native class which inherits non-native declarations is the separate question.

bashor: >No matter whether A and B are native or not?

I think it's no matter here

proposals/js/JsName.md
bashor: >What about none of A, B, C are native, but overridden method(s) is/are native.
I don't see what should be changed. Please provide a code fragment if you want to discuss it.

> What about fake override? 
see second example

> What about bar having implementation in one of (both of) A and B?

It already covered by current language rules

proposals/js/JsName.md
```diff
@@ -81,6 +88,108 @@ Additionally it can be used to force using functions for accessors to avoid prob
...
+
+interface B {
+    @JsName("fooB")
+    fun foo()
+}
+
+class C : A, B {
+    override fun foo() {}
+}
+```
```
konsoletyper: Consider the following (quite rare) case:

* `A` and `B` are written by different people and distributed in different libraries
* both `A` and `B` are non-native
* both `foo()` declarations have `@JsName` on them (probably, due to authors wishing to resolve some name conflict) with distinct values.

I see no reason to prohibit this. We could generate something like:

```
C.prototype.fooA = function() {
    // translated body of `C.foo`
};
C.prototype.fooB = C.prototype.fooA;
```

There's no matter which of `fooA` and `fooB` to use on a call site when receiver is known to be `C`. When receiver is either `A` or `B` we generate `fooA` and `fooB` respectively.

proposals/js/JsName.md
```diff
@@ -0,0 +1,197 @@
...
+- `ANNOTATION_CLASS` (Annotation class only)
+- `TYPE_PARAMETER` (Generic type parameter)
+- `FIELD` (Field, including property's backing field)
+- `LOCAL_VARIABLE`
+- `TYPE` (Type usage)
+- `EXPRESSION`
+- `FILE`
+
+**Questionable targets:**
++ `VALUE_PARAMETER` (parameter of a function or a constructor)
```
bashor: I'd prohibit this target now

proposals/js/JsName.md
bashor: It's covered by the alternative solution.

bashor: I can be useful to workaround problems in a hierarchy of native declarations where overrides have incompatible types in Kotlin.

proposals/js/JsName.md
konsoletyper: What should we do with delegated properties when we want to expose them as a pair of accessor methods?

konsoletyper: I would not recommend to use `JsName` to become compatible with google closure compiler. We should not force user to make such manual work for compatibility with minifier. Instead, I'd changed the way how we generate properties:

* for every property declare accessor method(s);
* for every property, declare an ECMA5 property whoose accessors simply call existing method(s). The only exception are *mutable* properties with default accessors;
* in the Kotlin2JS-generated code we always call accessor functions. The ECMA5 properties are only used to expose Kotlin code as a JavaScript library.

This should both simplify our own optimizer and make generated JS more suitable for GCC advanced optimizations.

In this case we could apply `@JsName` annotation independently on a property itself and its accessors.

proposals/js/JsName.md
```diff
@@ -0,0 +1,197 @@
...
+* prohibit on overrides and reuse names from overridden declaration
+    * how, then, resolve conflicts?
+
++ `PROPERTY_GETTER` and `PROPERTY_SETTER`
+<br/>
+Can be useful for native declarations to provide better api in Kotlin.
+Additionally it can be used to force using functions for accessors to avoid problems with some minifiers
+(e.g. closure-compiler) which treat access to properties as side effect free.
+<br/>
+    *__Frontend:__ the annotation can not be applied to only one of accessors.*
```
konsoletyper: What about making property effectively readonly? Sometimes I declare public mutable property with public getter and private setter, because I want this property to be changed indirectly by member functions, but still want to make its value publically available. With such constraint I can't make this property to be available via accessor method. Proposal is:

* enable `@JsName` on one of the accessors;
* when applied to an accessor, `@JsName` causes function with corresponding name to be generated;
* when there's no `@JsName` on an accessor, generate function with mangled name.

proposals/js/JsName.md
bashor: Nothing needs to be changed for delegated properties since they work through accessors.

konsoletyper: Yes, but how should we put `@JsName` on the accessors of a delegated property?

bashor: Compatibility with minifiers is a separate issue.

konsoletyper: > Compatibility with minifiers is a separate issue.
It is, indeed. But what if we'll write a proposal for compatibility and it'll accept things I wrote about accessors (and I insist on at least thinking about it carefully)? We'll have to change behaviour of `@JsName` annotation.

bashor: > We'll have to change behaviour of @JsName annotation.
I think it's ok.

Right now your proposal only relaxes some rule, so it'll affect to written code.

proposals/js/JsName.md
```diff
@@ -0,0 +1,197 @@
...
+
+interface B {
+    @JsName("fooB")
+    fun foo2()
+}
+
+class C : A, B
+```
+
+Proposed solution: prohibit to have more than one name candidate for all declarations including native,
```
konsoletyper: In general, there is case when we get different name candidates in the similar case:

```
interface A<T> {
    fun foo(value: T)
}
open class B {
    open fun foo(value: Int)
}
class C : B, A<Int> {
}
```

in this case we generate bridges and don't report error. So we should prohibit only certain cases. In my prototype I report error only when one of the candidates has `@JsName` on it.

PR #18, opened 2016-05-31
added links to KT issues, added use cases

voddan:


ilya-g:
Merged, and now it's also in the master.
Thanks.

PR #21, opened 2016-06-02
Proposal: Destructuring in lambda parameters

abreslav:


proposals/destructuring-in-parameters.md
```diff
@@ -0,0 +1,156 @@
...
+{ (a, b: B) -> ... }
+{ (a, b): Pair<A, B> -> ... }
+{ (a: A, b): Pair<A, B> -> ... }
+```
+
+> Note: no changes to the syntax of function types is needed since the syntax for call sites (where a lambda is invoked) is not changed.
+  
+### Semantics and checks
+
+Semantically, destructured parameters work as if they are normal parameters and a destructuring assignments are performed before the body of the lambda, in the left-to-right order:
```
VladimirReshetnikov: "a destructuring" -> "destructuring"

proposals/destructuring-in-parameters.md
```diff
@@ -0,0 +1,156 @@
...
+```
+{ 
+  _p1, _p2 ->
+  val (a, b) = _p1
+  val (c, d) = _p2 
+  body() 
+}
+```
+
+**Typing rule**: Types of the components must be compatible with component-functions (`componentN()`) resolvable on corresponding parameters. When such functions can not be applied or their return types are not assignable to declared types of components, an error is reported.
```
VladimirReshetnikov: Suppose a `componentN()` function is generic, and its type parameter occurs in its return type. Will the declared type of a parameter component be used in type inference for the `componentN()` invocation?

erokhins: I suppose that it should work. In general, we should do the following: compute type of `it`, then resolve statements: `val a(: Type) = it.component1()`.

proposals/destructuring-in-parameters.md
```diff
@@ -0,0 +1,156 @@
...
+  component1: P.() -> A,
+  component2: P.() -> B,
+  (a, b): P // error: component functions not found 
+  -> 
+  ...  
+```
+
+When type inference is concerned with the shape of a lambda, a destructured parameter acts as a normal one (single), and its type may be used if declared. Types of components do not affect type inference even when specified: there's no way to know what type would be destructured into the given component types. 
+
+**Naming rule**: all names declared in a parameter list of the same lambda (normal parameters and components of destructured ones) must be unique, i.e. this is an error: `{ (a, b), a -> ... }`
```
erokhins: What about JVM name for lambda parameters? `a_b`?

orangy:
Debugging experience is missing, mostly watch/locals windows. Should we see raw value passed into lambda? I think yes, because component functions could be arbitrary and not necessary preserve the full information of the original instance, especially if not all components are used.

erokhins:
Issue: https://github.com/Kotlin/KEEP/issues/32

PR #22, opened 2016-06-03
Proposal: Sealed class inheritance

erokhins:


abreslav:
Missing:
- current state (to give the reader some context)
- exact rules to be implemented in the checker, and why they don't break the promise of sealed classes

And regarding the questions listed there: what's the state of the current prototype? Does it allow subclasses to sealed classes that are nested elsewhere?

norswap:
This would be a great thing to have.

A question though, with this definition:

```
package foo
sealed class A
class B: A()
```

I would import B with `import foo.B` (not the current `import foo.A.B`), right?
I think that's the desirable behaviour. In current Kotlin I find the need to do `import foo.*` then `import foo.A.*` rather pesky (although I understand why that might be useful).

This is what I like about this proposal: it would let the user choose whether sealed class inheritors are visible at the top-level or not.

erokhins:
Please, use for discussions this issue: https://github.com/Kotlin/KEEP/issues/29

norswap:
Right, I will duplicate my comment there.

PR #24, opened 2016-06-09
Initial proposal for `g` and `G` literal constants suffixes

EPadronU:


yole:
Your discussion of alternatives is incomplete, because you don't mention the possibility of using extension functions for converting values to BigDecimal/BigInteger. `val x = "1.0".big`

Also, your proposed support for mathematical operations between BigInteger/BigDecimal and non-BigInteger values would require introducing overloads to standard library methods for all possible pairs of types where BigDecimal or BigInteger is on one side, with the corresponding increase of the standard library size which will hurt Android developers. This also needs to be discussed in the proposal.

EPadronU:
Thanks @yole. I'll try to improve the proposal with those considerations.

voddan:
How about an option to put those in a separate library on JB repository? Or even bundle it with the Kotlin plugin, but not to include in stdlib

EPadronU:
Maybe something along the lines of:

```kotlin
fun Number.toBigDecimal(): BigDecimal {
  ...
}

fun BigDecimal.plus(other: Number): BigDecimal {
  return this + other.toBigDecimal()
}
```

abreslav:
It seems to me that BigDecimal is no better than any other type (BigInteger, for example), and we'd better add custom suffixes as a language feature or not add any suffixes at all.

Example:

```
suffix fun Int.g(): BigDecimal = ...

// usage
val bd = 1g
```

cbruegg:
Suffix functions wouldn't eliminate the need for native BigDecimal literals though, since there are BigDecimals that are not representable by doubles. We'd have to resort to non-typesafe strings for those.

EPadronU:
I'm sorry for the long delay. I'll be waiting for your feedback. Cheers!

HaasJona:
Some suggestions:

- I think the extension functions should be a separate proposal. Mixing floats and BigDecimals for example seems to be dangerous to me and should require an explicit cast/conversion method in my opinion.

- The extension functions can also be easily implemented by the programmer if they are considered useful for some project. Literals not so much.

The main benefits for BigInteger/BigDecimal literals in my opinion are:

- Possible speed improvements vs. Strings. The compiler can parse the number and store the data more efficiently (maybe a byte array or serialized) so that the actual object can be constructed faster at runtime. Literals used multiple times maybe could also all be the same object.

- The compiler can check if the number is valid instead of getting a runtime error.

- Maybe number literals can be improved in general, to fix for example 
    - https://youtrack.jetbrains.com/issue/KT-4749
    - https://youtrack.jetbrains.com/issue/KT-2780#comment=27-1133713

- Make BigInteger/BigDecimal first class citizens. Operator overloading already has helped a lot vs. normal Java when using these classes. Literals would make them perfect.

voddan:
How often are `BigInteger` and others are used in projects? Please add a Github scan/statistics showing the frequency of use for those classes and long arithmetics in general. 

In the case those numbers are small, I would argue to avoid modifying the Kotlin syntax and/or move the proposed functionality out of stdlib into a separate library

HaasJona:
We have ca. [850,000](https://github.com/search?l=java&q=BigInteger&type=Code&utf8=%E2%9C%93) uses of BigInteger in Java code on GitHub, and 1,000,000 BigDecimal uses. Short and Float for example are only used about 8 to 10 times as much, which isn't super interesting, because they're native data types and much easier to use.

BigInteger/BigDecimal is necessary/useful for
 - a lot of cryptography
 - representing money in a safe way
 - handling unlimited precision user input
 - high precision mathematics
 - make code overflow safe if performance isn't as important as security.

They would be probably used much more in Java when they had proper operators and literals. Ruby for example uses unlimited precision integers by default for normal Integers and this seems to be no problem at all and widely used.

voddan:
@HaasJona Could you specify please the percentage of projects using long arithmetics? The number of total uses does not say much in this situation. For some reason I doubt that 10% of all (big enough) projects use `BigInteger`

HaasJona:
I have no idea how to search for that, sorry. I agree with you that probably less that 10% of the projects use BigInteger/BigDecimal, but the projects that use them, will generally use them a lot.

DonWills:
We write a lot of business applications that require exact decimal math.  IMO, interoperability with Java in this case is nice, but in reality it is the exact fractional arithmetic of BigDecimal that is necessary and otherwise unavailable.  I do not care about the "Bigness", I care about the exactness.  Thus BigInteger is of no interest.  And the unlimited nature of BigDecimal is also of no interest - 20 digits of precision more than sufficient.

I suggest the use the single character 'd' as the suffix.  And add new class called kotlin.Decimal that is a subclass of java.lang.Number.  For hardware that supports the IEEE d128 format, the implementation can be really(!) performant.

BenLeggiero:
Here's a better idea: Make it possible to represent any class as a literal, so long as it conforms to an interface (named something like `ExpressibleByIntegerLiteral` or `ExpressibleByFloatLiteral`). This would lead to much greater flexibility and allow 3rd-party developers to create their own big-decimal-esque implementations, etc.

gildor:
@BenLeggiero Looks like this proposal from @abreslav is exactly about that:
https://github.com/Kotlin/KEEP/pull/24#issuecomment-224888801

BenLeggiero:
@gildor kinda... But @abreslav's idea also alters the appearance of an integer literal. My comment is inspired by Swift's approach, which doesn't do this. 

For instance, in Swift, if you extend `NSDecimalNumber` (their `BigDecimal`) to conform to `ExpressibleByFloatLiteral`, you can simply do this:

```Swift
var foo: NSDecimalNumber = 4.2

func takeBigNumber(_ bar: NSDecimalNumber) { }

takeBigNumber(123.45)
```

So no weird letter suffix on the literal, just the natural literal we're used to.

Kotlin syntax might look like this:

```Kotlin
// In Kotlin stdlib:
interface ExpressibleByFloatLiteral {
    constructor(literal: Double) // This would also have to be added to the language. I don't like how it breaks the 100% Java interop, but it's very nice to have and would open up a world of possibilities. Like this one!
}

// In my code:
class MyNumberClass(val iPart: String, val fPart: String): ExpressibleByFloatLiteral {
    constructor(literal: Double) : this(iPart = literal.toLong().toString(), fPart = (literal - literal.toLong()).toString().substringAfter(".", "0")
}

var foo: MyNumberClass = 4.2 // Creates MyNumberClass("4", "20000000000000018")

fun takeBigNumber(bar: MyNumberClass) { }

takeBigNumber(123.45) // Creates MyNumberClass("123", "45000000000000284")

```

gildor:
@BenLeggiero I think it's not a topic of the current proposal. But my 2cents is that such implicit conversion is bad and doesn't follow Kotlin "be explicit" principle.

BenLeggiero:
@gildor I think of it more as separating the language's syntactic number literals from a runtime type. They can still have an implicit type if none is specified, like `Int` and `Double`, but with my above concept, we don't lock ourselves into a finite number of types a number can represent

abreslav:
@BenLeggiero My concern here would be that not all literals should fit into a Long with either their I or F part, so we'd have to resort to strings or something. Overall direction is interesting though. Thanks

BenLeggiero:
@abreslav That is my biggest peeve with Swift's implementation; all literals must fit into a `Float64` or `Int64`. If I were to design it, I'd have a much better way to do this that wouldn't lose precision. Anyway, glad that I got y'all thinking; maybe we could move in this direction, even if we don't go all the way :)

elizarov:
Linking here the corresponding youtrack issue: https://youtrack.jetbrains.com/issue/KT-20913

Issue #25, opened 2016-06-14
Local delegated properties


max-kammerer:

dzharkov:
The declarations has been released in Kotlin 1.1, thus I'm closing the KEEP. Do not hesitate to open a YouTrack issue for any additional suggestions.

PR #26, opened 2016-06-14
Initial proposal for local delegated properties

max-kammerer:


abreslav:
See #25

abreslav:
Merged manually

PR #33, opened 2016-07-12
Android Extensions: View holder pattern support and caching options

yanex:


denis-trailin-hs:
👍

Angryhead:
Showing my support for this!

KasparPeterson:
Really nice idea!

rongi:
There is already no need to write view holders with Kotlin etensions
http://stackoverflow.com/questions/33304570/kotlin-synthetic-in-adapter-or-viewholder

yanex:
@rongi Well, such view holder will support caching functionality seamlessly.

Judrummer:
Agree with you too!!

Wavesonics:
Without something like this we'd either be forced to use android extensions at the fragment level and something like Kotterknife at the ViewHolder level, or, more likely, just use KotterKnife everywhere.

szymenpl:
Anyone knows when caching in custom views and viewholders will be available? Or maybe what is reason for method `public View _$_findCachedViewById(int var1)`, which is generated in custom view, but not used. (`findViewById` is always used)

06needhamt:
👍

neiljaywarner:
(thumbsup)

rongi:
👍

neiljaywarner:
(thumbsup)

neiljaywarner:
When is this going to be production ready?

vkotovv:
@yanex Is it still experimental?

PR #36, opened 2016-07-21
Short syntax for property getters

abreslav:
Full proposal is [here](https://github.com/Kotlin/KEEP/pull/36/files).
## Synopsis

`Val`'s are often implemented with a one-line getter:

``` kotlin
val length: Int
    get() = size
```

This syntax is verbose in at least three ways:
- it requires a type annotation,
- it requires a `get()`, 
- the `get()` is usually on a separate line.

This is relevant for
- simple properties: `val length: Int get() = size`
- overridden properties: `override val length: Int get() = 0`

**NOTE**: The straightforward syntax like `override val length = 0` has the performance disadvantage of creating an unnecessary field for such a property.

### DIFF
--- a/proposals/short-syntax-for-getters.md
+++ b/proposals/short-syntax-for-getters.md
@@ -31,14 +31,15 @@ This is relevant for
 
 In the projects we have access to, we have the following frequency of such cases:
 
-| Project | Total `val`'s | Simple get-only `val`'s | Overridden get-only `val`'s |
-| ------- | ---------------- | ----------- | ------ |
-| Kotlin        | 34850 | 2624 (7.52%) | 826 (2.37%) |
-| Ktor          | 2095 | 116 (5.53%) | 73 (3.48%) |
-| IntelliJ IDEA | 7016 | 230 (3.27%) | 80 (1.14%) |
-| JB Project 1  | 1264 | 25 (1.97%) | 0 (0%) |
-| JB Project 2  | 9742 | 272 (2.79%) | 33 (0.33%) |
-| All of GitHub | # (TODO) | #(%) | #(%) |
+Repo | Total val's | With getter | % w/expression body | Overrides with getter | % overrides w/expression body
+---- | ---- | ---- | ---- | ---- | ----
+Kotlin | 34850 | 2624 | 94.93% (2491) | 826 | 94.79% (783)
+JB_Project_1 | 9742 | 272 | 71.69% (195) | 33 | 84.85% (28)
+IntelliJ IDEA | 7016 | 230 | 87.39% (201) | 80 | 83.75% (67)
+Ktor | 2095 | 116 | 99.14% (115) | 73 | 98.63% (72)
+All of GitHub | TODO
+
+> Computed with [ktrawler](https://github.com/abreslav/ktrawler/blob/master/src/val-getters-runner.kt)
 
 **NOTE**: The straightforward syntax like `override val length = 0` has the performance disadvantage of creating an unnecessary field for such a property.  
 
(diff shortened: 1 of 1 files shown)


cypressious:
What about the solution described in https://youtrack.jetbrains.com/issue/KT-550, i.e. simply allowing to omit the type declaration? `val foo get() = 1` doesn't look too bad IMO.

Other thoughts about the different solutions:

`val foo => 1`:

Pro: is similar to C# 

`val foo() = 1`

Pro: is symetrical to function definition `fun foo() = 1`
Con: breaks the characteristic that functions have parans while properties don't.

abreslav:
> What about the solution described in https://youtrack.jetbrains.com/issue/KT-550, i.e. simply allowing to omit the type declaration? val foo get() = 1 doesn't look too bad IMO.

This is not always a solution. Sometimes there's a need to specify the type (e.g. the rhs has a non-denotable/private type)/

cypressious:
>This is not always a solution. Sometimes there's a need to specify the type (e.g. the rhs has a non-denotable/private type)/

Ok, but if you need the type, you can still use the existing syntax.

The simple solution solves 2 of the 3 stated problems:

- it requires a type annotation
- the get() is usually on a separate line -> it doesn't have to be, especially when the type definition is gone

If you feel that removing the `get()` is the most important goal of the proposal, you should make that more clear.

abreslav:
> > This is not always a solution. Sometimes there's a need to specify the type (e.g. the rhs has a non-denotable/private type)/
> Ok, but if you need the type, you can still use the existing syntax.
> The simple solution solves 2 of the 3 stated problems:

This kind of solutions is usually referred to as "imperfect". Doesn't rule one out, but no one is eager to add such a feature to the language. We'd better think some more :)

cypressious:
Ok, I guess I'm not thinking big enough :D

If I had to choose between the given solutions, I'd take `val foo() = 1`. It can be seen as a compact form of `val foo`~~: String get~~() = `1` and it is consistent with having the `=` operator for both, property and function declarations.

fab1an:
I'm definitely in favor of `val x get() = "string"`. 

It reminds of the short function syntax `fun get() = "string"` and still shows the difference to a real `fun`.

drothmaler:
@udalov I just realized, that the light arrow '->' already has another meaning, as it is also used in 'when' expressions. So it would be logical to keep using it in the sense of "here comes the implementation for what is on the left side". :-)

cypressious:
I personally dislike the arrow `->` syntax for this proposal.

valentinkip:
"->" cannot be used because it conflicts with the syntax for functional types:

`val v: Type -> x`

becomes ambigous.

Why not use "=>" instead?

`val v => <expression>`

Looks nice for me.

abreslav:
Related feature that may leverage the same syntactic decisions taken here: https://youtrack.jetbrains.com/issue/KT-13414

ilya-g:
Please add option 4 to consideration proposed by @cypressious in [this comment](https://github.com/Kotlin/KEEP/pull/36#issuecomment-234282497): `val foo get() = 1`

BenLeggiero:
I rather like Swift's syntax:

```Swift
var computed: Int {
    return computation()
}
```

Or its verbose counterpart:
```Swift
var computed: Int {
    get {
        return computation()
    }
}
```

Or with a setter:
```Swift
var computed: Int {
    get {
        return computation()
    }
    set { // Alternatively set(myNewValueName)
        _backingField = newValue
    }
}
```

This should be possible in Kotlin without introducing ambiguity.

ZkHaider:
Agree on swift parity

CLOVIS-AI:
This is an old issue, but `val a get() = 1` has been stable for a few versions now. I don't believe this proposal brings any improvements anymore.

PR #37, opened 2016-07-22
Generic values() and valueOf() for enums

abreslav:
Generic operations on enums naturally involve retrieving all enum values or converting a string into a value. Example:

``` kotlin
// Print options on the screen
fun <reified T : Enum<T>> renderOptions(render: (T) -> String) {
    val values = /* get T.values() somehow */
    for (v in values) {
        println(" * " + render(v))
    }
}
```

Currently, there's no performant way of retrieving `values()`/`valueOf()` from an enum that is given as a generic parameter. This can be done through reflection, but it would be too slow for many use cases.

> The initial YouTrack issue: 
> - _[KT-10569](https://youtrack.jetbrains.com/issue/KT-10569) Cannot iterate over values of an enum class when it is used as a generic parameter_  

We propose to support `values()`/`valueOf()` for reified type parameters that extend `Enum<T>` through a compiler intrinsic that would generate appropriate static calls to the particular enum class.

### DIFF
--- a/proposals/generic-values-and-valueof-for-enums.md
+++ b/proposals/generic-values-and-valueof-for-enums.md
@@ -1,0 +1,98 @@
+# Generic values() and valueOf() for enums
+
+* **Type**: Design proposal
+* **Author**: Andrey Breslav
+* **Contributors**: Stanislav Erokhin, Alexander Udalov
+* **Status**: Under consideration
+* **Prototype**: Not started
+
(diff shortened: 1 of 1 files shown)


cypressious:
I suppose `enumValues()` and `enumValueOf()` is the current solution for this?

ilya-g:
The discussion continues in the issue #61

PR #38, opened 2016-07-25
Scope control for implicit receivers

abreslav:
When we nest builders of different types into one another, members of the outermost implicit receiver are available in the innermost lambda (unless they are shadowed by something in inner lambdas):

``` kotlin
table {
    tr {
        tr {} // PROBLEM: Table.tr() is available here, 
              // because the implicit receiever of `table {}` is in scope
    }
}
```

Sometimes this effect is useful (the outermost receiver may provide some "global" services relevant to all necessary scopes, e.g. some context-like data), but often times it's problematic.

### DIFF
--- a/proposals/scope-control-for-implicit-receivers.md
+++ b/proposals/scope-control-for-implicit-receivers.md
@@ -2,9 +2,9 @@
 
 * **Type**: Design proposal
 * **Author**: Andrey Breslav
-* **Contributors**: Ilya Ryzhenkov
+* **Contributors**: Ilya Ryzhenkov, Denis Zharkov
 * **Status**: Under consideration
-* **Prototype**: Not started
+* **Prototype**: In Progress
 
 Discussion of this proposal is held in [this issue](TODO).
 
@@ -50,7 +50,37 @@ This approach has the following major disadvantages:
 - one can't anticipate all cases of DSL composition, so there's no way to know the complete set of functions to screen against,
 - there's no way to know in advance all possible extensions defined for outer receivers, so screening for them is also impossible.  
 
-## Proposed solution
+## Proposed solutions
+
+### DslMarker (the one currently implemented)
+
+We introduce `DslMarker` annotation with the following definition:
+```kotlin
+@Target(ANNOTATION_CLASS)
+@Retention(RUNTIME)
+@MustBeDocumented
+annotation class DslMarker
+```
+
+An annotation class `Ann` is called *a DSL marker* if it is annotated with the `@DslMarker` annotation.
+
+The general rule:
+* an implicit receiver may *belong to a DSL `@Ann`* if it's marked with a corresponding DSL marker annotation
+* two implicit receivers of the same DSL are not accessible in the same scope
+    * the closest one wins
+    * other available receivers are resolved as usual, but if the resulting resolved call binds to such a receiver, it's a compilation error
+
+Marking rules: an implicit receiver is considered marked with `@Ann` if
+* its type is marked, or
+* its type's classifier is marked
+    * or any of its superclasses/superinterfaces
+
+Additional notes
+* Receivers can be accessed through `this@label` regardless of their marking
+* Multiple markers on the same receivers:
+    * If receivers tower looks like (2, 1+2, 1), we consider both 2 and 1+2 unavailable
+
+### ScreenFrom (obsolete)
 
 We propose to control availability of outer receivers inside a builder lambda through an annotation on the lambda parameter:
 
@@ -74,7 +104,7 @@ An empty set of receivers may mean either
 - screen from all (equivalent to `@ScreenFrom(Any::class)`)
 - screen from nothing.
 
-## Terminology and Naming
+#### Terminology and Naming
 
 The proposed technique requires a name. Possible options include:
 - scope control for implicit receivers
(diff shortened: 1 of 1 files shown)


cypressious:
I think you should add a code snippet with the annotation applied and show show which receivers are/aren't available.

ymolists:
I think the default should be the "no spilling behavior. Meaning "members of the outermost implicit receiver should *not*  be available in the innermost lambda unless i specifically mark the outermost receiver as such. The default should be what modern OO languages do. If someone wants to take shortcuts and mark certain members as available "implicitly" to all innermost lambdas then they should be asked to mark the outermost  interface as such. I am asking why is it necessary here to deviate from the norm ?

cypressious:
@ymolists That would be a breaking change. The solution needs to be backward compatible.

cy6erGn0m:
@ymolists For sure it would be too annoying to annotate every lambda every time I need to get members available inside.

cy6erGn0m:
The main issue here is that we can't screen from the type we don't know. So if there are some extra 3rd party extensions for DSL then there is no way to screen from it and neither the library owner nor extensions owner can fix that, isn't it?

ilya-g:
>  The default should be what modern OO languages do. 
> I am asking why is it necessary here to deviate from the norm ?

@ymolists Could you provide some references to the definition of the _norm_ and how the modern OO languages implement it? (better by editing the original comment to keep the discussion compact)

Thanks in advance.

ymolists:
Can you use the syntax from [pony ref cap](http://tutorial.ponylang.org/capabilities/reference-capabilities.html) instead of annotations to allow/permit scoping syntax ? (pretty) please ?

ymolists:
@abreslav i realised the problem is somewhere else (((

Anyone can tell me why i need the implicit receiver when i can do it explicitly with not much fuss like below ??? The point here is that with this approach inner.type will have the IDE flagging the compilation error. Is the proposed annotation based solution going to show the error at compile time instead of run time ?

```

    outerType {
        val outer = this
        
        type {
            val inner = this

            inner.type {

            }

            outer.type {
                
            }
            
        }


    }

}

```

abreslav:
> Anyone can tell me why i need the implicit scope based receiver when i can do it explicitly with not much fuss like above via a variable declaration ??? ...

@ymolists I totally understand it when people are getting emotional about language design :), but I'd like to make it clear that any changes we envision here must be backward-compatible. What you are talking about (regadless of its merits) seems a few bits too radical and would break a crazy lot of existing code, so we should look for something else.

mfulton26:
What about defining some way to "opt-out" of implicit outer scope?

```kotlin
class HTML() : TagWithText("html") {
    fun head(init: Head^.() -> Unit) = initTag(Head(), init)

    fun body(init: Body^.() -> Unit) = initTag(Body(), init)
}

fun html(init: HTML.() -> Unit): HTML {
    val html = HTML()
    html.init()
    return html
}

fun <T> T.toHTML() = html { // this@html == this@toHTML
    head { /* ... */ }      // this@head == this@toHTML
    body {                  // this@body == this@toHTML
        head { } // Unresolved reference: head
    }
}
```

The above syntax uses `^` to represent 'outer scope'. Another possible syntax would be to use `../` (like a relative path to a parent). e.g.:
```kotlin
    fun head(init: Head../.() -> Unit) = initTag(Head(), init)
```

apatrida:
@mfulton26 `^` has bad luck as a choice for operator in historical Kotlin, but please use  it here so that no one suggests using the hat operator ever again for `emit` expressions.

I think you are basically saying that `^` puts up a wall that can't been seen past.  Then how do you see past it when you want to use something from outside the lambda while inside, you have to apply a specific `this@somewhere.title` for example to reach out?

mfulton26:
@apatrida, I'm not aware of the history here with `^` but the proposal isn't for a wall but to exclude the caller's implicit receiver.

abreslav:
@mfulton26 Spending our very limited symbol budget on this doesn't look great at all, but an opt-out is also an option to consider, of course

flstaats:
Ideally as a DSL designer I would prefer to allow the definition of any specific element in a lambda scope opt-out of being an implicit receiver for lambda lower scopes.  But I would not like that to be a generic constraint for all other enclosing scopes.  In most programming contexts you want a lambda to be able to capture any visible variables or functions.  However the grammar of a DSL may restrict what is "valid" to be captured and what is not valid.  But lambdas that are part of the DSL may be intermixed with lambdas that are being used as normal programming constructs.

Has we considered a model where instead of marking the entire lambda as in/out of scope we mark the scope with compile time annotations on the receiver classes values, variables, properties and and functions individually?  I expect that when I create a DSL my job is to make it easy for the user, not the DSL author.  As the designer of the DSL I should know what should be safe to expose in contained scope and what is not.  If we do it this way then it could be as simple as applying annotations to the items we do not want to be visible.  In the HTML example above we would only need to add the annotation to fun head and fun body inside the class HTML().

One strength of this proposal is that it would not depend on resolving how to implement annotations on Types.  It would also be much more flexible for the DSL author to define the scoping that best meets their domains needs.  Also this would be backwards compatible to the existing behavior.

abreslav:
@flstaats Please check the renewed proposal. It seems that it is more in line with what you are talking about

elizarov:
Let's imbue DslMarker annotation with additional superpowers:

**Definition:** A class or interface annotated with _DSL marker_ annotation is called a _DSL class_.

1. **Proposed rule:** All members inherited by DSL class from non-DSL classes are not available in extension functions of this DSL class. **Benefits:** It gives DSL designer a fine-grained and exlusive control on what consumers of their DSL see. This natuarally excludes from DSLs equals and hashcode functions inherited from Any, sinve Any is not a DSL class. If a designer of a DSL wants to expose members of superclass in their DSL, they can explicitly provide the corresponding functions or properties in their DSL. 

2. **Proposed rule:** DSL class reveiver cannot be accessed by unqualified or qualified 'this' from its extension functions, moreover it does not shadow 'this' from outer scope. **Benefits:** This enables quite unobstructive use of various DSLs in large variety of contexts and makes DSLs _really_ look like build-into-the-language control constructs. If a designer of a DSL wants to expose 'this' reference in their DSL, they can explicitly provide the corresponding functions or properties in their DSL. For example, HTML DSL designer can provide 'body' property in its Body class that returns 'this' if its children are expected to access it. 

3. We can foresee a following _boilerplate_ pattern, where DSL designers are writing a small DSL with just one class and are forced to define DSL marker annotation just for this one class. **Proposed rule:** DslMarker annotation can be used directly on DSL classes or interfaces in adddition to annotations. This DSL class or interfrace instances and of all classes or interfaces inheriting from it _belong_ to the DSL defined by the class annotated with DslMarker. 

Closing notes
--------------

These rules will be very useful for all kinds of DSLs, but they will especially make it easy to write _coroutine DSLs_ (controllers) to enable their seamless and unobtrusive integration into their surrounding code.

dzharkov:
I think it's worth moving the discussion to the [issue](https://github.com/Kotlin/KEEP/issues/57) rather than in the pull request that may be closed soon. // cc @elizarov

Groostav:
For those looking for a follow-up, this was introduced in 1.1 as `@DslMarker`, documented [here](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-dsl-marker/)

PR #39, opened 2016-07-25
Types as annotation arguments

abreslav:
Currently, we can only mention classes in annotation arguments: `@Ann(Foo::class)`. Although in some contexts terms _class_ and _type_ can be used interchangeably, those are two different things, for example, a class `Foo` can be mentioned in many different types: `Foo`, `Foo<Bar>`, `Foo<Foo<Bar>>`, `Foo<*>`, `Bar<Foo>, etc.

Sometimes we need to refer to types, not classes. For example, in many cases it's useless to say `List::class`, we need more information like `List<String>` or `List<Nothing>`.

We propose adding support for types as annotation arguments (and, optionally, elsewhere in the language):
- an annotation parameter may be declared to be of type `KType`,
- an annotation argument may be a type (in some syntactic form, e.g. `@Ann(Foo<Bar>)`).

Example (syntax is subject to discussion):

``` kotlin
annotation class Ann(val type: KType)

@Ann(type = Foo<Bar>)
fun test() {} 
```

This is relevant for #38

### DIFF
--- a/proposals/types-as-annotation-arguments.md
+++ b/proposals/types-as-annotation-arguments.md
@@ -1,0 +1,126 @@
+# Types as annotation arguments
+
+* **Type**: Design proposal
+* **Author**: Andrey Breslav
+* **Status**: Under consideration
+* **Prototype**: Not started
+
+Discussion of this proposal is held in [this issue](TODO).
+
+## Synopsis
+
+Currently, we can only mention classes in annotation arguments: `@Ann(Foo::class)`. Although in some contexts terms *class* and *type* can be used interchangeably, those are two different things, for example, a class `Foo` can be mentioned in many different types: `Foo`, `Foo<Bar>`, `Foo<Foo<Bar>>`, `Foo<*>`, `Bar<Foo>, etc.
+   
+Sometimes we need to refer to types, not classes. For example, in many cases it's useless to say `List::class`, we need more information like `List<String>` or `List<Nothing>`.
+    
+We propose adding support for types as annotation arguments (and, optionally, elsewhere in the language):
+- an annotation parameter may be declared to be of type `KType`,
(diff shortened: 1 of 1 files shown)


damianw:
How would this be represented in Java?

This might also be a good opportunity to bring up built-in type literals for the rest of the language, to replace the java-style hacks like `TypeLiteral` and `TypeToken`. I'd think that if I can represent a type literal in an annotation, I should be able to do so in code as well.
```Kotlin
val listOfStringType = List<String>::type
```

abreslav:
> How would this be represented in Java?

We have two options:
- represent such types as strings, e.g. `"Foo<Bar>"`
- represent them as annotations, e.g. `@Ann(type = @Type(name = "Foo", args = {@Type(name = "Bar`)})`

> This might also be a good opportunity to bring up built-in type literals for the rest of the language

Fair point.

proposals/types-as-annotation-arguments.md
```diff
@@ -0,0 +1,126 @@
...
+### Reflection-like literals
+
+Something like `Foo<Bar>::type`.
+
+- **pro**: looks like reflection,
+- **con**: the type on the left may have a member named `type` (which is too popular to be made a keyword), so this syntax is ambiguous with a member reference.
+
+### `typeof`
+
+Something like `@Ann(typeof(Foo<Bar>))`. Note: `typeof` is [reserved as a keyword](https://github.com/JetBrains/kotlin/blob/master/compiler/frontend/src/org/jetbrains/kotlin/lexer/Kotlin.flex#L249) already.
```
LouisCAD: This link is dead. Permalinks should be used instead to prevent future occurrence.

udalov: Thanks, fixed

nikitabobko:
Closing stale PR

PR #40, opened 2016-07-25
Opening classes for frameworks

abreslav:
Kotlin has classes and their members `final` by default. This has been causing debates that are summarized in [this forum post](https://discuss.kotlinlang.org/t/a-bit-about-picking-defaults/1418). While all the reasoning behind this default remains valid, there are use cases that tend to be painful, most notable are frameworks using CGLib proxies (e.g. Spring AOP, Mockito) and Gradle that decorates task classes at runtime.

We propose to enable a compilation strategy that does not classes and their members `final` if they are marked with framework-specific annotations, such as `@Service`, `@Repository`, `@Bean`,  `@Configuration`.

Open issues:
- [ ] Support for custom notion of "indirectly annotated" (see Spring's `@Component`)
- [ ] Investigate possible impact on build systems/IDE integration of having this implemented as a compiler plugin

### DIFF
--- a/proposals/opening-classes-for-frameworks.md
+++ b/proposals/opening-classes-for-frameworks.md
@@ -14,8 +14,7 @@ Discussion of this proposal is held in [this issue](TODO).
 
 Kotlin has classes and their members `final` by default. This has been causing debates that are summarized in [this forum post](https://discuss.kotlinlang.org/t/a-bit-about-picking-defaults/1418). While all the reasoning behind this default remains valid, there are use cases that tend to be painful, most notable are frameworks using CGLib proxies (e.g. Spring AOP, Mockito) and Gradle that decorates task classes at runtime.
 
-We propose to enable a compilation strategy that does not classes and their members `final` if they are marked with framework-specific annotations, such as `@Service`, `@Repository`, `@Bean`,  `@Configuration`.
- 
+We propose to enable a compilation strategy that does not make classes and their members `final` if they are marked with framework-specific annotations, such as `@Service`, `@Repository`, `@Bean`,  `@Configuration`. 
 
 ## References
 
@@ -27,10 +26,12 @@ We propose to enable a compilation strategy that does not classes and their memb
 - [~~KT-11098~~](https://youtrack.jetbrains.com/issue/KT-11098) Add quick fix for "Spring @Configuration/@Component annotated classes or @Bean annotated methods should be open"
 
 > There's a somewhat similar issue with default constructors and JPA: 
-- [Hibernate/JPA](http://stackoverflow.com/questions/32038177/kotlin-with-jpa-default-constructor-hell)
-- [Hibernate + Guice Persist](https://discuss.kotlinlang.org/t/jpa-guice-gotchas/425)  
-- [JPA](https://discuss.kotlinlang.org/t/feature-request-a-modifier-annotation-for-data-classes-to-provide-a-non-arg-constructor-on-jvm/1549/4)
-- [Sugar ORM](https://discuss.kotlinlang.org/t/using-sugar-orm-with-kotlin/439/4)
+>- [Hibernate/JPA](http://stackoverflow.com/questions/32038177/kotlin-with-jpa-default-constructor-hell)
+>- [Hibernate + Guice Persist](https://discuss.kotlinlang.org/t/jpa-guice-gotchas/425)  
+>- [JPA](https://discuss.kotlinlang.org/t/feature-request-a-modifier-annotation-for-data-classes-to-provide-a-non-arg-constructor-on-jvm/1549/4)
+>- [Sugar ORM](https://discuss.kotlinlang.org/t/using-sugar-orm-with-kotlin/439/4)
+>
+> And another one about [SAM conversions vs extension function types](https://youtrack.jetbrains.com/issue/KT-12848).
 
 ## Possible implementations
 
--- a/proposals/opening-classes-for-frameworks.md
+++ b/proposals/opening-classes-for-frameworks.md
@@ -1,0 +1,113 @@
+# Opening classes for frameworks
+
+* **Type**: Design proposal
+* **Author**: Andrey Breslav
+* **Contributors**: Dmitry Jemerov, Sébastien Deleuze, Yan Zhulanow, Andy Wilkinson 
+* **Status**: Under consideration
+* **Prototype**: Not started
+
+## Feedback
+
+Discussion of this proposal is held in [this issue](TODO).
+
+## Synopsis
+
+Kotlin has classes and their members `final` by default. This has been causing debates that are summarized in [this forum post](https://discuss.kotlinlang.org/t/a-bit-about-picking-defaults/1418). While all the reasoning behind this default remains valid, there are use cases that tend to be painful, most notable are frameworks using CGLib proxies (e.g. Spring AOP, Mockito) and Gradle that decorates task classes at runtime.
+
+We propose to enable a compilation strategy that does not classes and their members `final` if they are marked with framework-specific annotations, such as `@Service`, `@Repository`, `@Bean`,  `@Configuration`.
+ 
+
+## References
+
+- [KT-12149](https://youtrack.jetbrains.com/issue/KT-12149) Provide a way to avoid mandatory open qualifier for CGLIB proxies
+- [KT-6256](https://youtrack.jetbrains.com/issue/KT-6256) Allow a class to be open by default
+- A. Breslav, [A bit about picking defaults](https://discuss.kotlinlang.org/t/a-bit-about-picking-defaults/1418)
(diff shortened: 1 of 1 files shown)


sdeleuze:
What I like in the "META-INF file" solution is that's:
 - Not intrusive
 - Will avoid some IDE and tooling integration issues compared to other solutions
 - Allow each library to contribute its "open annotation" list without some custom configuration

We are ok to integrate such META-INF file directly in Spring Framework JAR.

Instead of listing every single annotation like `@Service` or `@Configuration`, if think that could be better to just identify classes annotated directly or indirectly by `@Component` (see [Configuration](https://github.com/spring-projects/spring-framework/blob/master/spring-context/src/main/java/org/springframework/context/annotation/Configuration.java) annotated with `@Component`).

abreslav:
@sdeleuze 

> What I like in the "META-INF file" solution is that's:
> * Will avoid some IDE and tooling integration issues compared to other solutions

Could you clarify? What issues do you mean?

> * Allow each library to contribute its "open annotation" list without some custom configuration

Isn't such file an example of custom configuration?

abreslav:
> Instead of listing every single annotation like @Service or @Configuration, if think that could be better to just identify classes annotated directly or indirectly by @Component (see Configuration annotated with @Component).

I see your point, and it's very appealing.

One problem: this brings the notion of "indirectly annotated" into the picture, and while it may be very obvious what that means for Spring, I'm not sure that it can be defined in a one-size-fits-all way. This speaks in favor of a compiler plugin where each framework could potentially define its own predicate using the full power of a programming language, not a restricted declarative format in `META-INF`.

sdeleuze:
One of my main concern is to make such behavior reliable and minimize the amount of customization for the end user.

I would like to avoid issues like the default open class behavior is taken in account by the Gradle/Maven but not the IDE. META-INF file solution is maybe a kind of custom configuration but it would be taken in account by any Kotlin 1.1 based compiler without any custom configuration at build tool level.

My remark about `@Component` is also about a similar concern, we have a lot of different `@Component` annotations is Spring projects, and users can define their own. So the ability to detect any kind of indirectly annotated `@Component` would make such support much more reliable.

If the chosen solution is a compiler plugin, what would be the requirement for us? Some kind of `org.springframework.kotlin.CompilerPlugin` class parameter added to the Gradle or Maven plugin configuration? Do you plan to bundle such class with Maven/Gradle Kotlin plugin in order to avoid requiring another plugin or do you expect us to provide it as an additional dependency?

abreslav:
> I would like to avoid issues like the default open class behavior is taken in account by the Gradle/Maven but not the IDE. META-INF file solution is maybe a kind of custom configuration but it would be taken in account by any Kotlin 1.1 based compiler without any custom configuration at build tool level.

Fair point, a **pro** for the static config file is that it can be read with equal ease by all tools. Effectively, all tools would inherit this functionality from the compiler. I'm not entirely convinced that adding such (arguably ad hoc) things to the compiler is generally a good course to take.

abreslav:
> My remark about @Component is also about a similar concern, we have a lot of different @Component annotations is Spring projects, and users can define their own. So the ability to detect any kind of indirectly annotated @Component would make such support much more reliable.

This is an important refinement of the initial requirements, we should probably think a little more about it.

> If the chosen solution is a compiler plugin, what would be the requirement for us? Some kind of org.springframework.kotlin.CompilerPlugin class parameter added to the Gradle or Maven plugin configuration? Do you plan to bundle such class with Maven/Gradle Kotlin plugin in order to avoid requiring another plugin or do you expect us to provide it as an additional dependency?

I'd imagine that it should be something like `apply plugin: "kotlin-spring"` for Gradle, and not exactly sure what we can do for Maven, but we should definitely look more into it. Adding a `<kotlin-spring/>` tag somewhere would be ideal, IMO.

sdeleuze:
@abreslav Is that still expected to be available in Kotlin 1.1 ?

cy6erGn0m:
btw, solution "META-INF file` also provides ability to specify nullable/not-nullable annotation types. Some frameworks (for example vert.x) use their own annotations so they can supply such meta-inf file for better Kotlin compatibility

Writtscher:
@abreslav  Any estimation when we'll have this _..fixed''_? I'd really like to apply shiny kotlin syntax to my projects. But  it's a showstoppper for now as we are heavily using spring..

abreslav:
We are currently working on a prototype with the following properties:
- there's a compiler plugin that takes a list of "all-open" annotation class names,
- annotated classes and their members and their descendants become open,
- annotated annotations are added to the "all-open" list.

@sdeleuze @Writtscher This plugin will be available from the command line, and in build scripts we are planning to have plugins that configure a particular framework with more or less a one-liner, like

```
apply plugin: 'kotlin-spring'
```

If everything works out alright, this will ship in 1.0.x soon enough (long before 1.1).

We also plan to do the same for no-arg constructors for JPA and other serialization frameworks.

I'll update the KEEPs accrodingly next week.

sdeleuze:
That's awesome news @abreslav thanks a lot!!!

nilsschmidthamburg:
Brilliant work, thanks a lot!
Can't wait to use it. Especially when migrating big java projects to kotlin (automatically) this would avoid many problems. :-)

pbartoszek:
Can I check if the proposed solution will work for tests run directly from IntelliJ? For instance for spring integration tests? As far as I know IntelliJ uses Kotlin compiler directly bypassing maven/gradle

cypressious:
@pbartoszek According to https://blog.jetbrains.com/idea/2016/10/intellij-idea-2016-3-eap-refactorings-to-java-8-vcs-integration-and-gradle/ this won't be an issue in IDEA 2016.3

abreslav:
@pbartoszek The IDE will run everything correctly after importing the settings from Gradle (as it routinely does for dependencies)

pbartoszek:
It looks like it's time to move from maven to Gradle :)

abreslav:
@pbartoszek We are planning to have the same support for Maven as well

kenyee:
Looking forward to the plugin as well... Definitely one of the reasons I've been holding off on trying kotlin for a spring project 😁

oluwasayo:
@abreslav `apply plugin: 'kotlin-spring'` makes me cringe a bit. This is a problem faced by a wider dev community (not just spring). We're heavily invested in CDI/Weld and this is a pain point for us too. Meanwhile, with CDI when bean discovery mode is set to `all` in `beans.xml`, it is possible to not annotate the bean class at all and the CDI container treats it as `@Default`.

yanex:
@oluwasayo There is a generic plugin that uses the list of "all-open" annotations provided by the user, as well as some plugins for the popular frameworks (such as "kotlin-spring").
Also, it is easy to write the new Gradle/Maven plugin (as an "extension" for the generic one) for any other framework.

oluwasayo:
@yanex Awesome!

pbartoszek:
Guys, do you have any update on plugin?

yanex:
@pbartoszek We are planning to release it as a part of Kotlin 1.0.6 in the middle of December.

eygraber:
Any chance there will be an option to open all classes?

yanex:
@eygraber Why do you need this?

eygraber:
@yanex If I want to mock a lot of classes, but I don't want to annotate each one of them.

angryziber:
@yanex - try mocking with Mockito 2.2 with *mock-maker-inline* enabled - works like a charm with closed by default Kotlin classes and methods.

eygraber:
@angryziber it doesn't work on Android devices

yanex:
@eygraber Why don't make all mockable classes implement some interface with the all-open annotation on it?

eygraber:
@yanex I have a large number of complex classes that I would need to mock, and I don't like the idea of adding an annotation (or interface) that exists only for testing. It's much easier to just say that for testing everything should be open.

cypressious:
Isn't there an internal Kotlin-specific annotation that you could abuse for
this?

Kirill Rakhman

Am 27.12.2016 8:18 nachm. schrieb "Eliezer Graber" <notifications@github.com

yole:
The appropriate solution for allowing all classes to be mocked in tests is to use a custom JUnit test runner that removes the `final` modifier using bytecode processing. We plan to include such a runner in the `kotlin-test` library.

eygraber:
@yole again that would cause issues on Android because running tests on Android uses a custom runner. I tried creating an Android [transform](http://tools.android.com/tech-docs/new-build-system/transform-api) in a gradle Plugin a while back, but I wasn't able to get it to work well ([no_more_finals_plugin](https://github.com/eygraber/no_more_finals_gradle)).

I'd say this issue on Android will most likely need to be addressed differently unless there's a method to remove final from all classes.

eygraber:
Another issue with Android is that whatever is done needs to happen at compilation time before dexing, otherwise typical bytecode manipulation won't work (because it's dex not Java bytecode).

folkyatina:
Maybe it is not right place to ask, but is it possible to use kotlin-allopen for android anyhow?

PR #41, opened 2016-07-27
API for BigInteger and BigDecimal

voddan:


ilya-g:
Could you explain in the proposal comments on operations like `// optimize for TEN^n`?

`toInt()`, `toShort()` etc are already available as members of `BigInteger` and `BigDecimal`.

Regarding questions:

> Converting `Double.POSITIVE_INFINITY`, `Double.NEGATIVE_INFINITY` into `BigDecimal`

Options 2 and 3 are ruled out by the definintion of infinity.

> Exception in `BigDecimal. divide(BigDecimal)`

This question has already a decision, see [KT-10462](https://youtrack.jetbrains.com/issue/KT-10462) (tldr: we're using HALF_EVEN rounding)

proposals/stdlib/long-ariphmetics.md
```diff
@@ -0,0 +1,155 @@
...
+* The JDK API for operations on `BigInteger` and `BigDecimal`
+* Converting basic numeric types to and from "big" types like `1.toBigInteger()` or `2.toBig()` or possibly `3.big`
+
+## Dependencies
+
+* JDK6 `BigInteger` and `BigDecimal`
+
+## Placement
+
+* Standard Library, `java.math`, since there is already a part of the prosed API for [BigInteger](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/java.math.-big-integer/) and [BigDecimal](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/java.math.-big-decimal/)
```
ilya-g: Currently implemented part of API is in `kotlin` package: https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/src/kotlin/util/BigNumbers.kt

voddan:
> Could you explain in the proposal comments on operations like // optimize for TEN^n?

For some reason I thought that `BigInteger` used decimal system internally and hoped to optimize multiplication / division for powers of 10 using shift operations.

Since this assumption is (obviously) not true, the optimisation cannot be done. Thank you for pointing that out.

voddan:
> toInt(), toShort() etc are already available as members of BigInteger and BigDecimal.

@ilya-g I was unable to find those methods in JDK or in the Kotlin stdlib. JDK has similar methods like `intValue`, but the naming scheme does not mach the kotlin's `toInt`, so method-aliases are needed.

ilya-g:
> I was unable to find those methods in JDK or in the Kotlin stdlib.

Don't find, just try to invoke them. It's a little kotlin magic, similar to how can you have `size` property in collections that only have `size()` function in java.

Actually these functions are defined in [`kotlin.Number`](https://github.com/JetBrains/kotlin/blob/master/core/builtins/native/kotlin/Number.kt) interface, and `BigInteger`/`BigDecimal` implement `kotlin.Number`

ilya-g:
ok, seems it's ready to merge, I'll squash all commits and rename file to `bignumber-operations.md`.

voddan:
Cool, thanks!

Will you be able to be a "Shepherd" for that one?

ilya-g:
I've merged it and created issue #49 for discussion.

voddan:
Thanks!

PR #42, opened 2016-08-10
Fix small typo in code.

hastebrot:
Code should be `Map<out K, V, M : MutableMap<in K, in V>>.toMap(destination: M): M`.


ilya-g:
Actually this was intentionally written in such "pseudocode" in order not to list generic parameters and their constraints. `M <: MutableMap<in K, in V>` means some subtype `M` of `MutableMap`, and that `M` is being returned back from the function.

Btw, the new version also isn't correct: `Map` type has only two type parameters.

ilya-g:
So I'm closing this PR for now. If you'd have better idea how to express the signature, let me know.

Issue #5, opened 2016-04-20
Bound callable references


abreslav:
Discussions about the [Bound callable references proposal](../blob/master/proposals/bound-callable-references.md) will be held here.

@udalov is the _shepherd_ for this proposal.
### Short description

**Bound callable references** are expression like `foo::bar` where `foo` is itself an expression, not a class. A type of `foo::toString` would be `() -> String` for any `foo`. 

See [full proposal](../blob/master/proposals/bound-callable-references.md) for more details

b005t3r:
Just to clarify:
```
class C {
    fun foo() {}
}

fun test(c: C?) {
    c::foo             // error
    c::class           // error
    null::class        // error
}
```

But this would still work, right?
```
class C {
    fun foo() {}
}

fun test(c: C?) {
    c?::foo             // error
    c?::class           // error
    null?::class        // error
}
```

> API for "unbinding" a reference

I'd rather go for API for getting an unbound reference from a bound reference (and vice versa).

> if unbound already, throw or return null, or provide both?

Of course there would be a way to check if it's bound/unbound, right?

udalov:
> But this would still work, right?

Currently this is not planned, but it's a valid option.

> I'd rather go for API for getting an unbound reference from a bound reference (and vice versa).
> Of course there would be a way to check if it's bound/unbound, right?

This is exactly what that open question is about. It's not clear at this point whether such API is needed and what declarations would be available there. I'll elaborate on this in the proposal.

damianw:
I noticed that re-introducing type inference for an empty LHS isn't listed as a "possible future advancement". Has this been considered? i.e.:
```Kotlin
val foo = listOf<String>()
foo.filter(::isEmpty) // String::isEmpty is inferred for arguments of type String.() -> Boolean or (String) -> Boolean
```

It's not specifically related to bound references, but as I recall, it was removed to reserve the syntax for an implicit `this` for bound references.

I wouldn't mind having to be explicit to resolve ambiguities with `this`, but I'd take an implicit type on the LHS rather than `this` any day. Having to specify `SomeComplexType::isFrobnicated` is often pretty inconvenient... almost as much as using a lambda when there are many arguments.

b005t3r:
>> But this would still work, right?
> Currently this is not planned, but it's a valid option.

Hmm, I think this would be really limiting and lead to unnecessary tricks (to get the non-nullable reference) in many normally valid scenarios. Like instead of being able to do this:
```event += obj::handler```
[...]

PR #43, opened 2016-08-10
Reword callable reference and class literal resolution algorithm, explain some corner cases (#5)

udalov:


proposals/bound-callable-references.md
```diff
@@ -123,20 +131,61 @@ fun test() {
...
+
+```
+class C {
+    companion object
+}
+
+fun test() {
+    C::class            // class of C
+    C()::class          // class of C
+    (C)::foo            // class of C.Companion
```
cypressious: shouldn't this be `::class`?

udalov: Fixed, thanks

proposals/bound-callable-references.md
```diff
@@ -82,13 +86,17 @@ The LHS may now be interpreted as an expression, or a type, or both.
...
-The algorithm is the following:
+The semantics are different when the LHS is interpreted as an expression or a type, so we introduce the following algorithm to choose the resulting interpretation when both are applicable:
 
-1. Try interpreting the LHS as an **expression** with the usual resolution algorithm for qualified expressions.
-   If the result represents a companion object of some class, continue to p.2.
-   Otherwise continue resolution of the member in the scope of the expression's type.
-2. Resolve the unbound reference with the existing algorithm.
+1. Type-check the LHS as an **expression**.
+   - If the result represents a _companion object of some class_ specified with the short syntax (via the short/qualified name of the containing class), discard the result and continue to p.2.
+   - If the result represents an _object_ (either non-companion, or companion specified with the full syntax: `org.foo.Bar.Companion` or just `Bar.Companion`), remember the result and continue to p.2.
```
erokhins: or `(Bar)`

proposals/bound-callable-references.md
```diff
@@ -82,13 +86,17 @@ The LHS may now be interpreted as an expression, or a type, or both.
...
-The semantics are different when the LHS is interpreted as an expression or a type, so we establish a priority of one over another when both interpretations are applicable.
-The algorithm is the following:
+The semantics are different when the LHS is interpreted as an expression or a type, so we introduce the following algorithm to choose the resulting interpretation when both are applicable:
 
-1. Try interpreting the LHS as an **expression** with the usual resolution algorithm for qualified expressions.
-   If the result represents a companion object of some class, continue to p.2.
-   Otherwise continue resolution of the member in the scope of the expression's type.
-2. Resolve the unbound reference with the existing algorithm.
+1. Type-check the LHS as an **expression**.
+   - If the result represents a _companion object of some class_ specified with the short syntax (via the short/qualified name of the containing class), discard the result and continue to p.2.
```
erokhins: represents a companion object **or object**. And what about enum entries?

proposals/bound-callable-references.md
```diff
@@ -123,20 +131,61 @@ fun test() {
...
+Resolution of the LHS of a class literal expression is performed with the same algorithm.
+
+```
+class C {
+    companion object
+}
+
+fun test() {
+    C::class            // class of C
+    C()::class          // class of C
```
erokhins: `С()` class literal for instance?

proposals/bound-callable-references.md
```diff
@@ -123,20 +131,61 @@ fun test() {
...
+    (C)::class          // class of C.Companion
+    C.Companion::class  // class of C.Companion
+}
+```
+
+Once the LHS is type-checked and its type is determined to be `T`, the type of the whole class literal expression is `kotlin.reflect.KClass<T'>` where `T'` is a type obtained by _substituting `T`'s arguments with star projections (`*`)_. Example:
```
erokhins: What about `Array<String>::class`?

proposals/bound-callable-references.md
udalov: No, `(Bar)` is covered by the first clause because it's a companion object of some class specified with the _short_ syntax

proposals/bound-callable-references.md
udalov: Why do you think non-companion objects are relevant here? They are handled in the second clause

proposals/bound-callable-references.md
udalov: Yes, `arrayOfStrings::class` should have type `KClass<Array<String>>`. Thanks, I'll add this here

proposals/bound-callable-references.md
erokhins: What about `object Foo` `(Foo)`?

udalov: I'm not sure what do you propose, could you elaborate? The two examples here explain what does "companion specified with the full syntax" mean, they do not relate to non-companion objects. I don't think usages of a non-companion object needs any explanation. Or are you talking about the `(Foo)::class` case?

proposals/bound-callable-references.md
```diff
@@ -123,20 +131,61 @@ fun test() {
...
+```
+fun test() {
+    "a"::class               // KClass<String>
+    listOf("a")::class       // KClass<List<*>> (not "KClass<List<String>>"!)
```
abreslav: I'd add another example here: `listOf(listOf("a"))::class // KClass<List<*>>`

Otherwise, the reader might think that we only erase the innermost argument (I did think that for a second)... But maybe it's only me

abreslav:
No other comments from me

udalov:
@abreslav @erokhins I've updated the class literal type checking part with a more precise definition of what happens with expressions of type `kotlin.Array<...>`. Please review, I'm not sure I've formulated it well enough

proposals/bound-callable-references.md
```diff
@@ -162,12 +162,24 @@ fun test() {
 }
 ```
 
-Once the LHS is type-checked and its type is determined to be `T`, the type of the whole class literal expression is `kotlin.reflect.KClass<T'>` where `T'` is a type obtained by _substituting `T`'s arguments with star projections (`*`)_. Example:
+Once the LHS is type-checked and its type is determined to be `T`, the type of the whole class literal expression is `KClass<erase(T)>`, where `erase(T)` represents the most accurate possible type-safe representation of the type `T` and is obtained by substituting `T`'s arguments with star projections (`*`), except when the type is an array.
```
abreslav: Type-arguments of arrays are not fully known at runtime either: their nullability is unclear. Does it matter in this context?

udalov: It does matter, and we may revisit it in the future, but type-checking `arrayOfStrings::class` to `KClass<Array<*>>` would be rather unhelpful from Java user's point of view. We've briefly discussed this with @erokhins and decided (for now) to type-check `arrayOfStrings::class` to `KClass<Array<String>>`. We should discuss this in detail later anyway.

bashor: Note in JS we don't know anything about type-arguments of arrays at runtime.

udalov:
Added a TODO about JS, we'll have to discuss later. I'm merging this set of changes now, thanks to everyone for review!

PR #44, opened 2016-08-12
Create peek proposal

cbruegg:


ilya-g:
This proposal looks good and shows some research effort. :+1:

I'd like to merge it, so we could then discuss the details, but I suggest to elaborate use cases before.
  1. I'm not sure, how the mentioned `PeekStream` relates to this proposal.
  2. Could you show several of "many more" usage examples, excluding already given `peek { println(it) }`?

cbruegg:
@ilya-g: Thanks for the feedback! I've updated the use case section with some more examples and removed the `PeekStream`, which indeed wasn't an example for what is being proposed here.

I've also moved the proposal to the `stdlib` folder.

ilya-g:
Thanks, I've merged the proposal and opened an issue #47 for the discussion.

PR #50, opened 2016-09-22
Create moveXXX() functions for MutableList

thomasnield:
A proposal to implement `moveXXX()` operators for `MutableList` to manage the movement of items to different indexes.

### DIFF
--- a/proposals/stdlib/mutablelist-move.md
+++ b/proposals/stdlib/mutablelist-move.md
@@ -37,6 +37,9 @@ Cannot think of an explicit precendent in another API.
 Developers could implement the extension functions themselves, as shown below. Moving items can be tricky and verbose to coordinate with local implementations.
 
 ```kotlin
+
+import java.util.*
+
 /**
  * Moves the given **T** item to the specified index
  */
@@ -50,22 +53,24 @@ fun <T> MutableList<T>.move(item: T, newIndex: Int)  {
 /**
  * Moves the given item at the `oldIndex` to the `newIndex`
  */
-fun <T> MutableList<T>.move(oldIndex: Int, newIndex: Int)  {
+fun <T> MutableList<T>.moveAt(oldIndex: Int, newIndex: Int)  {
     val item = this[oldIndex]
     removeAt(oldIndex)
-    add(newIndex, item)
+    if (oldIndex > newIndex)
+        add(newIndex, item)
+    else
+        add(newIndex - 1, item)
 }
 
 /**
  * Moves all items meeting a predicate to the given index
  */
 fun <T> MutableList<T>.moveAll(newIndex: Int, predicate: (T) -> Boolean) {
-
-    var newIndexIncrement = newIndex
-
-    asSequence().toList().asSequence().withIndex()
-            .filter { v -> predicate.invoke(v.value) }
-            .forEach { move(it.index, newIndexIncrement++) }
+    check(newIndex >= 0 && newIndex < size)
+    val split = partition(predicate)
+    clear()
+    addAll(split.second)
+    addAll(if (newIndex >= size) size else newIndex,split.first)
 }
 
 /**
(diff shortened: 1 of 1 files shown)


proposals/stdlib/mutablelist-move.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Summary
+
+It would be helpful to create some additional extension functions for `MutableList` to move items to new indexes, as well as move them up/down, and one or all items that meet a predicate.
+
+These can be implemented using only extension functions on `MutableList`.
+
+## Similar API review
+
+Cannot think of an explicit precendent in another API.
```
thomasnield: I figured there was something out there. Not surprised the `Collections` helper static class has a method, although I've never used it. I will investigate those.

proposals/stdlib/mutablelist-move.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Alternatives
+
+Developers could implement the extension functions themselves, as shown below. Moving items can be tricky and verbose to coordinate with local implementations.
+
+```kotlin
+/**
+ * Moves the given **T** item to the specified index
+ */
+fun <T> MutableList<T>.move(item: T, newIndex: Int)  {
```
thomasnield: Thought I accounted for that, apparently not!

proposals/stdlib/mutablelist-move.md
```diff
@@ -0,0 +1,145 @@
...
+    add(newIndex, item)
+}
+
+/**
+ * Moves the given item at the `oldIndex` to the `newIndex`
+ */
+fun <T> MutableList<T>.move(oldIndex: Int, newIndex: Int)  {
+    val item = this[oldIndex]
+    removeAt(oldIndex)
+    add(newIndex, item)
```
thomasnield: I may need to rename this to `moveAt()` or something similar so the overload doesn't clash if its a `MutableList<Int>`. Is that what you were thinking with previous comment?

Good catch on inconsistent state. It crossed my mind but I reasoned incorrectly as my thought experiments thought of the item only moving up, not down. I may need to do a check to see if an offset occurs and adjust by `1` accordingly.

proposals/stdlib/mutablelist-move.md
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/**
+ * Moves all items meeting a predicate to the given index
+ */
+fun <T> MutableList<T>.moveAll(newIndex: Int, predicate: (T) -> Boolean) {
+
+    var newIndexIncrement = newIndex
+
+    asSequence().toList().asSequence().withIndex()
```
thomasnield: Yup. That's valid.

proposals/stdlib/mutablelist-move.md
```diff
@@ -0,0 +1,145 @@
...
+            .filter { v -> predicate.invoke(v.value) }
+            .forEach { move(it.index, newIndexIncrement++) }
```
thomasnield: My tests didn't see an issue. I think I know what you're saying, and I'm trying to think of what could go wrong here... but I can't think of anything.

proposals/stdlib/mutablelist-move.md
```diff
@@ -0,0 +1,145 @@
...
+    add(newIndex, item)
+}
+
+/**
+ * Moves the given element **T** up the **MutableList** by an index increment
+ * unless it is at the bottom already which will result in no movement
+ */
+fun <T> MutableList<T>.moveDown(item: T) {
+    val currentIndex = indexOf(item)
+    val newIndex = (currentIndex + 1)
```
thomasnield: So maybe a `false` needs to be returned to indicate a failure, or even throw an `Exception`.

proposals/stdlib/mutablelist-move.md
```diff
@@ -0,0 +1,145 @@
...
+The proposal is to put this in Standard Library, and make it immediately available wherever `MutableList` is used.
+
+## Reference implementation
+
+A proposed implementation can be [found in this repository](https://github.com/thomasnield/kotlin-collection-extras/blob/master/src/main/kotlin/MutableListMoveExt.kt).
+
+
+## Unresolved questions
+
+* Is there a way to move items without removing and adding them? (Unlikely to my knowledge).
```
thomasnield: Ok.

thomasnield:
I've use some variants of these in an application I'm building, and this is an API I've written from scratch in an attempt to formalize and fill in the gaps. I agree it is not polished. I'll make some changes this weekend and may ask for clarification on a few things, and thanks for taking the time to review this.

proposals/stdlib/mutablelist-move.md
thomasnield: It looks like `Collections.rotate()` simply shifts the entire `List` up or down for all elements in their current order. It does not move elements relative to each other.  When I specify a condition, I am moving elements with that condition to a certain area in the `List`. 

`subList()` looks only to extract part of the `List` between two indexes. If there is a way to make it work in conjunction with `Collections.rotate()`, I'm not sure what implementation you are envisioning.

ilya-g: I mean something like this:
```
fun <T> MutableList<T>.move(fromIndex: Int, toIndex: Int) {
    if (toIndex > fromIndex) {
        Collections.rotate(subList(fromIndex, toIndex + 1), -1)    
    } else if (toIndex < fromIndex) {
        Collections.rotate(subList(toIndex, fromIndex + 1), 1)
    }
}
```

proposals/stdlib/mutablelist-move.md
```diff
@@ -0,0 +1,145 @@
...
+fun <T> MutableList<T>.move(oldIndex: Int, newIndex: Int)  {
+    val item = this[oldIndex]
+    removeAt(oldIndex)
+    add(newIndex, item)
+}
+
+/**
+ * Moves all items meeting a predicate to the given index
+ */
+fun <T> MutableList<T>.moveAll(newIndex: Int, predicate: (T) -> Boolean) {
```
ilya-g: I'd also consider as an alternative partitioning the list by predicate and joining the parts again:
```
val reordered = list.partition(predicate).let { it.first + it.second }
```

proposals/stdlib/mutablelist-move.md
ilya-g: Imagine you have items with indices 0 and 1 matching the predicate and you want to move them to the end. First you move an item with index 0 to the end and all other items are moved to the beginning. Then you try to move an item with index 1, but it's no more that item that was matched by predicate.

thomasnield:
Started making edits on my end. Will push in the next day or two after I finish.

thomasnield:
Sorry for the delay with this. I'm doing an O'Reilly webcast next week. After that, I'll follow up as my colleagues and I do want this in the stdlib.

thomasnield:
Please refer to this file to see changes. Can you leave any further comments here?
https://github.com/thomasnield/kotlin-collection-extras/blob/master/src/main/kotlin/MutableListMoveExt.kt

thomasnield:
I implemented the proposed changes (I think). Please let me know if I missed anything.

ilya-g:
I still believe these `move*` extensions should not belong to stdlib. 

Given their use cases, the proposed implementations better fit for more scoped UI library operating on `ObservableList` from JavaFX.

Regarding `MutableList.swap` extension, it is considered to be added in https://youtrack.jetbrains.com/issue/KT-2460.

thomasnield:
Fair enough, that sounds valid. I'll track thatbissue and thanks for considering!

PR #54, opened 2016-10-19
Underscores in Numeric Literals

cypressious:


voddan:
What is the reason to keep that case valid?

    val x4 = 5_______2        // OK (decimal literal)

I would prefer the most strict rules, e.i. prohibiting double underscores

adi1133:
I suggest allowing double underscores. 
If people want to write unreadable code, they will always find a way. 
```java
int foo = 5_______2;
```
 is currently valid java 8 code.

PR #68, opened 2017-04-17
Create props

ex3ndr:


damianw:
Have a look at [the existing ideas for `dataarg`](https://youtrack.jetbrains.com/issue/KT-15471) -- it is another approach with similar goals.

ex3ndr:
I don't think that this is the same. We are trying to avoid declare multiple classes and doing hacks for helping type resolving. While our problem can't be easily solved, dataargs can be easily emulated via optional named arguments.

nikitabobko:
Closing stale PR

PR #71, opened 2017-06-09
Initial proposal for Parcelable support

abreslav:


proposals/extensions/android-parcelable.md
```diff
@@ -0,0 +1,254 @@
...
+   - it results in a compilation error  
+   - Question: maybe overriding `describeContents()` is OK?
+
+The annotations for Parcelable classes, transient fields, etc. should sit in a complimentary runtime library that ships together with the compiler extension.  
+
+TODO: Check interactions with inheritance by delegation.
+  
+Some syntactic options:    
+- The names `Parcelize` and `Transient` are just placeholders for now.
+- We could allow the user to omit teh supertype `Parcelable`, to be more DRY and have only the annotation to signify that the class is Parcelable, btu this would be more challenging wrt the tooling support.
```
JakeWharton: nit: teh --> the

JakeWharton: nit: btu --> but

proposals/extensions/android-parcelable.md
```diff
@@ -0,0 +1,254 @@
...
+- the `newArray()` method of the companion object is implemented automatically unless it's explicitly overridden
+- the type argument to `Parceler` must match the containing class, otherwise it's a compiler error
+
+Note: Indirect implementations (`object : Foo`, where `Foo : Parceler`) are questionable here, but we can allow them if there are use cases for it.
+
+Syntactic options:
+- It does not have to be a companion object, a named object, e.g. `object Parceler: Parceler<MyPercelable>` may be ok too
+- We may want to require the object to be annotated to explicitly show that the `newArray()` is auto-generated
+
+> Discussion: why not implement `newArray()` in teh `Parceler` interface itself?
```
JakeWharton: nit: teh --> the

proposals/extensions/android-parcelable.md
```diff
@@ -0,0 +1,254 @@
...
+### Simple case: completely automatic Parcelable
+
+A compiler extension can generate serialization/deserialization logic for all properties in a primary constructor of a class marked with a special annotation:
+
+```kotlin
+@Parcelize
+class MyParcelable(val data: Int): Parcelable
+```  
+
+Note that the class is not required to be a data class.
```
JakeWharton: But cannot be an `object`?

proposals/extensions/android-parcelable.md
```diff
@@ -0,0 +1,254 @@
...
+
+If a `@Parcelize` class has a companion object that implements `Parceler`:
+- the necessary boilerplate is generated for the `Parcelable` convention and delegated to the companion object
+- the `newArray()` method of the companion object is implemented automatically unless it's explicitly overridden
+- the type argument to `Parceler` must match the containing class, otherwise it's a compiler error
+
+Note: Indirect implementations (`object : Foo`, where `Foo : Parceler`) are questionable here, but we can allow them if there are use cases for it.
+
+Syntactic options:
+- It does not have to be a companion object, a named object, e.g. `object Parceler: Parceler<MyPercelable>` may be ok too
```
orangy: I think it is even _undesirable_ to implement it as companion object, because `MyParcelable.newArray` would be accessible. It's probably better has to be private object.

proposals/extensions/android-parcelable.md
orangy: I think it should be instantiatable entity, i.e. not an abstract class, not interface, not enum, should have accessible constructor, etc.

proposals/extensions/android-parcelable.md
```diff
@@ -0,0 +1,254 @@
...
+1. Java annotation processors can't alter Kotlin code, 
+2. Normally, annotation processors can't add methods to existing classes (this is possible though a private API only),
+3. A compiler extension can be potentially more flexible in terms of suppressing errors and providing syntactic means to the user.  
+
+### Simple case: completely automatic Parcelable
+
+A compiler extension can generate serialization/deserialization logic for all properties in a primary constructor of a class marked with a special annotation:
+
+```kotlin
+@Parcelize
```
EmmanuelVinas: Is there really a need for an extra annotation? why not a mapped type ?
`class MyParcelable(val data:Int) : kotlin.Parcelable `

proposals/extensions/android-parcelable.md
```diff
@@ -0,0 +1,254 @@
...
+
+The annotations for Parcelable classes, transient fields, etc. should sit in a complimentary runtime library that ships together with the compiler extension.  
+
+TODO: Check interactions with inheritance by delegation.
+  
+Some syntactic options:    
+- The names `Parcelize` and `Transient` are just placeholders for now.
+- We could allow the user to omit teh supertype `Parcelable`, to be more DRY and have only the annotation to signify that the class is Parcelable, btu this would be more challenging wrt the tooling support.
+- We could annotate the supertype itself to make it more local: 
+  - `class MyParcelable(val data: Int): @Auto Parcelable`
```
orangy: Really like the idea of annotating an interface, it can be extended to other use cases.

proposals/extensions/android-parcelable.md
yanex: Likely yes, just because all serialized values must be mentioned in the primary constructor, and objects don't have it.

By the way, are there any cases when the `object` support would be desirable?

proposals/extensions/android-parcelable.md
```diff
@@ -0,0 +1,254 @@
...
+- The class annotated with `@Parcelize` must implement `Parcelable` (directly or through a chain of supertypes)
+  - Otherwise it's a compiler error
+- All parameters of the primary constructor must be properties, otherwise they can not be (de)serialized
+  - If there is a non-property parameter, it is a compilation error
+- Properties with initializers declared in the class body are also difficult to deserialize correctly: we'd have to generate an alternative constructor that does not execute initializers (including `init` blocks) at all and only uses the serialized data, but this has all the issues that `java.io.Serializable` has wrt "magic" object creation.
+  - Properties in the class body must be marked with a `@Transient` annotation, otherwise it's a compiler warning (or error?) 
+- If some properties are not parcelable (i.e. do not implement `Parcelable`, are not supported by the `Parcel` interface, and are not customized (see below)), it's a compilation error
+- The user is not allowed to manually override methods of `Parcelable` or create the `CREATOR` field
+   - it results in a compilation error  
+   - Question: maybe overriding `describeContents()` is OK?
```
kevinmost: In what situation would a user be expected to manually implement `describeContents()`? Isn't this method mainly used to denote whether you're parceling a `FileDescriptor`, which the compiler should be able to determine?

proposals/extensions/android-parcelable.md
damianw: I think that might be too much magic. There's no indication that this type in particular is going to be treated specially... at least with annotations, synthetic properties, etc there is a hint that "the compiler is doing some magic here".

proposals/extensions/android-parcelable.md
adelnizamutdinov: hierarchies with sealed classes, maybe

damianw: @yanex In the past I have used both classes and objects inheriting from sealed classes as "smarter" enums. `Parcelable` support on `object`s is similar to `Serializable` support... [relevant issue](https://youtrack.jetbrains.com/issue/KT-9499). I think in that case the `readResolve` solution is appropriate... and a similar approach can be taken with `Parcelable`. Though... one or the other would be fine if not both.
```Kotlin
sealed class Maybe<out T : Parcelable> : Parcelable {

  override fun describeContents(): Int = 0

  object None : Maybe<Nothing>(), Parcelable.Creator<None> {
    override fun writeToParcel(dest: Parcel, flags: Int) {}
    override fun newArray(size: Int) = arrayOfNulls<None>(size)
    override fun createFromParcel(source: Parcel) = this
    @JvmField val CREATOR = this
  }
  
  class Some<out T : Parcelable>(val value: T) : Maybe<T>() {
    override fun writeToParcel(dest: Parcel, flags: Int) {
      dest.writeParcelable(value, 0)
    }
    companion object {
      @JvmField val CREATOR = // ...
    }
  }

}
```

proposals/extensions/android-parcelable.md
```diff
@@ -0,0 +1,254 @@
...
+
+The annotations for Parcelable classes, transient fields, etc. should sit in a complimentary runtime library that ships together with the compiler extension.  
+
+TODO: Check interactions with inheritance by delegation.
+  
+Some syntactic options:    
+- The names `Parcelize` and `Transient` are just placeholders for now.
+- We could allow the user to omit the supertype `Parcelable`, to be more DRY and have only the annotation to signify that the class is Parcelable, but this would be more challenging wrt the tooling support.
+- We could annotate the supertype itself to make it more local: 
+  - `class MyParcelable(val data: Int): @Auto Parcelable`
```
damianw: I think I like this more, if in the future this could be applied to other compiler plugins that can implement other interfaces automatically.

proposals/extensions/android-parcelable.md
```diff
@@ -0,0 +1,254 @@
...
+```
+
+Local customization is more flexible, but will likely result in a lot of duplication.
+Global customization is problematic wrt inceremental and separate compilation, but this can be addressed in the future.  
+
+## Handling hierarchies of Parcelable classes
+
+Hierarchies of Parcelable classes are challenging from two points of view:
+- organizing constructors (see above)
+- deserializing an instance of the right class from a Parcel, when the container has only a reference to a superclass.
```
damianw: Can you elaborate on this problem? My understanding is that this is already done automatically by the `Parcel` itself, and it will correctly choose the appropriate `CREATOR` from the exact class (and never a superclass)... unless a caller manually specifies a `Creator` such as in `readTypedList`.

proposals/extensions/android-parcelable.md
Takhion: `enum`/`object` should definitely be supported, although they can't be instantiated

proposals/extensions/android-parcelable.md
Takhion: It could be a type alias, that way the compiler would be able to distinguish but it would still map to `Parcelable` under the hood

proposals/extensions/android-parcelable.md
Takhion: Me too, I didn't even know it was possible!

loganj:
There's some prior art that's worth a look (or at least a reference): nsk-mironov/smuggler

Takhion:
@loganj technically smuggler is not a kotlinc plugin: it's using the [android transform api](http://tools.android.com/tech-docs/new-build-system/transform-api)

loganj:
@takhion oops, thanks

proposals/extensions/android-parcelable.md
elizarov: Sealed class hierarchies can be technically supported, indeed, including objects.

proposals/extensions/android-parcelable.md
elizarov: Enum support is very, very convenient to have, though. Without enums it is going to be too limiting in practise. My proposal, though, is not to require tagging enums with `Parcelzie` but to support all enums out-of-the box.

proposals/extensions/android-parcelable.md
yanex: I think that's a bit less visible to the reader, especially when the primary constructor accepts a long list of parameters.
But nothing is final right now, we can try several variants.

rharter: I think this is less desirable for readability since using this functionality (as directed by the `@Auto` annotation in this case) has implications for the primary constructor.  I would expect that since the annotation has implications like no params in primary constructor, only Parcelable types, for the class itself, not just the `Parcelable` interface, the annotation should go on the class.

proposals/extensions/android-parcelable.md
loganj: For an instance where it's desirable:

We use `object`s to represent some states in what's essentially a state machine (other states are data classes). We keep track of a history of states, and we persist that history in a Parcel.

proposals/extensions/android-parcelable.md
Takhion: @rharter it really shouldn't be a problem:
1. if the line is so long the you won't notice the annotation before the interface, it should probably be split on multiple lines anyway
2. keep in mind that the IDE will potentially highlight everything when things are incompatible, so in that case the position of the annotation doesn't really matter

proposals/extensions/android-parcelable.md
```diff
@@ -0,0 +1,262 @@
...
+
+The latter issue seems to be resolvable by "tagging" such records through writing fully-qualified names of concrete classes to parcels, but it's not clear whether it's a good idea, since
+- it involves reflection
+- it will increase the size of parcels
+
+This proposal does not support hierarchies for now.
```
loganj: If a hierarchy is detected, will that be a compile error? Would be nice.

proposals/extensions/android-parcelable.md
Takhion: This annotation is strictly tied to implementing the `Parcelable` interface: it's like a contract that can't be split, and so the 2 things should stay together. If the annotation was modifying the class in any other way, then placing it directly on the class would be OK.

proposals/extensions/android-parcelable.md
Takhion: If they are sealed hierarchies, could they be supported? All the information is available at compile time.

proposals/extensions/android-parcelable.md
rharter: It's not a matter of line length, it's that the annotation does have an effect on the class, outside of the Parcelable interface. The annotation (not the interface) adds requirements to the class (dealing with property types and constructor parameters) which are not inherited from the interface. Therefore it should be applied to the class since it's modifying that and not the interface, IMHO.

proposals/extensions/android-parcelable.md
yanex: @loganj Yes, of course.

yanex: @Takhion Well, the problem is that the primary constructor of the `@Parcelize` class should contain all significant data (all parameters of the primary constructor should be `val` or `var`). If your class hierarchy satisfies this requirement, it can be supported.

proposals/extensions/android-parcelable.md
```diff
@@ -0,0 +1,263 @@
...
+  - `@Parcelize` can't annotate an interface
+  - `sealed` and `enum` classes may be supported
+  - `object` are forbidden for now
+    - Objects can also be supported. We can serialize just the qualified name because we can't make the new instance of `object` anyway
+- The class annotated with `@Parcelize` must implement `Parcelable` (directly or through a chain of supertypes)
+  - Otherwise it's a compiler error
+- All parameters of the primary constructor must be properties, otherwise they can not be (de)serialized
+  - If there is a non-property parameter, it is a compilation error
+- Properties with initializers declared in the class body are also difficult to deserialize correctly: we'd have to generate an alternative constructor that does not execute initializers (including `init` blocks) at all and only uses the serialized data, but this has all the issues that `java.io.Serializable` has wrt "magic" object creation.
+  - Properties in the class body must be marked with a `@Transient` annotation, otherwise it's a compiler warning (or error?)
```
loganj: Warning would allow for calculated properties.

loganj: Error seems right. Could always relax to a warning if a major use case arises, easier than going the other way and will force discovery of that use case.

proposals/extensions/android-parcelable.md
```diff
@@ -0,0 +1,263 @@
...
+  - `object` are forbidden for now
+    - Objects can also be supported. We can serialize just the qualified name because we can't make the new instance of `object` anyway
+- The class annotated with `@Parcelize` must implement `Parcelable` (directly or through a chain of supertypes)
+  - Otherwise it's a compiler error
+- All parameters of the primary constructor must be properties, otherwise they can not be (de)serialized
+  - If there is a non-property parameter, it is a compilation error
+- Properties with initializers declared in the class body are also difficult to deserialize correctly: we'd have to generate an alternative constructor that does not execute initializers (including `init` blocks) at all and only uses the serialized data, but this has all the issues that `java.io.Serializable` has wrt "magic" object creation.
+  - Properties in the class body must be marked with a `@Transient` annotation, otherwise it's a compiler warning (or error?) 
+- If some properties are not parcelable (i.e. do not implement `Parcelable`, are not supported by the `Parcel` interface, and are not customized (see below)), it's a compilation error
```
loganj: nit: assume this means properties appearing in the constructor, but it's not stated explicitly.

proposals/extensions/android-parcelable.md
```diff
@@ -0,0 +1,263 @@
...
+   - Question: maybe overriding `describeContents()` is OK?
+     - Actually we can figure out is there any file descriptor serialized, and generate the appropriate `describeContents()` implementation
+
+The annotations for Parcelable classes, transient fields, etc. should sit in a complimentary runtime library that ships together with the compiler extension.  
+
+TODO: Check interactions with inheritance by delegation.
+  
+Some syntactic options:    
+- The names `Parcelize` and `Transient` are just placeholders for now.
+- We could allow the user to omit the supertype `Parcelable`, to be more DRY and have only the annotation to signify that the class is Parcelable, but this would be more challenging wrt the tooling support.
```
loganj: ... and less obvious to readers.

proposals/extensions/android-parcelable.md
GrahamBorland: No different from whatever "magic" happens with `Serializable`, really?

proposals/extensions/android-parcelable.md
veonua: it would be super nice to have DateParceler build in, many of apps use Date and copy-pasting same code is pretty ugly way

JakeWharton: Date is `Serializable` so it will already just work. I would be against any special handling otherwise.

veonua: unfortunately it doesn't 

 No virtual method writeSerializable(Ljava/util/Date;)V in class Landroid/os/Parcel; or its super classes (declaration of 'android.os.Parcel' appears in /system/framework/framework.jar)

gildor: @veonua This is just a bug and should be fixed in next Kotlin version:
https://youtrack.jetbrains.com/issue/KT-19749

PR #73, opened 2017-07-13
Proposal: lateinit property intrinsics

udalov:


proposals/lateinit-property-intrinsics.md
```diff
@@ -0,0 +1,224 @@
...
+    // Extract variable
+
+    val property = this::file
+    if (property.isInitialized) ...               // Error!
+
+    // Use .let
+
+    if (this::file.let { it.isInitialized }) ...  // Error!
+    ```
+* The names `isInitialized` and `reset` are somewhat asymmetric. Maybe we should rename `reset` to `uninitialize`, or rename `isInitialized` to `isSet`.
```
yanex: +1 for renaming `reset` to `uninitialize`.

proposals/lateinit-property-intrinsics.md
```diff
@@ -0,0 +1,224 @@
...
+    if (this::file.isInitialized) ...             // OK
+
+    // Extract variable
+
+    val property = this::file
+    if (property.isInitialized) ...               // Error!
+
+    // Use .let
+
+    if (this::file.let { it.isInitialized }) ...  // Error!
```
yanex: Also, looks like you can't `reset()` several `lateinit var`s at once:
`listOf(this::file, this::environment, this::other).all { it.reset() }`

Unlike `.let {}`, this scenario seems to me realistic.

udalov: I'm afraid there's no simple way to generate bytecode for that anyway.

elizarov:
Neither proposal nor the original issue provides enough motivation for `deinitialize` function. The code that is given as motivation for `deinitialize` does not actually need to be concerned with deinitialization at all.

Moreover, `deinitialize` is a very dangerous addition to the language and the proposal fails to mention this fact. Here is the problem. `lateinit` initialization checks are performed by the Kotlin compiler on every use of `lateinit` properties. Some code (esp in Android) is very heavy in `lateinit` properties and this causes some noticeable performance issues. Kotlin compiler could have been optimized to avoid emitting those check when initialization of `latent` property was checked at least once, but an introduction of `deinitialize` destroys the possibility of such optimization in the future.

udalov:
@elizarov Some test runners might keep a hard reference to the test even after it's completed. As far as I understand, we had this problem in our project (not sure if we still have it), which resulted in weird workarounds in multiple places like [this](https://github.com/JetBrains/kotlin/blob/b231bf0207c2e1fc2c24eb4aacd3634250cdf00b/compiler/tests-common/org/jetbrains/kotlin/codegen/CodegenTestCase.java#L255). I admit it's a weak argument, but I don't think it'd be very difficult to come up with a better use case.

Regarding optimizations, note that the backing field of a public lateinit property is public in the bytecode, which means it can be deinitialized easily in Java. So even today, we can only perform these optimizations for _private_ lateinit properties.

dzharkov:
Merged to master manually

PR #77, opened 2017-08-01
Scripting

ligee:


proposals/scripting-support.md
```diff
@@ -0,0 +1,434 @@
...
+
+Standalone scripting is a variant of the embedded scripting with hosts provided with the Kotlin distribution.
+
+#### Hosts
+
+The standalone script could be executed e.g. using command line Kotlin compiler:
+
+`kotlinc -script myscript.kts`
+
+Or with some hypothetical dedicated runner, that we might want to include into the distribution:
```
udalov: There's already a non-hypothetical dedicated runner called `kotlin` that we include into the distribution today, let's use it.

proposals/scripting-support.md
```diff
@@ -0,0 +1,434 @@
...
+
+`kotlinsh myscrtipt.ktsh`
```
udalov: This would be
```
kotlin myscript.kts
```

proposals/scripting-support.md
```diff
@@ -0,0 +1,434 @@
...
+
+To be able to use the Kotlin scripts in a Unix shell environment, the *shebang* (`#!`) syntax should be supported 
+at the beginning of the script:
+```
+#! /path/to/kotlin/script/runner -some -params
```
udalov: As an aside, note that there may be problems with passing multiple arguments in the shebang line: https://stackoverflow.com/a/4304187/288456

proposals/scripting-support.md
```diff
@@ -0,0 +1,434 @@
...
+```
+
+#### Parameters
+
+For the command line usage the support for script parameters is needed. The simplest form is to assume that the script
+has access to the `args: Array<String>` property/parameter. More advanced is to support a declaration of the typed 
+parameters, e.g.:
+
+```
+@file:param("name", String::class)
```
udalov: With the `::class` literal, you can only reference a class, not a type. For example, this way you can't specify a parameter with the type `String?`, or `List<String>`.

proposals/scripting-support.md
```diff
@@ -0,0 +1,434 @@
...
+
+The additional parameters may include params extracted by preprocessor (`@file:param("name", String::class)`) or could 
+be used for bindings mapping in JSR-223-like implementations ([KT-18917](https://youtrack.jetbrains.com/issue/KT-18917)).  
+
+### Script definition
+
+**Script Definition** is a configuration entity that combines platform, preprocessor, and selector in one entity 
+for simplified discovery and configuration. 
+
+The definition is a kotlin class declaration annotated with specific annotations. The class should be defined in the SAM
```
udalov: Can it be an interface?

proposals/scripting-support.md
```diff
@@ -0,0 +1,434 @@
...
+```
+@ScriptByFileNameSelector("*.gradle.kts")
+```
+
+### Compilation Platform
+
+```
+interface ScriptingPlatform<CompilationConfig> {
+    fun compile(scriptSource, scriptBaseClass: KClass<Scritpt>, compilationConfig: CompilationConfig): KClass<out Script>
+    fun <getIdeData>(scriptSource, scriptBaseClass: KClass<Script>, compilationConfig: CompilationConfig): <IdeData>
```
udalov: What does this notation mean?

proposals/scripting-support.md
```diff
@@ -0,0 +1,434 @@
...
+```
+@file:ScriptTemplate("org.jetbrains.kotlin.gradle.GradleScript")
+```
+
+### Compilation Configuration
+
+The `CompilationConfig` for the JVM platform may look like:
+```
+interface KotlinJVMScriptAdditionalParameters {
+    val classpath: Iterable<File> // dependencies classpath
```
udalov: Should we also have module path + names of modules to load, for Java 9+?

proposals/scripting-support.md
ligee: Yes, thanks, I remember that. Therefore the syntax of passing params in the arguments to the annotations should be supported as well.

proposals/scripting-support.md
ligee: Is there a less restrictive way to do it? (besides the stringly typing)

proposals/scripting-support.md
ligee: I do not see the reason why not. I'll fix the description.

proposals/scripting-support.md
ligee: If you mean angle brackets - it means that I cannot figure out names yet. Something I need to discuss with Pavel yet.

proposals/scripting-support.md
ligee: Yes, anything that platform needs, this is just an example. The actual definition of the `CompilationConfig` should not be specified, but rather taken from the platform definition.

proposals/scripting-support.md
udalov: OK, maybe it should be explained in the document because I was really confused. Maybe just use some names, and explain that those are tentative

proposals/scripting-support.md
udalov: No, there's no way to reference a type in an annotation, see https://github.com/Kotlin/KEEP/pull/39

proposals/scripting-support.md
```diff
@@ -0,0 +1,434 @@
...
+  - the `compilationConfig` is defined by the platform, and may affect code generation; typically contains:
+    - additional dependencies
+    - *required* scripts
+    - expected parameters
+  - should not keep the state of the script execution
+- **Executor** - the component that receives compiled script instantiates it, and then executes it in a required 
+  environment, supplying any arguments that the script requires:
+  - `(compiledScript, compilationConfig, environment) -> scriptObject`
+  - `(scriptObject, compilationConfig, environment) -> Unit`
+  - uses `compilationConfig` and `environment` for "feeding" the script with the arguments
```
abreslav: If "feeding" is not a specially coined term, replace it with "passing arguments to the script"

proposals/scripting-support.md
```diff
@@ -0,0 +1,434 @@
...
+  - *previously called DependenciesResolver*
+  - have access to specific compiler services to:
+    - extract script (file-level) annotations
+    - extract parts of the script (*source sections plugin*)
+    - (potentially) perform script lexing for advanced script analysis
+  - predefined platform-specific preprocessors are available for standard cases, custom one could be provided by the 
+    scripting host
+- **Selector** - receives the script file and/or text before further analysis to determine whether the script belongs
+  to this scripting host and the particular set of components
+  - `(scriptSource) -> Boolean`
```
abreslav: Does the file name belong to scriptSource?

proposals/scripting-support.md
```diff
@@ -0,0 +1,434 @@
...
+  - required for IDE support
+  - could be combined with the preprocessor, since it may require the same text analysing services 
+  - default and simple implementations are provided
+
+### Script object
+
+Script compiled into the following class
+
+```
+class Script(templateParams..., val implicitReceivers[], additionalArgs: Map<String, Any?>): ScriptDefinition(definitionClassParams...) {
```
abreslav: What does `[]` mean here?

proposals/scripting-support.md
```diff
@@ -0,0 +1,434 @@
...
+    
+    val val1: V1 by initOnce // for all vals/vars defined in the script body
+    
+    fun fn1(...): R1 {} // for all funcs defined in the script body
+    
+    class Cl1(...) {} // for all classes/objects defined in the script body
+    
+    var param1: P1 = additionalArgs["param1"] // for all additionalParams, the types are defined by `compilationConfig` extracted by Preprocessor 
+    
+    [suspend] fun <SAM name>(lambdaParams...): RetVal {
```
abreslav: The notation here is rather confusing

proposals/scripting-support.md
abreslav: Does it have to be abstract? We could use an annotation like `@ScriptBody` so that the script is generated into an override, no matter whether the member was abstract or not. Must not be final though

abreslav: Also, can it be a property, not a function?

proposals/scripting-support.md
abreslav: +1

proposals/scripting-support.md
```diff
@@ -0,0 +1,434 @@
...
+    fun createInstance(compiledScript: KClass<out Script>, compilationConfig: CompilationConfig, params...): Script
+    fun run(scriptObject: Script, compilationConfig: CompilationConfig): Unit
+}
+```
+
+### Preprocessor
+
+```
+interface ScriptPreprocessor<CompilationConfig> {
+    fun preprocess(scriptSource): CompilationConfig
```
abreslav: We have to clarify what "scriptSource" is here. Bot the file name and file contents should be available through it, IMO, but maybe something else too

proposals/scripting-support.md
```diff
@@ -0,0 +1,434 @@
...
+interface ScriptPreprocessor<CompilationConfig> {
+    fun preprocess(scriptSource): CompilationConfig 
+}
+```
+
+### Selector
+
+```
+interface ScriptSelector {
+    fun isScript(scriptSource): boolean
```
abreslav: Most like likely a bad name. Is it really checking whether this is a script or not? I thought that it was checking whether this script can be processed by something (what is it, btw: platform, host, definition?)

proposals/scripting-support.md
```diff
@@ -0,0 +1,434 @@
...
+### Implementing Scripting Support
+     
+To implement (custom) scripting support one should do the following
+
+1. Implement common (IDE and host) parts
+   - Choose target platform or implement custom one
+   - choose appropriate preprocessor or implement custom
+   - choose/implement selector
+   - define script class with the `@ScriptDefinition` annotation
+   - implement executor for the script class
```
abreslav: What do I do with all these implementations?

proposals/scripting-support.md
```diff
@@ -0,0 +1,434 @@
...
+   - initialize preprocessor with the platform (e.g. for using compilation services)
+   - initialize executor
+   - initialize selector
+   - for each script:
+     - (optionally) - check if the script is supported
+     - call preprocessor
+     - call platform to compile script
+     - call execute to get an instance
+     - call executor to run the script instance
+   - *(the base host may implement most of the boilerplate)*
```
abreslav: Not clear which APIs I can use while doing this

abreslav: It would be best to have a sample that does all this to refer to

proposals/scripting-support.md
ligee: @abreslav do you see it as an addition? I'm fine with the `@ScriptBody~ annotation, but I think that using SAM notation by default is quite convenient. 
And what would be the use of the non-abstract script body method?

proposals/scripting-support.md
```diff
@@ -0,0 +1,434 @@
...
+* In-process scripting and REPL for IDE
+* Consoles like IPython Notebook
+* Game scripting engines
+* ...
+
+## Basic definitions
+
+- **Script** - a text file written in Kotlin language but allowing top-level statements and having access to some 
+  implicit (not directly mentioned in the script text) properties, as if the whole script body is a body of an implicit
+  function
```
ligee: I meant `parameters` in fact. But the statement is ambiguous since we assume some implicit imports that are not referenced in the body as well.

proposals/scripting-support.md
ligee: It is more like "to compute arguments for the script"

proposals/scripting-support.md
ligee: yes, I meant that `scriptSource` (as `ScriptContent` class now) references the script either by file path or by contents or both.

proposals/scripting-support.md
ligee: That it is an array of receivers. I'll put it explicitly there.

ligee: though - this is an implementation detail, so I'll put it in the same way as `templateParams`

proposals/scripting-support.md
abreslav: Well, this means that I misunderstood the purpose of this, because of this terminonlgy

proposals/scripting-support.md
ligee: I agree. I was playing with `isKnownScript` name as well, but it looked cumbersome. May be it is the better one anyway.

proposals/scripting-support.md
ligee: They are used in the 2) and 3)

PR #78, opened 2017-08-14
JSR-305 custom nullability qualifiers

dzharkov:


proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,321 @@
...
+the `kotlin.annotation` package:
+```kotlin
+enum class MigrationStatus {
+    ERROR,
+    WARNING,
+    IGNORE
+}
+
+@Target(AnnotationTarget.ANNOTATION_CLASS)
+annotation class Migration(val status: MigrationStatus)
```
udalov: Why such a generic name? Is it supposed to be used for something else in the future?

proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,321 @@
...
+fun bar(a: A) {
+    a.foo(null) // a warning will be reported on 'null' argument
+}
+```
+
+In the example the warning will be reported until status argument is changed
+to `ERROR`, then it should become an error.
+
+It's important that migration annotation works only for its immediate usages
+on types or through `TypeQualifierDefault` annotation.
```
udalov: It says above
> it can be only used for custom nullability qualifier nicknames
, so this line is a bit confusing. Can I use it with `TypeQualifierDefault` or not? Maybe provide an example

proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,321 @@
...
+To make it work new annotation class must be annotated with 
+`@TypeQualifierDefault` and `@NonNull` or any type qualifier nickname to the 
+latter.
+The single argument for the `@TypeQualifierDefault` defines a set of element 
+types for which the annotation may enhance a type. 
+
+When determining immediate nullability of a type the Kotlin compiler should look
+for nullability annotation on the type itself at first.
+- If there is one, then it should use it
+- Otherwise, nullability is determined by the closest annotation marked as
```
udalov: It may be unclear to users what "closest" means here

proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,321 @@
...
+
+They may be set in a build systems configuration files or in the IDE.  
+
+#### Global state of JSR-305 support
+The flag `-Xjsr305-annotations` has three options: ignore, enable and warn.
+
+It effectively defines the migration status behavior only for nullability 
+qualifier nicknames that haven't yet been annotated with `kotlin.Migration`.
+
+This annotation is necessary, since using all of the custom nullability
```
udalov: This annotation -> This compiler flag?

proposals/jsr-305-custom-nullability-qualifiers.md
udalov: If yes, maybe it makes sense to rename `-Xjsr305-annotation-migration` to something like `-Xoverride-migration-status`

proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,321 @@
...
+migration status for the library: both they may want to have errors while the
+official migration status is `WARNING`, and vice versa, they may wish to 
+postpone errors reporting for some time because they're not completed their 
+migration yet.
+
+### Overriding migration status for specific annotation
+Sometimes it might be necessary to manage a migration phase for a particular
+library. That could be done with the flag `-Xjsr305-user-annotation`. It
+accepts fully-qualified name of a nullability qualifier annotation and its 
+desired status in a format `<fq.name>:ignore/warn/error`.
```
udalov: Please provide a concrete usage example of this compiler argument here in the proposal

proposals/jsr-305-custom-nullability-qualifiers.md
dzharkov: @udalov How do you think, would the phrase "the innermost enclosing element annotated with `TypeQualifierDefault`" be more clear in the context?

proposals/jsr-305-custom-nullability-qualifiers.md
dzharkov: The name is a subject for discussion as usual, but the flag name you proposed seems to be more appropriate here

proposals/jsr-305-custom-nullability-qualifiers.md
dzharkov: Of course, it's all about the compiler flag
This sentence is intended to describe the motivation behind its introduction (a lot of already annotated API that we haven't processed properly until now)

proposals/jsr-305-custom-nullability-qualifiers.md
udalov: Yes, that's more clear

proposals/jsr-305-custom-nullability-qualifiers.md
udalov: I mean, what other use cases for this annotation were considered?

If none were, then it probably should reflect that it's about JSR-305 in its name, e.g. `kotlin.jvm.Jsr305Migration`, and the compiler argument should indeed be named something like `-Xjsr305-annotation-migration`.

proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,321 @@
...
+it means that they can be used both as nullable and not-nullable. 
+
+In addition, Java types declarations may be annotated with some of the known 
+annotations to make them have [more precise types](https://github.com/JetBrains/kotlin/blob/master/spec-docs/flexible-java-types.md#more-precise-type-information-from-annotations)
+when used from Kotlin. Let's call this kind of annotating as types enhancement.
+
+Current proposal is supposed to introduce additional ways of types enhancement
+and instruments to control their migration status.
+
+### Type qualifier nicknames
```
abreslav: This part is very hard to understand for people having no background in how JSR-305 is supposed to work. Given that there's no comprehensive doc on JSR-305, I think we should include a brief description of its relevant parts

proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,321 @@
...
+with another type qualifier nickname to `NonNull`, including for example
+`@CheckForNull`.
+
+This new annotation class when being applied to a type container 
+declaration (a value parameter, field or method) must enhance related type
+in the same way it would be done if it was annotated as `@NonNull` with
+corresponding `when` argument. Namely:
+- `When.ALWAYS` makes the type not-nullable
+- `When.MAYBE` makes the type nullable
+- `When.NEVER/When.UNKNOWN` doesn't affect type's nullability
```
abreslav: Why do we treat `NEVER` this way?

proposals/jsr-305-custom-nullability-qualifiers.md
abreslav: Having read a bit further, I realize that this part is giving the necessary background, but it's not clear upfront. We should add a comment on this, along the lines of "JSR 305 has this and that semantics, and we support the following part of it"

proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,321 @@
...
+
+When determining immediate nullability of a type the Kotlin compiler should look
+for nullability annotation on the type itself at first.
+- If there is one, then it should use it
+- Otherwise, nullability is determined by the closest annotation marked as
+`TypeQualifierDefault` and having appropriate applicability in the argument:
+    - `ElementType.METHOD` for return type of methods
+    - `ElementType.PARAMETER` for value parameters
+    - `ElementType.FIELD` for fields
+    - `ElementType.TYPE_USE` for any top-level type
```
abreslav: Why only top-level?

proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,321 @@
...
+    
+Note, that `javax.annotation.ParametersAreNonnullByDefault` annotation should 
+work automatically
+by applying the rules above.
```
abreslav: It would be great to include its declaration here

abreslav: And a link to its JavaDoc or some other description would be great too

proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,321 @@
...
+public @interface NonNullApi {
+}
+
+@Nonnull(when = When.MAYBE)
+@TypeQualifierDefault({ElementType.METHOD, ElementType.PARAMETER})
+public @interface NullableApi {
+}
+```
+```java
+// FILE: test/package-info.java
```
abreslav: I'd suggest to put each file in a separate code block

proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,321 @@
...
+    @NotNullApi // overriding default from the class
+    String bar(String x, @Nullable String y); // fun bar(x: String, y: String?): String 
+}
+
+```
+
+### `@Migration` annotation
+Current proposal suggests to add the following meta-annotation in 
+the `kotlin.annotation` package:
+```kotlin
```
abreslav: I'd add a package statement here too

proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,321 @@
...
+ 
+    @NotNullApi // overriding default from the class
+    String bar(String x, @Nullable String y); // fun bar(x: String, y: String?): String 
+}
+
+```
+
+### `@Migration` annotation
+Current proposal suggests to add the following meta-annotation in 
+the `kotlin.annotation` package:
```
abreslav: Did we discuss the package much? Looks like it should be something jvm-specific, e.g. `kotlin.annotation.jvm`

proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,321 @@
...
+    IGNORE
+}
+
+@Target(AnnotationTarget.ANNOTATION_CLASS)
+annotation class Migration(val status: MigrationStatus)
+```
+
+Its aim is to define a migration status of an annotation it's been applied to.
+To the moment it can be only used for custom nullability qualifier 
+[nicknames](#Type qualifier nicknames) introduced in this proposal as well.
```
abreslav: Link is broken

proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,321 @@
...
+
+When the annotation is applied to a nullability nickname its argument specifies
+how the compiler handles the nickname:
+- `MigrationStatus.ERROR` makes annotation work just the same way as any plain
```
abreslav: We need to explain why we need this

proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,321 @@
...
+nullability annotation, i.e. reporting errors for inappropriate usages of an 
+annotated type. In other words, it's equivalent to absence of the 
+`@Migration` annotation.
+- `MigrationStatus.WARNING` should work just the same as 
+`MigrationStatus.ERROR`, but compilation warnings must be reported instead of 
+errors 
+for 
+inappropriate usages of an annotated type 
+- `MigrationStatus.IGNORE` makes compiler to ignore the annotation completely
```
abreslav: It's unclear which annotation is ignored. A motivating use case would help

proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,450 @@
...
+    // The type of `x` parameter remains flexible because there's explicit UNKNOWN-marked
+    // nullability annotation
+    String baz(@Nonnull(when = When.UNKNOWN) String x); // fun baz(x: String!): String?
+}
+
+```
+
+### `@UnderMigration` annotation
+Current proposal suggests to add the following meta-annotation in 
+the `kotlin.annotations.jvm` package:
```
udalov: You should probably also highlight that it's in a separate library, not in kotlin-stdlib

proposals/jsr-305-custom-nullability-qualifiers.md
```diff
@@ -0,0 +1,450 @@
...
+classpath, there's no need for the dependency as all necessary information 
+(annotation fully-qualified names, arguments, etc.) is already contained in 
+the resulting classfiles.
+
+Thus, it's worth explicitly declaring that the Kotlin compiler should be able
+to load information from nullability annotations (the built-in ones and 
+custom nullability qualifiers) without `jsr305.jar` as a dependency.
+
+# Open questions/issues
+- Should nullability default qualifiers with `TYPE_USE` applicability enhance
```
udalov: Not an open question anymore, since we agreed to do this?

PR #80, opened 2017-08-15
Renamed an old  'entityView' to 'containerView'

vdesmet93:
There was a small inconsistency in the code. All 'entityView' fields were renamed to 'containerView' recently, except the one in the LayoutContainer snippet.


yanex:
Merged. Thanks!

PR #88, opened 2017-10-05
correct kotlinlang slack invitation link

alteredworlds:
The link for getting an invite for kotlinlang Slack doesn't work: this PR should fix that.


yole:
Thanks!

PR #96, opened 2017-12-28
Add kotlin modifier to snippets for syntax colors

LouisCAD:
Because code is easier to read with syntax highlighting, right?


udalov:
Thanks!

Issue #99, opened 2018-02-19
Meta-annotations for Java types enhancement


dzharkov:
Discussion for the proposal:
https://github.com/Kotlin/KEEP/blob/jvm-meta-annotations-artifact/proposals/jvm-meta-annotations-artifact.md

PR for comments: https://github.com/Kotlin/KEEP/pull/100
YT issue: https://youtrack.jetbrains.com/issue/KT-23812

Feedback is very welcome, especially about the [remaing questions](https://github.com/Kotlin/KEEP/blob/jvm-meta-annotations-artifact/proposals/jvm-meta-annotations-artifact.md#remaining-questions) section

dzharkov:
We're archiving this issue in favor of [JSpecify](https://jspecify.dev/) project

PR #100, opened 2018-02-19
Artifact with meta-annotations for Java types enhancement

dzharkov:


proposals/jvm-meta-annotations-artifact.md
```diff
@@ -0,0 +1,160 @@
...
+* **Author**: Denis Zharkov
+* **Contributors**: Andrey Breslav
+* **Status**: Submitted
+* **Prototype**: Not started
+* **Discussion**: [KEEP-99](https://github.com/Kotlin/KEEP/issues/99)
+* **Related proposals**: [JSR-305 custom nullability qualifiers](https://github.com/Kotlin/KEEP/blob/master/proposals/jsr-305-custom-nullability-qualifiers.md)
+
+## Summary
+
+Add a separate artifact containing meta-annotations similar to ones from [JSR-305](https://jcp.org/en/jsr/detail?id=305).
```
abreslav: I'd say "covering use cases of", not just "similar"

proposals/jvm-meta-annotations-artifact.md
```diff
@@ -0,0 +1,160 @@
...
+
+So, there will only be fixed number of built-in qualifier annotations:
+- `nullability.Nullable`
+- `nullability.NotNull`
+- `mutability.Mutable`
+- `mutability.ReadOnly`
+
+Their target set would be the following:
+ElementType.METHOD, ElementType.FIELD, 
+ElementType.PARAMETER, ElementType.LOCAL_VARIABLE
```
abreslav: What about TYPE_USE?

proposals/jvm-meta-annotations-artifact.md
```diff
@@ -0,0 +1,160 @@
...
+```kotlin
+@Alias(Nullable::class)
+@UnderMigration(...)
+annotation class MyNullable
+
+@Alias(MyNullable::class, propagateTo=[ApplicabilityKind.PARAMETER])
+annotation class MyAllParametersAreNullableByDefault
+```
+- Also it might be confusing when one alias (probably with non-empty `propagateTo`)
+references another alias with non-trivial `propagateTo` argument
```
abreslav: This can be outlawed

proposals/jvm-meta-annotations-artifact.md
```diff
@@ -0,0 +1,160 @@
...
+of type `DefaultApplicabilityType`:
+```kotlin
+@Target(AnnotationTarget.ANNOTATION_CLASS)
+annotation class ApplyByDefault(val qualifier: KClass<*>, vararg val elements: ApplicabilityKind)
+```
+
+Basically, it should work just the same as `@TypeQualifierDefault`, e.g. it might be used like:
+```kotlin
+@ApplyByDefault(MyNullable::class)
+annotation class MyAllParametersAreNullableByDefault
```
abreslav: Looks like other annotations should be present as well, for comarison

proposals/jvm-meta-annotations-artifact.md
```diff
@@ -0,0 +1,160 @@
...
+@Alias(kotlin.annotations.jvm.nullability.Nullable::class)
+annotation class MyNullable
+```
+
+Thus, applying `@MyNullable` should have the same effect as `@Nullable` itself has.
+Beside it, `@MyNullable` may have `@UnderMigration` annotation on it that would change its 
+[migration status](https://github.com/Kotlin/KEEP/blob/master/proposals/jsr-305-custom-nullability-qualifiers.md#undermigration-annotation).
+
+### Default qualifiers options<a name="default-qualifiers"></a>
+This proposal has two options how
```
abreslav: I think we had three:

1. Alias, ApplyByDefault, UnderMigration
2. Alias(propagateTo), UnderMigration
3. Alias(propagateTo, migrationStatus)

proposals/jvm-meta-annotations-artifact.md
```diff
@@ -0,0 +1,160 @@
...
+having the same argument value.
+
+### Details on artifact
+
+There is already an artifact called [kotlin-annotations-jvm](https://mvnrepository.com/artifact/org.jetbrains.kotlin/kotlin-annotations-jvm)
+that might be the best candidate where the new meta annotations may be placed.
+
+Probably, the annotations that are already there should be moved to a different packages:
+
+- `kotlin.annotations.jvm.MigrationStatus` -> `kotlin.annotations.jvm.meta`
```
abreslav: What about UnderMigration?

proposals/jvm-meta-annotations-artifact.md
```diff
@@ -0,0 +1,160 @@
...
+
+Thus, the idea is to add `migrationStatus` parameter to `meta.Alias` 
+(and to `meta.ApplyByDefault` if we decide to introduce the latter) with default value to be `STRICT`.
+
+This argument should be processed by the compiler in the same way as for [`@UnderMigration`](https://github.com/Kotlin/KEEP/blob/master/proposals/jsr-305-custom-nullability-qualifiers.md#undermigration-annotation) annotation 
+having the same argument value.
+
+### Details on artifact
+
+There is already an artifact called [kotlin-annotations-jvm](https://mvnrepository.com/artifact/org.jetbrains.kotlin/kotlin-annotations-jvm)
```
sdeleuze: Same open question than for the package, isn't Kotlin scope too narrow for such general nullability need?

proposals/jvm-meta-annotations-artifact.md
```diff
@@ -0,0 +1,160 @@
...
+that might be the best candidate where the new meta annotations may be placed.
+
+Probably, the annotations that are already there should be moved to a different packages:
+
+- `kotlin.annotations.jvm.MigrationStatus` -> `kotlin.annotations.jvm.meta`
+- `kotlin.annotations.jvm.Mutable`, `kotlin.annotations.jvm.ReadOnly` -> `kotlin.annotations.jvm.collections`
+
+## Remaining questions
+- Should aliasing JSR-305 qualifiers like `javax.annotation.Nonnull` be allowed?
+- What option is the best to choose for supporting [default qualifiers](#default-qualifiers)?
```
sdeleuze: Option 2 looks cleaner to me, but no strong opinion here. Also something to keep in mind is that JVM seems to generates warning when a library (for example Spring) is using enum annotation attributes that are not in the classpath, which was a use case of JSR 305. Maybe this library will be in the projects classpath since it won't have JSR 305 package issue, but I prefer to mention it in order to be sure this is not an issue for projects that will use this library.

proposals/jvm-meta-annotations-artifact.md
```diff
@@ -0,0 +1,160 @@
...
+- Should there be a `migrationStatus` parameter in `@Alias` or migration status
+should be tracked through `@UnderMigration` annotation?
+- What is the best name for `ApplicabilityKind` enum class? 
+Would it be better to place it inside the `ApplyByDefault` annotation class (if it will be there)
```
sdeleuze: `ApplyByDefault.Scope` ?

sdeleuze:
Hey, since we are actively working on [Spring Framework 5.1 RC1](https://jira.spring.io/browse/SPR/fixforversion/16630/), I would be interested to have a feedback and some visibility about the roadmap of this topic. Thanks!

proposals/jvm-meta-annotations-artifact.md
dzharkov: Only if they both are written in Koltin

proposals/jvm-meta-annotations-artifact.md
dzharkov: Your 2d and 3d options are different only by the way how the migration phase is defined.
And this is a different issue described in the next section.

proposals/jvm-meta-annotations-artifact.md
```diff
@@ -0,0 +1,160 @@
...
+- There is a [modules-related issue](https://blog.codefx.org/java/jsr-305-java-9/) with JSR-305 and Java 9.
+- It's worth simplifying the way how JSR-305 nullability meta-annotations are being used 
+and integrating them with Kotlin-specific meta annotations. Namely, `@UnderMigration` and `@ApplyToTypeArguments`.
+
+## Description
+
+This section describes proposed semantics of the new annotations and partly the layout of resulting artifact.
+
+### Root package 
+Root package for this artifact will be `kotlin.annotations.jvm`.
```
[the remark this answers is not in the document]
dzharkov: Fair point!
We'll think about it

dzharkov:
Hey, unfortunately at the moment I can't tell anything concrete on the roadmap, but will try to figure it out

proposals/jvm-meta-annotations-artifact.md
sdeleuze: What I mean is that the JVM is lenient about meta-annotations not being in the classpath except when these meta-annotations are using enums not in the classpath for their attribute value.

For example if I compile a Spring project without JSR 305 dependency, and if that project is using [`@org.springframework.lang.NonNull`](https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/lang/NonNull.java), no warning when compiling because only `@Nonnull` and `@TypeQualifierNickname` meta-annotations are used.

But if I use [`@org.springframework.lang.Nullable`](https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/lang/Nullable.java), then I get the following warning during compilation: `Warning:java: unknown enum constant javax.annotation.meta.When.MAYBE reason: class file for javax.annotation.meta.When not found` because this annotation is meta-annotated with `@Nonnull(when = When.MAYBE)`.

This issue very painful for such usual use case would not occur if JSR 305 was using `@Nonnull(when = "MAYBE")` or `@Nonnull @Maybe`, so I would be strongly for the new meta-annotation library you will provide avoiding the same mistake as it will allow all libraries to use your meta-annotations without forcing all users of these libraries to have crappy compilation warnings in their logs if they don't have Jetbrains lib in the classpath. IMO this is a key point for wide adoption.

dzharkov: NB: AFAIK it only happens when using these annotations in a user's project that depends on Spring project but doesn't have `jsr305.jar` in the classpath. I.e.. when calling a piece of annotated API no warning is emitted.

remal:
@dzharkov are there any news about the roadmap?

dzharkov:
> @dzharkov are there any news about the roadmap?
Currently, there are no plans to release it really soon. But to some extent, [Jspecify](https://jspecify.dev/) project might serve as a replacement

nikitabobko:
Closing because https://github.com/Kotlin/KEEP/issues/99#issuecomment-746370004

PR #103, opened 2018-04-04
Inline classes

zarechenskiy:
Discussion of this proposal is held in [this issue](https://github.com/Kotlin/KEEP/issues/104).

### DIFF
--- a/proposals/inline-classes.md
+++ b/proposals/inline-classes.md
@@ -13,7 +13,7 @@ Discussion of this proposal is held in [this issue](https://github.com/Kotlin/KE
 Currently, there is no performant way to create wrapper for a value of a corresponding type. The only way is to create a usual class, 
 but the use of such classes would require additional heap allocations, which can be critical for many use cases.    
 
-We propose to support identityless inline classes that would allow to introduce wrappers for a values without additional overhead related 
+We propose to support identityless inline classes that would allow to introduce wrappers for values without additional overhead related 
 to additional heap allocations.
 
 ## Motivation / use cases
@@ -24,7 +24,7 @@ This is similar to type aliases but inline classes are not assignment-compatible
 Use cases:
 
 - Unsigned types
-```
+```kotlin
 inline class UInt(private val value: Int) { ... }
 inline class UShort(private val value: Short) { ... }
 inline class UByte(private val value: Byte) { ... }
@@ -38,12 +38,57 @@ inline class ULong(private val value: Long) { ... }
 - Units of measurement
 - Result type (aka Try monad) [KT-18608](https://youtrack.jetbrains.com/issue/KT-18608)
 - Inline property delegates
+```kotlin
+class A {
+    var something by InlinedDelegate(Foo()) // no actual instantiation of `InlinedDelegate`
+}
+
+
+inline class InlinedDelegate<T>(var node: T) {
+    operator fun setValue(thisRef: A, property: KProperty<*>, value: T) {
+        if (node !== value) {
+            thisRef.notify(node, value)
+        }
+        node = value
+    }
+
+    operator fun getValue(thisRef: A, property: KProperty<*>): T {
+        return node
+    }
+}
+```
+
 - Inline wrappers
+    - Typed wrappers
+    ```kotlin
+    inline class Name(private val s: String)
+    inline class Password(private val s: String)
+    
+    fun foo() {
+        var n = Name("n") // no actual instantiation, on JVM type of `n` is String
+        val p = Password("p")
+        n = "other" // type mismatch error
+        n = p // type mismatch error
+    }
+    ```
+
+    - API refinement
+    ```
+    // Java
+    public class Foo {
+        public Object[] objects() { ... }
+    }
+    
+    // Kotlin
+    inline class RefinedFoo(val f: Foo) {
+        inline fun <T> array(): Array<T> = f.objects() as Array<T>
+    }
+    ``` 
 
 ## Description
 
 Inline classes are declared using soft keyword `inline` and must have a single property:
-```
+```kotlin
 inline class Foo(val i: Int)
 ```
 Property `i` defines type of the underlying runtime representation for inline class `Foo`, while at compile time type will be `Foo`.
@@ -52,7 +97,7 @@ From language point of view, inline classes can be considered as restricted clas
 have generics. 
 
 Example:
-```
+```kotlin
 inline class Name(val s: String) : Comparable<Name> {
     override fun compareTo(other: Name): Int = s.compareTo(other.s)
     
@@ -74,6 +119,10 @@ Currently, inline classes must satisfy the following requirements:
 - Inline class must have a public primary constructor with a single value parameter
 - Inline class must have a single read-only (`val`) property as an underlying value, which is defined in primary constructor
 - Underlying value cannot be of the same type that is containing inline class
+- Inline class with undefined (recursively defined) generics, e.g. generics with an upper bound equal to the class, is prohibited
+    ```kotlin
+    inline class A<T : A<T>>(val x: T) // error
+    ```
 - Inline class cannot have `init` block
 - Inline class must be final
 - Inline class can implement only interfaces
@@ -81,6 +130,42 @@ Currently, inline classes must satisfy the following requirements:
     - Hence, it follows that inline class can have only simple computable properties (no lateinit/delegated properties)
 - Inline class cannot have inner classes
 - Inline class must be a toplevel class 
+
+Sidenotes:
+
+- Let's elaborate requirement to have public primary constructor and restriction of `init` blocks.
+For example, we want to have an inline class for some bounded value:
+    ```kotlin
+    inline class Positive(val value: Int) {
+        init { 
+            assert(value > 0) "Value isn't positive: $value" 
+        }
+    }
+  
+    fun foo(p: Positive) {}
+    ```
+    
+    Because of inlining, method `foo` have type `int` from Java POV, so we can pass to method `foo` everything we want and `init` 
+    block will not be executed. Since we cannot control behaviour of `init` block execution, we restrict it for inline classes.
+    
+    Unfortunately, it's not enough, because `init` blocks can be emulated via factory methods:
+    ```kotlin
+    inline class Positive private constructor(val value: Int) {
+        companion object {
+            fun create(x: Int) {
+                assert(x > 0) "Value isn't positive: x"
+                return Positive(x)  
+            }  
+        }
+    }
+  
+    fun foo(p: Positive) {}
+    ```
+    
+    Again, method `foo` have type `int` from Java POV, so we can indirectly create values of type `Positive` 
+    even with the presence of private constructor.
+    
+    To make behaviour more predictable and consistent with Java, we demand public primary constructor and restrict `init` blocks.
 
 ### Other restrictions
 
@@ -101,7 +186,7 @@ Basically, rule for boxing can be formulated as follows: inline class is boxed w
 Unboxed inline class is used when value is statically known to be inline class.
 
 Examples:
-```
+```kotlin
 interface I
 
 inline class Foo(val i: Int) : I
@@ -128,22 +213,70 @@ Since boxing doesn't have side effects as is, it's possible to reuse various opt
 ### Type mapping on JVM
 
 Depending on the underlying type of inline class and its declaration site, inline class type can be mapped to the underlying type or to the 
-wrapper type.  
-
+boxed type.  
 
 | Underlying Type \ Declaration Site | Not-null | Nullable | Generic |
 | --------------------------------------------- | --------------- | ---------------- | ---------------- |
-| **Reference type** | Underlying type | Underlying type | Wrapper type |
-| **Nullable reference type** | Underlying type  | Wrapper type | Wrapper type |
-| **Primitive type** | Underlying type  | Wrapper type | Wrapper type |
-| **Nullable primitive type** | Underlying type  | Wrapper type | Wrapper type |
+| **Reference type** | Underlying type | Underlying type | Boxed type |
+| **Nullable reference type** | Underlying type  | Boxed type | Boxed type |
+| **Primitive type** | Underlying type  | Boxed type | Boxed type |
+| **Nullable primitive type** | Underlying type  | Boxed type | Boxed type |
 
 For example, from this table it follows that nullable inline class which is based on reference type is mapped to the underlying type:
-```
+```kotlin
 inline class Name(val s: String)
 
 fun foo(n: Name?) { ... } // `Name?` here is mapped to String
 ```
+
+Also, if inline class type is used in generic position, then its boxed type will be used:
+```
+// Kotlin: sample.kt
+
+inline class Name(val s: String)
+
+fun generic(names: List<Name>) {} // generic signature will have `List<Name>` as for parameters type
+fun simple(name: Name) {}
+
+// Java
+class Test {
+    void test() {
+        String name = SampleKt.simple();
+        List<Name> ls = Samplekt.generic(); // from Java POV it's List<Name>, not List<String>
+    }
+}
+```
+
+
+#### Generic inline class mapping
+
+Consider the following sample:
+```kotlin
+inline class Generic<T>(val x: T)
+
+fun foo(g: Generic<String>) {}
+```
+
+Now, type `Generic<String>` can be mapped either to `java.lang.String` or to `java.lang.Object`. 
+To make the whole rule more consistent, currently we propose to map `Generic<SomeType>` always to `java.lang.Object`, 
+i.e. we'll map upper bound of type parameter.
+
+* Sidenote: maybe it's worth to consider inline classes with reified generics:
+    ```kotlin
+    inline class Reified<reified T>(val x: T)
+    
+    fun foo(a: Reified<Int>, b: Reified<String>) // a has type `Int`, b has type `String`
+    ``` 
+
+Generic inline classes with underlying value not of type that defined by type parameter are mapped as usual generics:
+```kotlin
+inline class AsList<T>(val ls: List<T>)
+
+fun foo(param: AsList<String>) {}
+```
+
+In JVM signature `param` will have type `java.util.List`, 
+but in generic signature it will be `java.util.List<java.lang.String>` 
 
 ## Methods from `kotlin.Any`
 
@@ -152,7 +285,7 @@ Inline classes are indirectly inherited from `Any`, i.e. they can be assigned to
 Methods from `Any` (`toString`, `hashCode`, `equals`) can be useful for a user-defined inline classes and therefore should be customizable. 
 Methods `toString` and `hashCode` can be overridden as usual methods from `Any`. For method `equals` we're going to introduce new operator 
 that represents "typed" `equals` to avoid boxing for inline classes:
-```
+```kotlin
 inline class Foo(val s: String) {
     operator fun equals(other: Foo): Boolean { ... }
 }
--- a/proposals/inline-classes.md
+++ b/proposals/inline-classes.md
@@ -2,7 +2,7 @@
 
 * **Type**: Design proposal
 * **Author**: Mikhail Zarechenskiy
-* **Contributors**: Andrey Breslav, Denis Zharkov, Roman Elizarov, Stanislav Erokhin
+* **Contributors**: Andrey Breslav, Denis Zharkov, Ilya Gorbunov, Roman Elizarov, Stanislav Erokhin
 * **Status**: Under consideration
 * **Prototype**: Implemented in Kotlin 1.2.30
 
@@ -35,6 +35,19 @@ inline class ULong(private val value: Long) { ... }
 - Inline enum classes 
     - Int enum for [Android IntDef](https://developer.android.com/reference/android/support/annotation/IntDef.html)
     - String enum for Kotlin/JS (see [WebIDL enums](https://www.w3.org/TR/WebIDL-1/#idl-enums))
+    
+    Example:
+    ```kotlin
+    inline enum class Foo(val x: Int) {
+        A(0), B(1);
+        
+        fun example() { ... }
+    }
+    ```
+    
+    The constructor's arguments should be constant values and the values should be different for different entries.
+
+
 - Units of measurement
 - Result type (aka Try monad) [KT-18608](https://youtrack.jetbrains.com/issue/KT-18608)
 - Inline property delegates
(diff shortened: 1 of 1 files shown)


proposals/inline-classes.md
```diff
@@ -0,0 +1,163 @@
...
+
+- Referential equality (`===`) is prohibited for inline classes
+- vararg of inline class type is prohibited
+```
+inline class Foo(val s: String)
+
+fun test(vararg foos: Foo) { ... } // should be an error  
+```
+
+## Java interoperability
```
dzharkov: I'd also add some short description/examples of how these wrappers look like and where the methods' bodies are

proposals/inline-classes.md
```diff
@@ -0,0 +1,163 @@
...
+that represents "typed" `equals` to avoid boxing for inline classes:
+```
+inline class Foo(val s: String) {
+    operator fun equals(other: Foo): Boolean { ... }
+}
+```
+
+Compiler will generate original `equals` method that is delegated to the typed version.  
+
+By default, compiler will automatically generate `equals`, `hashCode` and `toString` same as for data classes.
```
dzharkov: BTW, are data inline classes allowed?

proposals/inline-classes.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Type mapping on JVM
+
+Depending on the underlying type of inline class and its declaration site, inline class type can be mapped to the underlying type or to the 
+wrapper type.  
+
+
+| Underlying Type \ Declaration Site | Not-null | Nullable | Generic |
+| --------------------------------------------- | --------------- | ---------------- | ---------------- |
+| **Reference type** | Underlying type | Underlying type | Wrapper type |
```
dzharkov: So, we decided that `fun foo(x: List<Name>)` here it compiles to `List<Name>` from Java POV?

proposals/inline-classes.md
```diff
@@ -0,0 +1,163 @@
+# Inline classes
```
dzharkov: I'd add a section about our arrays struggles

proposals/inline-classes.md
```diff
@@ -0,0 +1,163 @@
...
+    - Hence, it follows that inline class can have only simple computable properties (no lateinit/delegated properties)
+- Inline class cannot have inner classes
+- Inline class must be a toplevel class 
+
+### Other restrictions
+
+The following restrictions are related to the usages of inline classes:
+
+- Referential equality (`===`) is prohibited for inline classes
+- vararg of inline class type is prohibited
```
gildor: Why `vararg` is not allowed?

proposals/inline-classes.md
```diff
@@ -0,0 +1,163 @@
...
+}
+```
+
+## Current limitations
+
+Currently, inline classes must satisfy the following requirements:
+
+- Inline class must have a public primary constructor with a single value parameter
+- Inline class must have a single read-only (`val`) property as an underlying value, which is defined in primary constructor
+- Underlying value cannot be of the same type that is containing inline class
```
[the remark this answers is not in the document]
zarechenskiy: Agree with restriction about recursively defined generics, thanks!

About the example, it's OK if we'll map `Id<T>` always to `Any`, I'll add a note about this

proposals/inline-classes.md
```diff
@@ -0,0 +1,163 @@
...
+    name.greet() // method `greet` is called as a static method
+}
+```
+
+## Current limitations
+
+Currently, inline classes must satisfy the following requirements:
+
+- Inline class must have a public primary constructor with a single value parameter
+- Inline class must have a single read-only (`val`) property as an underlying value, which is defined in primary constructor
```
gildor: Why is it impossible to use `var`? Is it technical limitation or design decision?

proposals/inline-classes.md
gildor: In general would be good to provide some basic comment about particular limitations/restrictons

proposals/inline-classes.md
```diff
@@ -0,0 +1,163 @@
...
+Depending on the underlying type of inline class and its declaration site, inline class type can be mapped to the underlying type or to the 
+wrapper type.  
+
+
+| Underlying Type \ Declaration Site | Not-null | Nullable | Generic |
+| --------------------------------------------- | --------------- | ---------------- | ---------------- |
+| **Reference type** | Underlying type | Underlying type | Wrapper type |
+| **Nullable reference type** | Underlying type  | Wrapper type | Wrapper type |
+| **Primitive type** | Underlying type  | Wrapper type | Wrapper type |
+| **Nullable primitive type** | Underlying type  | Wrapper type | Wrapper type |
```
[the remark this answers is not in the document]
zarechenskiy: Agree, I think I'll replace it to a "Boxed type"

proposals/inline-classes.md
zarechenskiy: Yes, I'll add an example

proposals/inline-classes.md
zarechenskiy: Well, they can be allowed... I'm not sure it makes sense, I'll add a note

proposals/inline-classes.md
zarechenskiy: Because it's not clear what type should it have. I'll add a section about arrays and then cover this question

proposals/inline-classes.md
zarechenskiy: I'd say it's both :)
Short answer is that because inline classes are identityless, they don't have state

proposals/inline-classes.md
```diff
@@ -0,0 +1,163 @@
...
+* **Prototype**: Implemented in Kotlin 1.2.30
+
+Discussion of this proposal is held in [this issue](https://github.com/Kotlin/KEEP/issues/104).
+
+## Summary
+
+Currently, there is no performant way to create wrapper for a value of a corresponding type. The only way is to create a usual class, 
+but the use of such classes would require additional heap allocations, which can be critical for many use cases.    
+
+We propose to support identityless inline classes that would allow to introduce wrappers for a values without additional overhead related
```
felipecsl: Typo "a values"

proposals/inline-classes.md
hzsweers: Related - is there a reason sealed classes can't support inlining as long as every subclass is also inline?

proposals/inline-classes.md
```diff
@@ -0,0 +1,163 @@
...
+```
+inline class UInt(private val value: Int) { ... }
+inline class UShort(private val value: Short) { ... }
+inline class UByte(private val value: Byte) { ... }
+inline class ULong(private val value: Long) { ... }
+```
+
+- Native types like `size_t` for Kotlin/Native
+- Inline enum classes 
+    - Int enum for [Android IntDef](https://developer.android.com/reference/android/support/annotation/IntDef.html)
```
gildor: How does this suppose to work?
Of course, you can create one inline class for each int, but it's not clear how this help with enum case.
Some kind enums based on inline classes or sealed would be really helpful tho
Are inline enums part of this proposal?

proposals/inline-classes.md
fvasco: Hi @zarechenskiy, can you expose some clarification about `List<T>`, `Array<T>` and `vararg T`?

proposals/inline-classes.md
```diff
@@ -0,0 +1,163 @@
...
+inline class Name(val s: String) : Comparable<Name> {
+    override fun compareTo(other: Name): Int = s.compareTo(other.s)
+    
+    fun greet() {
+        println("Hello, $s")
+    }
+}    
+
+fun greet() {
+    val name = Name("Kotlin") // there is no actual instantiation of class `Name`
```
beworker: Would it make sense to add implicit type declaration to better visualise the statement `while at compile time type will be `Foo``?

```kotlin
val name : Name = Name("Kotlin")
```

proposals/inline-classes.md
```diff
@@ -0,0 +1,163 @@
...
+- Inline property delegates
+- Inline wrappers
+
+## Description
+
+Inline classes are declared using soft keyword `inline` and must have a single property:
+```
+inline class Foo(val i: Int)
+```
+Property `i` defines type of the underlying runtime representation for inline class `Foo`, while at compile time type will be `Foo`.
```
beworker: Will the `inline` class automatically inherit from the type of the wrapped property? In other words, can an instance of `Name` class from the example down below be used as `String`?

proposals/inline-classes.md
```diff
@@ -0,0 +1,163 @@
...
+```
+
+## Current limitations
+
+Currently, inline classes must satisfy the following requirements:
+
+- Inline class must have a public primary constructor with a single value parameter
+- Inline class must have a single read-only (`val`) property as an underlying value, which is defined in primary constructor
+- Underlying value cannot be of the same type that is containing inline class
+- Inline class cannot have `init` block
```
LouisCAD: Why would `init` block be prohibited? Isn't it where checks for the single value parameter would take place?

proposals/inline-classes.md
zarechenskiy: I believe inline enum classes deserve another proposal, but I'll put here a bit information about them

proposals/inline-classes.md
cypressious: You can't really check anything if at runtime the class isn't actually instantiated. Remember, it's `inline` after all.

zarechenskiy: Because of Java interop. We can't ensure execution of `init` block when value comes from Java. Consider this case:
```
// file: sample.kt
inline class Positive(val u: Int) {
    init { assert(0 < u) }
}

fun foo(p: Positive) {} // in JVM bytecode `p` is just `int`

// Java
...
SampleKt.foo(-10) // it's OK from Java POV
```

Also, this is the reason for public primary constructor requirement.
I'll write about this in the proposal

proposals/inline-classes.md
LouisCAD: @hzsweers Since there's no instance, how would you do `is` checks at runtime for use in a `when` or boolean expression?

proposals/inline-classes.md
```diff
@@ -0,0 +1,163 @@
...
+Methods from `Any` (`toString`, `hashCode`, `equals`) can be useful for a user-defined inline classes and therefore should be customizable. 
+Methods `toString` and `hashCode` can be overridden as usual methods from `Any`. For method `equals` we're going to introduce new operator 
+that represents "typed" `equals` to avoid boxing for inline classes:
+```
+inline class Foo(val s: String) {
+    operator fun equals(other: Foo): Boolean { ... }
+}
+```
+
+Compiler will generate original `equals` method that is delegated to the typed version.
```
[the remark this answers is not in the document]
fvasco: I consider harmful expose the private value for the `UInt` example, or the `component1` must have the same visibility of the `value` attribute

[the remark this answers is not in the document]
zarechenskiy: Could you please elaborate? Do you have a use case? Maybe it's a point to allow inline data classes

proposals/inline-classes.md
LouisCAD: @zarechenskiy Why not rule out giving the underlying value of an inline class in Java at compile time? Java would only be allowed to receive inline classes values as the underlying type, but only Kotlin would be allowed to invoke the inline "constructor".

proposals/inline-classes.md
hzsweers: I would assume that they would work the same way `is` checks would work in normal use (I didn't see that listed as a listed caveat and assumed that meant it was supported, though admittedly I don't see how)

LouisCAD: I expect `is` check to not be supported on inline classes since they are identity-less, as said [here](https://github.com/Kotlin/KEEP/pull/103#discussion_r179085704)

zarechenskiy: @LouisCAD Could you please elaborate? Type check should work the same as it works for primitives.
`<expr> is InlineClassType`
There are two options: 
- type of `expr` is statically known to be an inline class type, then type check is trivial
- type of `expr` is statically unknown, so if it's actually of an inline class type then it should be boxed, therefore we do type check for boxed type

proposals/inline-classes.md
zarechenskiy: I believe this question should be discussed in the [issue](https://github.com/Kotlin/KEEP/issues/104).
I see here several options: 
- we relax this requirement and have inconsistent behaviour
- we introduce some kind of assertions as for nullable values that come from Java
- use conversions for values that come from Java

Each of this option have cons, but it's worth to discuss it anyway

proposals/inline-classes.md
zarechenskiy: @LouisCAD Yes, same as for primitives

zarechenskiy: We can this problem in the [issue](https://github.com/Kotlin/KEEP/issues/104). We had some design of what you thinking :) 
But it has too much restrictions, you can check [opaque](https://docs.scala-lang.org/sips/opaque-types.html) types in Scala, for example

proposals/inline-classes.md
```diff
@@ -38,12 +38,57 @@ inline class ULong(private val value: Long) { ... }
 - Units of measurement
 - Result type (aka Try monad) [KT-18608](https://youtrack.jetbrains.com/issue/KT-18608)
 - Inline property delegates
+```kotlin
+class A {
+    var something by InlinedDelegate(Foo()) // no actual instantiation of `InlinedDelegate`
+}
+
+
+inline class InlinedDelegate<T>(var node: T) {
```
fvasco: > Inline class must have a single read-only (val) property as an underlying value, which is defined in primary constructor

There is an inconsistency

zarechenskiy: I'd say that this is an open question. We can support mutable inline classes if they will be used only with delegated properties. Currently we're looking for use cases

proposals/inline-classes.md
```diff
@@ -128,31 +213,79 @@ Since boxing doesn't have side effects as is, it's possible to reuse various opt
...
+inline class Name(val s: String)
+
+fun generic(names: List<Name>) {} // generic signature will have `List<Name>` as for parameters type
+fun simple(name: Name) {}
+
+// Java
+class Test {
+    void test() {
+        String name = SampleKt.simple();
+        List<Name> ls = Samplekt.generic(); // from Java POV it's List<Name>, not List<String>
```
fvasco: In the Kotlin code `Name` is a parameter type.
In Java code `Name` is a return type.
`Name` is a parameter or a return type?

proposals/inline-classes.md
```diff
@@ -128,31 +213,79 @@ Since boxing doesn't have side effects as is, it's possible to reuse various opt
 ### Type mapping on JVM
 
 Depending on the underlying type of inline class and its declaration site, inline class type can be mapped to the underlying type or to the 
-wrapper type.  
-
+boxed type.
```
fvasco: Annotate each parameter using an explicit annotation with a boxed type can really improve code readability.
This code should be injected by compiler in the same manner of `@Nullable` annotation.

proposals/inline-classes.md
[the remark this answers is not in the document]
Miha-x64: Oh, you're right, auto `component1` is useless.
Vice versa, it may be used by hand, for example:
```
inline class IntPair(private val value: Long) {
    operator fun component1() = value ushr 32
    operator fun component2() = value and 0xFFFFFFFF
}
```

proposals/inline-classes.md
fvasco: Hi @zarechenskiy,
having a delegate with a backing field is interesting, however rules exceptions are harder to explain and understand.

proposals/inline-classes.md
fvasco:
```kotlin
inline class Name(var s: String)

val name = Name("Alan")
name.s = "Bob"
```

Is `name` read-only or not?
How estimate the impact of change the `Name.s` property in a shared boxed type?

proposals/inline-classes.md
```diff
@@ -0,0 +1,296 @@
...
+    ```kotlin
+    inline class A<T : A<T>>(val x: T) // error
+    ```
+- Inline class cannot have `init` block
+- Inline class must be final
+- Inline class can implement only interfaces
+- Inline class cannot have backing fields
+    - Hence, it follows that inline class can have only simple computable properties (no lateinit/delegated properties)
+- Inline class cannot have inner classes
+- Inline class must be a toplevel class
```
artem-zinnatullin: hm, that's pretty limiting, why is that?

Use-case: inside a class I want to use an inline class as an implementation detail, but I don't want other classes to know about it

proposals/inline-classes.md
```diff
@@ -0,0 +1,296 @@
...
+    Again, method `foo` have type `int` from Java POV, so we can indirectly create values of type `Positive` 
+    even with the presence of private constructor.
+    
+    To make behaviour more predictable and consistent with Java, we demand public primary constructor and restrict `init` blocks.
+
+### Other restrictions
+
+The following restrictions are related to the usages of inline classes:
+
+- Referential equality (`===`) is prohibited for inline classes
```
artem-zinnatullin: But why? Shouldn't it just do referential equality check on underlying object?

proposals/inline-classes.md
```diff
@@ -0,0 +1,296 @@
...
+- vararg of inline class type is prohibited
+```
+inline class Foo(val s: String)
+
+fun test(vararg foos: Foo) { ... } // should be an error  
+```
+
+## Java interoperability
```
artem-zinnatullin: I guess we need JS and Native interoperability sections as well :)

proposals/inline-classes.md
LouisCAD: @artem-zinnatullin That would be dangerous. Let's say two different inline classes both wrap an int which happens to have a value of `0`. It would be considered equals reference while not representing the same thing at all?
Allowing reference equality would defeat the  purpose of inline classes.

proposals/inline-classes.md
```diff
@@ -82,6 +131,42 @@ Currently, inline classes must satisfy the following requirements:
...
+        }
+    }
+  
+    fun foo(p: Positive) {}
+    ```
+    
+    Again, method `foo` have type `int` from Java POV, so we can indirectly create values of type `Positive` 
+    even with the presence of private constructor.
+    
+    To make behaviour more predictable and consistent with Java, we demand public primary constructor and restrict `init` blocks.
```
LouisCAD: Why not allow init blocks and private constructors by enforcing boxed type when used from Java? This would work similarly to inline functions that are inline when invoked in Kotlin, but not when called from Java or the debugger.

proposals/inline-classes.md
zarechenskiy: Indeed, this is too limiting, thanks! 
It's definitely can be a nested member, I'll think more about it

proposals/inline-classes.md
zarechenskiy: At some point we'll have it :) 
For now, I'm going to put more information about common and expect/actual inline classes

proposals/inline-classes.md
zarechenskiy: Yes, this approach is nice, some kind of "closed world" for inline classes. 
But note that inline functions cannot be virtual, while inline class types can be used everywhere. To enforce boxing for Java we'll have to duplicate methods, plus there are some issues with arrays (if we'll allow them, we'll have to box each element of an array on the border)

I'd say that we should think more about this approach, at least to understand our design space

proposals/inline-classes.md
```diff
@@ -0,0 +1,296 @@
...
+
+- Inline class must have a public primary constructor with a single value parameter
+- Inline class must have a single read-only (`val`) property as an underlying value, which is defined in primary constructor
+- Underlying value cannot be of the same type that is containing inline class
+- Inline class with undefined (recursively defined) generics, e.g. generics with an upper bound equal to the class, is prohibited
+    ```kotlin
+    inline class A<T : A<T>>(val x: T) // error
+    ```
+- Inline class cannot have `init` block
+- Inline class must be final
```
LouisCAD: Why would open inline classes not be allowed?
Couldn't they be extended by other inline classes that would only be additional API wrappers that could add more restrictions and more functions?

proposals/inline-classes.md
LouisCAD: @zarechenskiy It should not duplicate methods, but rather, add a method that takes the boxed type that is visible in Java, which unboxes the value and call the static method that takes the unboxed value.
In case the defined method is also `inline`, the method with the boxed type would contain all the logic directly, since no method for the unboxed type would be generated (since they would all be inlined at call places).

proposals/inline-classes.md
```diff
@@ -0,0 +1,296 @@
...
+that represents "typed" `equals` to avoid boxing for inline classes:
+```kotlin
+inline class Foo(val s: String) {
+    operator fun equals(other: Foo): Boolean { ... }
+}
+```
+
+Compiler will generate original `equals` method that is delegated to the typed version.  
+
+By default, compiler will automatically generate `equals`, `hashCode` and `toString` same as for data classes.
```
artem-zinnatullin: There is an interesting conflict with this section about equals/hashCode and implementing interfaces. 

Let me quote:

>Compiler will generate original `equals` method that is delegated to the typed version. 

and

>Inline class can implement only interfaces

Which means that comparing two *different* inline classes that use same underlying type and implement same interface can result in wrong behavior because of comparing underlying values which can be indeed equal.

Example:

```kotlin
interface Bytable {
    fun toBytes(): ByteArray
}

inline class UserId(private val id: String): Bytable {
    override fun toBytes(): ByteArray = "userId:$id".toByteArray()
}

inline class OrderId(private val id: String) : Bytable {
    override fun toBytes(): ByteArray = "orderId:$id".toByteArray()
}

fun f(bytable1: Bytable, bytable2: Bytable) {
    assert(bytable1 == bytable2)
}

fun test() {
    f(bytable1 = UserId("123"), bytable2 = OrderId("123"))
}

    test()
}
```

If you change `inline` classes to `data` classes, test will fail.
But I think it'll pass with `inline` classes which is interesting.

proposals/inline-classes.md
zarechenskiy: Ah, indeed, `Name` and `List<Name>` should be used in return type position in Kotlin, thanks!

proposals/inline-classes.md
alanfo: I'm intrigued as to how exactly unsigned types will work under this proposal.

I can see how one could do something like this:

```kotlin
val u = UInt(42)
```

but how then would you access the `value` property if it's private or is the idea that `u` itself would be treated as the value?

Also, if you wanted to create a UInt that was greater than Int.MAX_VALUE what value would you present to the primary constructor? Presumably one would not be expected to work out the appropriate negative Int value corresponding to this so would Int literals be extended to permit higher values up to UInt.MAX_VALUE?

proposals/inline-classes.md
```diff
@@ -0,0 +1,296 @@
...
+
+In JVM signature `param` will have type `java.util.List`, 
+but in generic signature it will be `java.util.List<java.lang.String>` 
+
+## Methods from `kotlin.Any`
+
+Inline classes are indirectly inherited from `Any`, i.e. they can be assigned to a value of type `Any`, but only through boxing.
+
+Methods from `Any` (`toString`, `hashCode`, `equals`) can be useful for a user-defined inline classes and therefore should be customizable. 
+Methods `toString` and `hashCode` can be overridden as usual methods from `Any`. For method `equals` we're going to introduce new operator
```
artem-zinnatullin: Wait, how will overridding `toString()`, `hashCode()` and `equals()` work? 

```kotlin
inline class UserId(private val id: String) {
    override fun toString() = "UserId($id)"
}

fun test(userId: UserId) {
    System.out.println(userId) // ?
}
```

It'll require a wrapper class which is what `inline class` is trying to remove

proposals/inline-classes.md
```diff
@@ -0,0 +1,296 @@
...
+    name.greet() // method `greet` is called as a static method
+}
+```
+
+## Current limitations
+
+Currently, inline classes must satisfy the following requirements:
+
+- Inline class must have a public primary constructor with a single value parameter
+- Inline class must have a single read-only (`val`) property as an underlying value, which is defined in primary constructor
```
ricmf: What is the reason for this restriction?

proposals/inline-classes.md
LouisCAD: @artem-zinnatullin It'll be a static method that'll be called at invocation places, so still no additional allocation. If you put the `inline class` instance in a more generic type, it'll automatically be boxed (like `Int` and other primitives are), so it'll call the virtual version of `toString()` and alike methods.

proposals/inline-classes.md
LouisCAD: @ricmf It can only work on a single reference. If you have more than one property, you need two references, and that only works with a real enclosing instance (that is allocated, not inline). However, you can perfectly make an inline class enclosing a `Pair` if it makes sense.

fvasco: Maybe it is possible to use multiple values, I am thinking about it.
It should work like JPA `Embeddable` class , JVM escape analysis should avoid extra allocation, probably this is out of scope and we should wait for Valhalla.

The problem become tricky in presence of annotations.

proposals/inline-classes.md
artem-zinnatullin: How would `System.out.println()` or any other runtime-linked use-site know about any of these methods?

ilya-g: @artem-zinnatullin `userId` will be boxed into an instance of `UserId` wrapper class, then this object instance is being passed to `println`

artem-zinnatullin: aha, I guess it falls into `nullable/platform`-type case.

Still don't like this whole wrapping story, gotta think about it more.

proposals/inline-classes.md
Miha-x64: Any problems with single-property class which violates other statements?
```
inline class Something(t: T) {
    init {
        // some init, checks
    }
    val value: U = transform(t)
}
```

zarechenskiy: @Miha-x64 This is the same as if we have private primary constructor + public secondary constructor, therefore we are going to prohibit such cases for now

proposals/inline-classes.md
```diff
@@ -0,0 +1,296 @@
...
+| **Reference type** | Underlying type | Underlying type | Boxed type |
+| **Nullable reference type** | Underlying type  | Boxed type | Boxed type |
+| **Primitive type** | Underlying type  | Boxed type | Boxed type |
+| **Nullable primitive type** | Underlying type  | Boxed type | Boxed type |
+
+For example, from this table it follows that nullable inline class which is based on reference type is mapped to the underlying type:
+```kotlin
+inline class Name(val s: String)
+
+fun foo(n: Name?) { ... } // `Name?` here is mapped to String
```
Xfel: Shouldn't it be `String?`

proposals/inline-classes.md
Miha-x64: How runtime should resolve which subtype of inline class it deals with?

LouisCAD: @Miha-x64 The inline classes have a boxed (non inline) type for cases where you use generics or super types like `Any`, so either it can be done statically, without autoboxing, or it is boxed, so behaves like a regular class, until it reaches an auto-unboxing place again.

To make an analogy to something existing, it would work the same as primitive number types, who have their boxed types extending the `Number` class.

Miha-x64: @LouisCAD to assign Int value to a Number variable, you must box it, which eliminates the whole purpose of inline classes.

LouisCAD: @Miha-x64 Does boxing an `Int` somewhere in your program eliminates the whole purpose of primitive, non boxed integers in your whole program? I don't think so, and the same applies for inline classes.
If you _always_ use them in ways that force autoboxing, then, yes, its pointless, but you'll usually not write autoboxing code everywhere.

proposals/inline-classes.md
zarechenskiy: No, here we know that `null` can be associated only with the type `Name?`, because underlying type of the inline class `Name` is a non-null reference type.

zarechenskiy: This is similar to what we have for `IntArray`/`LongArray`/...: `IntArray?` is just an array of primitive types

proposals/inline-classes.md
Wasabi375: I don't think number literals for unsigned values are part of this proposal, but I guess if they are implemented they need to allow for higher values than there signed parts. How else are you gonna get values for `ULong`. For `UInt` you could use a function converting a `long`. That being said, maybe you are doing something wrong if you have to hardcode numbers that big.

PR #105, opened 2018-04-23
Companion values

sakno:
Propose syntax for implicit access to members of the value declared as companion value.


sakno:
Pull request will be re-created with different branch

PR #108, opened 2018-04-24
Lambda expressions in annotations

sakno:
This proposal describes language feature that allows to use lambda expressions as arguments for annotations.


bhargavms:
Would be a nice to have feature.

nikitabobko:
Thanks for the proposal, unfortunately, the feature idea is very non-trivial and requires much more analysis and work. As the first step, I suggest creating a YouTrack issue to collect use cases.

I'm closing the PR

PR #115, opened 2018-05-16
Update underscore-for-unused-parameters.md

jribe:
Added Go to the list of other languages. Go's behavior is very similar to the behavior in this proposal.


nikitabobko:
Thanks, merged manually

PR #125, opened 2018-06-19
Update path in `scripting-support.md` to reflect the current path in …

mkobit:
…the upstream Kotlin repository

### Notes

* Looks like the path is now https://github.com/JetBrains/kotlin/tree/master/libraries/examples


nikitabobko:
Thanks, merged manually

Issue #127, opened 2018-07-05
Encapsulate successful or failed function execution


elizarov:
This issue is for discussion of the proposal to provide a Standard Library class to encapsulate successful or failed function execution.

Proposal text: [result.md](https://github.com/Kotlin/KEEP/blob/master/proposals/stdlib/result.md)

pakoito:
```
fun getURLContent(url: String): List<String> {
    val url = parseURL(url) // here parseURL returns URL, throws on failure
    val connection = url.openConnection()
    val input = connection.getInputStream()
    val source = Source.fromInputStream(input)
    return source.getLines()
}
```

What happens when `parseURL(url)`, `openConnection()`,  `getInputStream()` return `SuccessOrFailure<T>`? Do you have to use the getOrNull and nullcheck it every line? That's how Go does it :D

raulraja:
https://github.com/Kotlin/KEEP/blob/success-or-failure/proposals/stdlib/success-or-failure.md#appendix-why-flatmap-is-missing just want to note that this section is not accurate. Direct style is not equivalent to monadic comprehensions unless in the Id Monad. All other monads take care of their effects including those capable of failure without blowing the stack throwing exceptions. From all the things Kotlin has gotten from Scala monad comprehensions is the big one missing. The entire Scala community even those uninterested by FP like for comprehensions. I think a lot of people in the Kotlin community would benefit from safer error handling and generalized syntax for types that structurally include a way to flatMap like List, Deferred, Observable,...

pakoito:
Someone has pointed me that if you look closely, `parseURL(url).mapCatching { url -> url.openConnection().getOrThrow())` is equivalent to `parseURL(url).flatMap { url -> url.openConnection() }`.

elizarov:
@pakito Let me try to answer your concerns

> What happens when `parseURL(url)`, `openConnection()`, `getInputStream()` return `SuccessOrFailure<T>`? 

We explicitly discourage writing functions that return `SuccessOrFailure<T>`.  Don't fall in the trap of thinking that we are proposing to add Arrow's `Try` class in disguise into the Kotlin Standard library. 

That is why we don't have to answer this "what happens" question nor do we have to write any of the verbose alternatives you are suggesting. Moreover, we explicitly discourage using `SuccessOrFailure` to catch some exceptions, while letting others fall through. `SuccessOrFailure` is explicitly designed to encapsulate _any failure_ in the function execution.

[...]

PR #128, opened 2018-07-05
Encapsulate successful or failed function execution

elizarov:
Discussion of this proposal is held in [this issue](https://github.com/Kotlin/KEEP/issues/127).

### DIFF
--- a/proposals/stdlib/success-or-failure.md
+++ b/proposals/stdlib/success-or-failure.md
@@ -288,6 +288,8 @@ class SuccessOrFailure<out T> /* internal constructor */ {
 inline fun <R> runCatching(block: () -> R): SuccessOrFailure<R>
 inline fun <T, R> T.runCatching(block: T.() -> R): SuccessOrFailure<R>
 
+fun <R, T : R> SuccessOrFailure<T>.getOrDefault(defaultValue: R): R
+
 inline fun <R, T : R> SuccessOrFailure<T>.getOrElse(onFailure: (exception: Throwable) -> R): R
 inline fun <R, T> SuccessOrFailure<T>.fold(onSuccess: (value: T) -> R, onFailure: (exception: Throwable) -> R): R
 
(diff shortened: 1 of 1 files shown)


proposals/stdlib/success-or-failure.md
```diff
@@ -0,0 +1,687 @@
...
+### Functional error handling
+
+In mostly functional code `try { ... } catch(e: Throwable) { ... }` construct looks
+out of style. For example, consider this piece of code that uses [RxKotlin](https://github.com/ReactiveX/RxKotlin) 
+for asynchronous processing.
+It invokes `doSomethingAsync` that returns `Single<Data>` and processes potential error in a functional style:
+
+```kotlin
+doSomethingAsync()
+    .doOnError { showErrorDialog(it) }
```
artem-zinnatullin: note: this doesn't really handle errors in rx, it's just a side-effect callback. 

In rx errors can be handled by switching to another value/stream or by the final Observer at the end. Otherwise they'll be propagated as exceptions (in RxJava2 they go to [`RxJavaPlugins.onError()`](http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/plugins/RxJavaPlugins.html#onError-java.lang.Throwable-))

elizarov: @artem-zinnatullin Good point. What would an idiomatic Rx piece of code that I can use here?

artem-zinnatullin: Well, it'll break your comparison down below a bit but since you want to do some final action on the value/error it'll be something like this:

```kotlin
doSomethingAsync()
    .subscribe(
        { processData(it) },
        { showErrorDialog(it) }
    )
```

Please also note that even though RxJava and [Reactive Streams spec](https://github.com/reactive-streams/reactive-streams-jvm) support error (`Throwable`) handling for exceptions thrown in the streams, that's not very scalable way of handling errors in rx world.
Especially on larger projects, regular rx error handling allows bugs to slip to production due to error type information loss. Ie you don't know if you need to handle error on particular rx stream or not unless it's written in the javadoc or you've read its source code.

Modern projects convert expected subset of errors (`Throwable`s) to values of the stream (works especially well with Kotlin's sealed classes).

In a way it's like returning Java's `throws` back to the codebase, but with a better control over types.

elizarov: Thanks. I've updated Rx example and added more explanatory details there. I'm also glad to hear that using sealed classes for business errors is getting tracking. Maybe we could improve something in the language (in the future) to make _that_ pattern easier to use.

Issue #131, opened 2018-07-09
Random API


ilya-g:
Discussions about the [Random API](https://github.com/Kotlin/KEEP/blob/master/proposals/stdlib/random.md) proposal will be held here.

Pull request: https://github.com/Kotlin/KEEP/pull/132

voddan:
Recently I came across a twitter thread on randomness by @colmmacc and wanted to share it here.

A short summary:
- Implementations like `list[rand() % list.size]`  provide biased distributions and should be treated as bugs.

  - Example: say `rand()` returns an int uniformly from 0 to `MAX_INT`, and `list.size` is approximately `2/3 * MAX_INT`. Then the code above will return elements from the first half of the list twice as often as from the other half.

- Similar solutions with floating point numbers ` list[ (randDouble() * list.size).toInt() ]` are also flawed because double numbers from `0.0` to `1.0` are not uniform.

- The correct solution for getting a random element of a list is:

        fun random(size: Int): Int {
            val limit = RAND_MAX - RAND_MAX % size
            while(1) {
                 val r = rand():

                 if (r < limit) {
                      return r % size;
                 }
           }
        }

        val e = list[random(list.size)]

- _some more twits on shuffling arrays..._


I hope this will help us design Kotlin random APIs in a way that prevents such common misuses.

The original thread:
https://twitter.com/colmmacc/status/1012719876706840578

Wasabi375:
On the kotlin forum there was just a post about adding shuffle for arrays as well. There is also an issue about this.
https://discuss.kotlinlang.org/t/adding-shuffle-to-array-t-and-bytearray-intarray-et-al/8687
https://youtrack.jetbrains.com/issue/KT-25651

I think it might be worth to consider to add those functions to this proposal.

ilya-g:
@voddan Thanks for the link. 

We have taken care of the implementation and API, so that our users could avoid these and some other pitfalls of random number generation. For example, there are overloads of `nextInt` that generate an unbiased number in a given range, so one doesn't have to worry about `rand() % n` in their code.

If you like, you can review the implementation [here](https://github.com/JetBrains/kotlin/compare/4d51d13~1...f7337cc).

voddan:
I would like to pitch for some random sampling methods such as

    fun <E> List<E>.sampleRandom(random: Random = DefaultRandom): E
    fun <E> List<E>.sampleRandomList(size: Int, random: Random = DefaultRandom): List<E>
[...]

PR #132, opened 2018-07-09
Random API proposal

ilya-g:
Discussion goes in [KEEP-131](https://github.com/Kotlin/KEEP/issues/131)

### DIFF
--- a/proposals/stdlib/random.md
+++ b/proposals/stdlib/random.md
@@ -204,6 +204,17 @@ it can override them too, since they all are open.
 
 On JVM there's an extension function `asKotlinRandom()` to wrap any `java.util.Random` implementation into `kotlin.random.Random`.
 
+### Collection shuffling
+
+It becomes possible to provide extensions for collection shuffling with the specified source of randomness in the common standard library:
+
+```kotlin
+fun <T> MutableList<T>.shuffle(random: Random): Unit
+fun <T> Iterable<T>.shuffled(random: Random): List<T>
+```
+
+The existing `shuffle()` and `shuffled()` can be reimplemented by delegating to `shuffle(Random)` and `shuffled(Random)` respectively.
+
 ## What has to be done
 
 - [ ] Make the implementations serializable
--- a/proposals/stdlib/random.md
+++ b/proposals/stdlib/random.md
@@ -242,6 +242,11 @@ This makes it problematic to refer a correct overload of the name in the documen
 
 Instead of `nextBits(n)` one can use `nextInt()` shifting the result right by `32 - n` bits.
 
+**Unclear names of `origin` and `bound` parameters**
+
+It was noted that it's unclear that these parameters specify an half-open range, 
+where `origin` is inclusive start and `bound` is exclusive end.
+
 ## Future advancements
 
 * Extension on `List`/`Array`/`CharSequence` to select a random element from the collection.
--- a/proposals/stdlib/random.md
+++ b/proposals/stdlib/random.md
@@ -227,6 +227,12 @@ The existing `shuffle()` and `shuffled()` can be reimplemented by delegating to 
 What are the guarantees about the implementation of the seeded generator?
 Should we fix its implementation and never change it in the future?
 
+- An option here is to state that `Random(seed)` returns an unspecified repeatable RNG implementation
+that can be changed later, for example in some 1.M Kotlin release, and one can obtain a fixed repeatable RNG with
+an additional enum parameter:
+    
+        Random(seed, RandomImplementation.XORWOW)
+
 **`Random` identifier overloading**
 
 In the current naming scheme `Random` name is used to denote the abstract class, its companion, and two constructor-like functions. 
--- a/proposals/stdlib/random.md
+++ b/proposals/stdlib/random.md
@@ -238,6 +238,8 @@ Instead of `nextBits(n)` one can use `nextInt()` shifting the result right by `3
 
 ## Future advancements
 
+* Extension on `List`/`Array`/`CharSequence` to select a random element from the collection.
+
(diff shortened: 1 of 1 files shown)


proposals/stdlib/random.md
```diff
@@ -0,0 +1,245 @@
...
+        // overrides all methods delegating them to some platform-specific RNG implementation
+    }
+}
+
+// constructor-like functions to get a reproducible RNG implementation with the specified seed
+fun Random(seed: Int): Random
+fun Random(seed: Long): Random
+
+// JVM-only: functions to wrap java.util.Random in kotlin.random.Random and vice-versa
+fun java.util.Random.asKotlinRandom(): Random
```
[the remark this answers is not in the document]
Wasabi375: No, because "to" would suggest that you create a new random instance which is no longer dependent on the original. This only wraps the java random instance. 
This is consistent with the naming. One other example of this would be [`List<T>.asReversed`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/as-reversed.html).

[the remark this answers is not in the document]
vmichalak: I personally dont understand why we need `Random.asJavaRandom()` and `java.util.Random.asKotlinRandom()` on the Random object. Anyone can explain it to me ?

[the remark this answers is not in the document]
vmichalak: Thanks for your explanation, it's more clear for me now. If tomorrow, the implementation of ThreadLocalRandom change in Java, these methods will be a problem no ? (because of multiplatform compatibility constrain)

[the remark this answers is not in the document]
ilya-g: There's no multiplatform compatibility constraint for an arbitrary implementation of `Random`, only for that returned by `Random(seed)` function.

proposals/stdlib/random.md
```diff
@@ -0,0 +1,247 @@
...
+    abstract fun nextBits(bitCount: Int): Int
+
+    open fun nextInt(): Int 
+    open fun nextInt(bound: Int): Int 
+    open fun nextInt(origin: Int, bound: Int): Int
+    open fun nextInt(range: IntRange): Int
+
+    open fun nextLong(): Long
+    open fun nextLong(bound: Long): Long
+    open fun nextLong(origin: Long, bound: Long): Long
```
voddan: I find the parameter names `origin` and `bound` are very non-intuitive. I kinda guess the mean `min` and `max`, but I still have no idea if they are inclusive or exclusive. I don't think we should copy obscure parameter names from JDK.

IMHO the most intuitive naming would be the same as for Kotlin ranges: `start` and `endInclusive`

ilya-g: `bound` is an exclusive end, something that shouldn't be crossed. This might look unclear in the API summary, but I hope that the method [documentation](https://github.com/JetBrains/kotlin/blob/720298b53302797788778a16cd95dc43b7689e74/libraries/stdlib/src/kotlin/random/Random.kt#L53) will clarify that. If it won't, we can reconsider the parameter names.

proposals/stdlib/random.md
```diff
@@ -0,0 +1,253 @@
...
+    open fun nextLong(range: LongRange): Long
+    
+    open fun nextBoolean(): Boolean
+   
+    open fun nextDouble(): Double
+    open fun nextDouble(bound: Double): Double
+    open fun nextDouble(origin: Double, bound: Double): Double
```
voddan: What was the reason to have 3 overloaded methods instead of one with default parameters like `fun nextDouble(origin: Double = 0.0, bound: Double = Double.MAX_VALUE)`?

proposals/stdlib/random.md
voddan: Ok, then it's `diapasonStart` and `diapasonEndExclusive`. You see what I mean?

udalov: `origin` is really a bit confusing. It may give an impression that this parameter is maybe somehow used as the source of randomness instead of the internal state of the class, or that the returned value is tied to this parameter value (more than to `bound`) in some other sense.

proposals/stdlib/random.md
ilya-g: - `nextDouble()` returns value from 0.0 and up to 1.0
- overloads with less parameters allow more efficient implementations (less parameter checking, less edge cases)

voddan: I see. Then 2 questions:
 1) How much more efficient is `nextDouble()` than a method with default parameters?
 2) Is that efficiency worth having a separate `nextDouble()`? 
 3) Is that efficiency still worth having a relatively rare `nextDouble(bound: Double)`? 

Note that 2) and 3) combined effectively bloat the API two fold.

ilya-g: I don't have the precise measurements, but all of the following operations definitely add up. `nextDouble` without parameters doesn't have to:
- check if the optional parameters are specified and substitute them with their default values
- check if those parameters are valid
- calculate the range `bound - origin`
- check if it is finite
- multiply the random [0, 1) by the range and add the origin
- check if the resulting value is equal to bound.

proposals/stdlib/random.md
ilya-g: I've recorded the question about parameter names.

> Ok, then it's diapasonStart and diapasonEndExclusive

While we have other places in the library, where we specify half-open range with two parameters, we haven't used `exclusive` word in identifiers yet: usually it's just `startIndex`/`endIndex`

JLLeitschuh:
Given everyone who participated in this discussion, I'm guessing that you'd all be interested in this proposal as well.

**Add a CSPRNG (eg. SecureRandom) to the Kotlin StdLib #184**

I'd love your feedback.

PR #153, opened 2018-09-10
@JvmField for interface companion properties

max-kammerer:
This in an issue to discuss proposal for @JvmField for interface companion properties (PR #153).

### DIFF
--- a/proposals/jvm-field-annotation-in-interface-companion.md
+++ b/proposals/jvm-field-annotation-in-interface-companion.md
@@ -29,8 +29,6 @@ interface Foo {
 }
 ```
 
-NB: field is removed from companion class to interface one (so adding or removing `@JvmField` annotation is binary incompatible change
-
 ## Open questions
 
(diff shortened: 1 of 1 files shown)


proposals/jvm-field-annotation-in-interface-companion.md
```diff
@@ -0,0 +1,42 @@
...
+* **Discussion**: [KEEP-152](https://github.com/Kotlin/KEEP/issues/152)
+
+## Summary
+
+Allow to use '@JvmField' annotation on interface companion properties similar to class companion ones.
+
+## Description
+
+'@JvmField' annotation on interface companion properties has same effect and similar restrictions as annotation on class companion members:
+* generates static field in interface with initialization in interface <clinit>.
```
dsavvinov: I think it worth mentioning explicitly that the field will be removed from the companion class, so it's a binary incompatible change for Java clients.

max-kammerer:
@dsavvinov Thanks!

PR #163, opened 2018-10-14
JVM overloads in interface methods

adam-arold:
In this PR I've added a design proposal document for *JVM overloads in interface methods*.


elizarov:
I'm closing this PR since it does not present an implementation design (see https://github.com/Kotlin/KEEP/blob/master/README.md) Please, follow and vote for the corresponding YT issue: https://youtrack.jetbrains.com/issue/KT-36102 It would be helpful to see more use-cases there.

PR #164, opened 2018-10-19
Kotlin Coroutines

elizarov:


LouisCAD:
It seems the coroutines branch is still there. Maybe it could be removed now, or is there any reason to keep it?

Issue #168, opened 2018-11-14
Flexible delegated properties convention


dnpetrov:

fvasco:
> When resolving delegated property convention...

I consider preferable to provide more arguments as possible to the `getValue` function, so I suggest to invert the proposed order.
Moreover this choice is consistent with the parameterless `main` function and, finally, extending language syntax is not a breaking change.

dnpetrov:
The intent is completely opposite: to be able to provide less arguments. So, inverting the order will actually defeat the purpose.

Parameterless `main` is simply irrelevant here.

This change does not extend the language syntax.

This is not a breaking change, because it works only in presence of new declarations in code (operator fun provideDelegate/getValue/setValue with less arguments).

ilya-g:
For the problem of creating reflection objects for delegated properties please consider an alternative of erasing them when the `getValue`/`setValue` are inline and don't use `property` parameter.

elizarov:
@ilya-g While erasing objects during compilation solves the problem of optimizing the code, it does not solve the problem of readability. The current convention with mandatory `KProperty` is just hard to read/understand when you don't actually need said `KProperty` which, in fact, happens quite often.

fvasco:
Moreover inlining code is not always a good choice.

Are you considered `setValue`?

dnpetrov:
@ilya-g Currently we expect that all [inline] function parameters, including unused ones, should be visible in the debugger. We can consider such optimization only if we relax this requirement - e.g., for `@InlineOnly` functions.

elizarov:
Closing this in favor of tracking this issue in https://youtrack.jetbrains.com/issue/KT-16997

PR #169, opened 2018-11-14
#168 Flexible delegated property convention

dnpetrov:


elizarov:
https://github.com/Kotlin/KEEP/issues/168#issuecomment-750810307

PR #176, opened 2019-01-04
Compound extension

chuckjaz:
A compound extension proposal as discussed here: https://discuss.kotlinlang.org/t/compound-extension/10722

### DIFF
--- a/proposals/compound-extensions.md
+++ b/proposals/compound-extensions.md
@@ -28,6 +28,48 @@ receiver, expressions of type `String` have a unary `+` operator.
 
 ### Motivation / use cases
 
+* Contextual extensions ([from](https://youtrack.jetbrains.com/issue/KT-10468))
+
+  In Android it is common to refer to density independent pixels. In Kotlin this can be
+  improved by the use of an extension `val` but the calculation requires the current
+  `Context`.
+
+  Given this proposal, `View` can be extended with a  `Float` and `Int` extensions to simplify
+  usage in a `View`.
+
+  ```kotlin
+  val View.Float.dp get() = context.displayMetrics.density * this
+  val View.Int.dp get() = this.toFloat().dp
+  ```
+
+  which would then be used like,
+
+  ```kotlin
+  class SomeView : View {
+    val someDimension = 4.dp
+  }
+  ```
+
+* DSL by extension ([from](https://discuss.kotlinlang.org/t/add-infix-fun-to-a-class-without-extend-the-class/10772))
+
+  Consider the Android type `JSONObject`, the following extensions can be written:
+
+  ```kotlin
+  inline operator fun JSONObject.String.invoke(build: JSONObject.() -> Unit) = put(this, JSONObject().build())
+  inline infix operator fun JSONObject.String.to(value: Any?) = put(this, value)
+  fun json(build: JSONObject.() -> Unit) = JSONObject().build()
+  ```
+
+  which could be used like,
+
+  ```kotlin
+  val obj = json {
+      "key" {
+          "key1" to "value"
+          "key2" to 3
+      }
+  }
+
 * Allows a fluent programming style in a receiver scope
 
   Consider wishing to use `produce` from `CoroutineScope` to implement a filter
@@ -202,6 +244,41 @@ operator fun Body.String.unaryPlus() = escape(this)
 
 is syntactically valid but reports that `String` is not valid as there is no nested
 `String` class of `Body`.
+
+The type expression syntax is similarly interpreted. For example the type of,
+
+```kotlin
+fun A.B.C.someMethod(v: Int): Int = ...
+```
+
+would be written, `A.B.C.(Int) -> Int`.
+
+#### Alternate syntax
+
+Syntax is always contentious. The problem with the primary proposal is the ambiguity it has
+with namespaces. Two other syntaxes can be considered:
+
+##### Using parenthesis
+```kotlin
+fun (A, B, C).someMethod(v: Int)-> Int = ...
+```
+
+The types are unambiguous such that `fun (A.B, C).someMethod(v: Int)`, the `A.B` is
+unambiguously a reference to the nested type `B` in `A`.
+
+The type expression syntax would be similarly extended to allow `(A.B, C).() -> Unit`.
+
+However, the type expression syntax is ambiguous at the `(` and it is not until the `.` is
+seen after the initial closing `)`.
+
+#### Using brackets
+```kotlin
+fun [A, B, C].someMethod(v: Int)
+```
+
+The types are similarly unambiguous without introducing the same syntactic ambiguity. However,
+this require introducing a use for `[` in a type expression that might be better reserved by
+a feature more widely leveraged such as tuples.
 
 ### Lookup rules
 
--- a/proposals/compound-extensions.md
+++ b/proposals/compound-extensions.md
@@ -28,6 +28,8 @@ receiver, expressions of type `String` have a unary `+` operator.
 
 ### Motivation / use cases
 
+* Context-oriented programming as discussed [here](https://proandroiddev.com/an-introduction-context-oriented-programming-in-kotlin-2e79d316b0a2).
+
 * Contextual extensions ([from](https://youtrack.jetbrains.com/issue/KT-10468))
 
   In Android it is common to refer to density independent pixels. In Kotlin this can be
@@ -256,11 +258,13 @@ would be written, `A.B.C.(Int) -> Int`.
 #### Alternate syntax
 
 Syntax is always contentious. The problem with the primary proposal is the ambiguity it has
-with namespaces. Two other syntaxes can be considered:
+with namespaces. Other syntaxes to consider are:
 
 ##### Using parenthesis
 ```kotlin
-fun (A, B, C).someMethod(v: Int)-> Int = ...
+fun (A, B, C).someMethod(v: Int): Int = ...
+
+val a: (A,B,C).(v: Int)->Int = ...
 ```
 
 The types are unambiguous such that `fun (A.B, C).someMethod(v: Int)`, the `A.B` is
@@ -273,21 +277,54 @@ seen after the initial closing `)`.
 
 #### Using brackets
 ```kotlin
-fun [A, B, C].someMethod(v: Int)
+fun [A, B, C].someMethod(v: Int): Int = ...
+
+val a: [A,B,C].(Int)->Int = ...
 ```
 
 The types are similarly unambiguous without introducing the same syntactic ambiguity. However,
 this require introducing a use for `[` in a type expression that might be better reserved by
 a feature more widely leveraged such as tuples.
 
-### Lookup rules
+#### Using a pseudo-keyword
 
-When resolving the an extension, if the type expression following a `.` prior to the member
-name is not resolvable in scope of the type to the left, the ambient lookup scope is consulted
-and if the expression is resolvable to a type then the extension declaration is interpreted to
-extend the type to the left with, the potentially compound, extension of the type resolved.
+```kotlin
+extension fun (A, B, C).someMethod(v: Int): Int = ...
+val a: extension (A, B, C).(Int)->Int = ...
+```
 
-Type aliases or import aliases can be used to extend types that are obscured by nested types.
+Using a pseudo keyword avoids the ambiguity of the `(` in the type expression. A pseudo-
+keyword could also be used with the `[`...`]` syntax to leave a unadorned `[`...`]` to
+mean a tuple in the future.
+
+### Matching rules
+
+1. **Create an ordered list of implicit receivers** in scope with types
+<code>I<sup>1</sup></code>...<code>I<sup>m</sup></code> where <code>I<sup>1</sup></code>
+is the outer most implicit receiver and the <code>I<sup>m</sup></code> is the inner most.
+
+2. **Checking** given an extension function with receivers
+<code>T<sup>1</sup></code>...<code>T<sup>n</sup></code> and a receiver scope type of `R`
+and implicit receiver scopes in the scope tower of
+<code>I<sup>1</sup></code>...<code>I<sup>m</sup></code> an extension is valid if `R` :>
+<code>T<sup>n</sup></code> and there exists some permutation of (<code>T<sup>i</sup></code>,
+<code>I<sup>j</sup></code>) in `i` `1`...`n-1` and `j` in `1`..`m` where
+<code>I<sup>j</sup></code> :> <code>T<sup>i</sup></code> and for each element in the
+permutation all `i` and `j` values are in increasing order. If multiple valid permutations are
+possible then the last permutation is selected when the permutations are ordered by the values
+of `i` and `j`.
+
+3. **Report ambiguous calls** If multiple valid candidates are possible the call is ambiguous
+and an error is reported.
+
+#### Implementation note
+
+Finding a last valid permutation can be done efficiently if the permutations are produced by
+pairing  <code>T<sup>n-1</sup></code> to the last <code>I<sup>j</sup></code> where
+<code>T<sup>n-1</sup></code> :> <code>I<sup>j</sup></code> and then pairing
+<code>T<sup>n-2</sup></code> to some <code>I<sup>k</sup></code> where `k` < `j` until a valid
+permutation is found. The first valid pairing found will be the last permutation given the
+ordering described above. This can be accomplished in worst case *O(nm)* steps.
 
 ### Disambiguation of `this`
 
--- a/proposals/compound-extensions.md
+++ b/proposals/compound-extensions.md
@@ -1,0 +1,266 @@
+# Compound Extensions
+
+* Type: Design proposal
+* Author: Chuck Jazdzewski
+* Status: proposed
+* Prototype: none
+* Discussion: tbd
+
+## Summary
+
+Extensions can be used to extend a type with an extension.
+
+## Introduction
+
+### Description
+
+Compound extensions are extensions that introduce an extension onto another type.
+
+A compound extension is declared using dotted list of types. For example,
+
+```kotlin
+operator fun Body.String.unaryPlus(s: String) = escape(s)
+```
+
+is interpreted as extending `Body` with the extension `String.unaryPlus()` which further
+extends `String` with a unary `+` operator. In other words, in the scope of a `Body`
+receiver, expressions of type `String` have a unary `+` operator.
+
+### Motivation / use cases
+
+* Allows a fluent programming style in a receiver scope
+
+  Consider wishing to use `produce` from `CoroutineScope` to implement a filter
+  of a `ReceiverChannel`, the developer is faced with three choices,
+
+  1. Create a `filter` that takes `CoroutineScope` and the `ReceiverChannel` as
+       such as,
+
+       ```kotlin
+       fun <T> filter(
+           scope: CoroutineScope,
+           channel: ReceiveChannel<T>,
+           predicate: (value: T) -> Boolean) = scope.produce<T> {
+         for (value in channel) if (predicate) send(value)
+       }
+       ```
+
+       ```kotlin
+       fun main() = runBlocking {
+         filter(this, numbersFrom(2)) { it % 2 }.forEach { println(it) }
+       }
+       ```
+
+  2. Create a `filter` that extends `CoroutineScope` and takes `ReceiverChannel` as
+       a parameter such as,
+
+       ```kotlin
+       fun <T> CoroutineScope.filter(
+          channel: ReceiveChannel<T>,
+          predicate: (value: T) -> Boolean) = produce<T> {
+          for (value in channel) if (predicate) send(value)
+       }
+       ```
+
+       ```kotlin
+       fun main() = runBlocking {
+         filter(numbersFrom(2)) { it % 2 }.forEach { println(it) }
+       }
+       ```
+
+  3. Create a `filter` that `ReceiverChannel` and takes `CoroutineScope` as a
+       parameter such as,
+
+       ```kotlin
+       fun <T> RecieverChannel<T>.filter(
+           scope: CoroutineScope,
+           predicate: (value: T) -> Boolean) = scope.produce<T> {
+         for (value in this) if (predicate) send(value)
+       }
+       ```
+
+       ```kotlin
+       fun main() = runBlocking {
+         numbersFrom(2).filter(this) { it % 2 }.forEach { println(it) }
+       }
+       ```
+
+    With this proposal the developer could introduce use a more fluent `filter`,
+
+    ```kotlin
+    fun <T> CoroutineScope.ReceiverChannel<T>.filter(
+        predicate: (value: T) -> Boolean) = produce<T> {
+      for (value in this) if (predicate) send(value)
+    }
+    ```
+
+    ```kotlin
+    fun main() = runBlocking {
+        numbersFrom(2).filter { it % 2 }.forEach { println(it) }
+    }
+    ```
+
+* Generalizing local extensions
+
+  Extensions in local scope are difficult to extract. Consider a simplified HTML DSL
+  such as:
+
+    ```kotlin
+    open class HtmlDsl() { fun write(s: String) {  } fun escape(s: String) { } }
+    class Html(): HtmlDsl() {}
+    class Body(): HtmlDsl() { }
+
+    fun html(block: Html.() -> Unit) {
+        with (Html()) {
+            write("<html>")
+            block()
+            write("</html>")
+        }
+    }
+
+    fun Html.body(block: Body.() -> Unit) {
+        write("<body>")
+        Body().block()
+        write("<body>")
+    }
+
+    fun Body.div(block: Body.() -> Unit) {
+        write("<div>")
+        block()
+        write("</div>")
+    }
+
+    fun Body.span(block: Body.() -> Unit) {
+        write("<span>")
+        block()
+        write("</span>")
+    }
+    ```
+
+   To emit string content it would be convenient to overload the unary `+` operator to
+   mean emit text with html escapes. The developer cannot use extension methods for this
+   to be generally available and must add the operator to the correct classes directly
+   such as,
+
+   ```kotlin
+   class Body(): HtmlDsl() {
+       operator fun String.unaryPlus(s: String) = escape(s)
+   }
+   ```
+
+   In a local scope the developer can declare extension operators for special purposes
+   such as introducing a unary `-` operator to underline text,
+
+   ```kotlin
+   html {
+        body {
+            operator fun String.unaryMinus() {
+                underline {
+                    +this@unaryMinus
+                }
+            }
+            div {
+                -"Customer"
+                customer(customer)
+
+                -"Orders"
+                orders(orders)
+            }
+        }
+    }
+    ```
+
+    However, the developer is required to modify `Body` class to allow the unary `-` to be
+    used outside its current local scope. This is problematic if the HTML DSL is a third-party
+    library.
+
+    Compound extensions allows the original unary `+` operator to be introduced as an extension
+    function such as,
+
+    ```kotlin
+    operator fun Body.String.unaryPlus() = escape(this)
+    ```
+
+    therefore allowing unary `-` to be provided generally by extension of `Body`,
+
+    ```kotlin
+    operator fun Body.String.unaryMinus() = underline { +this@unaryMinus }
+    ```
+
+## Proposal
+
+### Syntax
+
+No new syntax is required. Extension declarations that are currently invalid are given
+a valid interpretation.
+
+For example, in the above examples,
+
+```kotlin
+operator fun Body.String.unaryPlus() = escape(this)
+```
+
+is syntactically valid but reports that `String` is not valid as there is no nested
+`String` class of `Body`.
+
+### Lookup rules
+
+When resolving the an extension, if the type expression following a `.` prior to the member
+name is not resolvable in scope of the type to the left, the ambient lookup scope is consulted
+and if the expression is resolvable to a type then the extension declaration is interpreted to
+extend the type to the left with, the potentially compound, extension of the type resolved.
+
+Type aliases or import aliases can be used to extend types that are obscured by nested types.
+
+### Disambiguation of `this`
+
+The `this` keyword refers to the most nested receiver. Reference to an outer receiver can
+be disambiguated by using the simple type name (without type parameters). In cases where that is
+not possible (such as a function type expression) a type alias with a simple name would need
+to be used in the declaration.
+
+For example,
+
+```kotlin
+fun A.B.C.doSomething() = this@A.amember(this@B.bmember(this.cmember())
+```
+
+### Semantics
+
+A synthetic receiver parameter is introduced for every receiver type in a compound receiver.
+
+For example,
+
+```kotlin
+fun A.B.C.doSomething() = this@A.amember(this@B.bmember(this.cmember())
+```
+
+would roughly translate into the Java function
+
+```Java
+  void doSomething(A receiver2, B receiver1, C receiver0) {
+      receiver2.amember(receiver1.bmember(receiver0.cmember()))
+  )
+```
+
+A call to a compound receiver collects ambient receivers into the synthetic parameter,
+
+```kotlin
+var a: A
+var b: B
+var c: C
+
+with (a) {
+    with (b) {
+        with (c) {
+            doSomething()
+        }
+    }
+}
+```
+
+would roughly translate into,
+
+```Java
+doSomething(a, b, c)
+```


proposals/compound-extensions.md
```diff
@@ -0,0 +1,380 @@
...
+## Introduction
+
+### Description
+
+Compound extensions are extensions that introduce an extension onto another type.
+
+A compound extension is declared using dotted list of types. For example,
+
+```kotlin
+operator fun Body.String.unaryPlus(s: String) = escape(s)
```
JakeWharton: This can't work. What does `Map.Entry.unaryPlus` apply to? `Map` and `Entry` or `Map.Entry`?

Wasabi375: I think this is discussed later on in the syntax part of the proposal. As far as I can tell, this example uses the current syntax as an introduction to the topic. This syntax clearly does not work, neither with namespaces (as pointed out below) not nested classes. 
The question is which of the other possible syntax variants should be used primarily. I personally prefer parenthesis.

altavir: It seems like square brackets give the best readability and parseability. We should remember that we need the syntax for function types. `[].()->` reads better than `().()->`. I think that alternative syntax solutions should be added to proposal.

altavir:
I do not really like the title. Event `compound receiver` would be be better. But commonly used name for this feature is `multiple receiver extension`.

altavir:
As a motivation for this proposal (not exactly current version but the one we got after discussion), I wrote an [article about context-oriented approach in kotlin program design](https://proandroiddev.com/an-introduction-context-oriented-programming-in-kotlin-2e79d316b0a2).

Additional motivation could be found in [kmath documentation](https://github.com/altavir/kmath/tree/dev/doc) where I actually use this approach.

Currently I kind of really need it for performance optimization.  I have contexts which define operations on some objects (for example nd-structures). Now I want to some functions like `plus` to be defined in the context:
```kotlin
class NDField<T>{
  ...
  fun NDStruncture<T>.plus(arg: T): NDStructure<T>
}
```

This operation could benefit a lot from being moved to extension. It allows not to bloat the `NDField` class and avoid complicated type constructs in inheritance. What is more important it allows to create an optimized implementations for specific types without complicated inheritance structures of field itself:
```kotlin
  fun <T> [NDField<T>, NDStruncture<T>].plus(arg: T): NDStructure<T>
  fun [NDField<Double>, NDStruncture<Double>].plus(arg: Double): NDStructure<Double>
```
Being dispatched statically, those functions could use performance optimization like inlining. Context-oriented paradigm could work solely on such extensions since the method resolution is done based not on runtime type (runtime type could be used, but it is could be avoided), but on scope type. In fact, one could use empty marker scopes and dispatch all functions statically which should also benefit the performance.

proposals/compound-extensions.md
chuckjaz: It works in either case as addressed below.

In the where there is an existing meaning then such as `Map.Entry` it refers to `Map.Entry`. If `Map` doesn't have a nested type named `Entry` it refers to the `Entry` in scope. If you have an `Entry` in scope you wish to extend, it collides with `Entry` but a alias can be introduced to allow you to disambiguate.

chuckjaz:
This seems similar in character to the `escape` html DSL example or am I missing something?

chuckjaz:
As for the name, I am open to other names but "compound receiver" doesn't seem right to me as the receive is not compound. "multiple receiver extension" is better as it is more directly what is happening.

erokhins:
@chuckjaz Thank you for proposal!

I want to highlight some not covered/problematic places.
### Syntax

As @JakeWharton mentioned before, syntax `fun A.B.foo()` can has 2 different meaning: function with receiver `A.B` (where `B` is nested class of `A`) or function with two receivers: `A` and `B`. 
Ambiguity can be resolved as you proposed (nested class wins), but user cannot force compiler use another meaning: two receivers.

Another problem here is that sometimes you want to use fully-qualified name for type i.e. `fun a.b.Foo.x.y.Bar.bas()` and in such cases resolution rules will be even worse (especially in package `a` we have class `b` and so on).

### Resolution rules

Consider the following example:
```kotlin
class A {
    fun B.C.foo() {}
}

class B
class C

fun A.B.foo() {}

fun test() {
    with(A()) {
        with(B()) {
            with(C()) {
                foo()
            }
        }
    }
}
```
This example illustrates 2 different problems:
1) what we should do if our candidates have different amoung of receivers? (We can consider that member extension function `foo` in class `A` have 3 receivers: `A`, `B`, `C`)
2) should be members wins over extensions?

### Receivers handling
Disambiguation via `this@A` where `A` is type name introduce entirely new meaning for labels -- label by type name. I would try to avoid that. Instead of that maybe we can mark receiver explicitly if necessary.

Related not covered theme -- possibility to explicitly pass all implicit receivers on call site:
```kotlin
class A
class B

fun A.B.foo() {}

fun test(a: A, b: B) {
    (a, b).foo() // one possible synthaxs for that
}
```

altavir:
The alternative syntax in discussion (round or square brackets) covers all syntactic problems. @chuckjaz could you move it from discussion to the proposal as alternative? I really do not like this dot solution.

As for member vs extension problem. As @orangy mentioned, we should try to keep the same behavior as was before for member extensions. I do not think that multiple receivers would change that. The member always win rule is still in place.

erokhins:
Sometimes extensions win over members:
```
class A {
    fun foo() {}
}
class B

fun B.foo() {}

fun test() {
    with(A()) {
        with(B()) {
            foo() // resolved to B.foo
        }
    }
}
```
So particular rules should be clarified.

altavir:
Are those rules explicitly written somewhere? It would be easier just to look at them and see what should be changed.

altavir:
I am moving two my latest posts form discussion here with little remarks

## Next iteration of resolution proposal
Here I will try to combine best parts of all current proposal and account for backward compatibility. The resolution could be done in following steps:
1. **Create a list of actual contexts** `G, A, B, C, ...` where `G` is a global context. Top level non-extension functions have only `G` as context. Class members have that class as context. Extension function have their receivers added to context where they are defined. Running a function with receiver adds that receiver to the list of contexts where this function is defined.Types in the list could be duplicating. We will call those actual types `Cy` where `y` is the index
2. **Checking** the definition of the function. Function receiver list is written in form of `[R1, R2, R3]`. Duplicate types or ambiguities throw compile error. Types could have parameters defined outside type list like `fun <T> [T, Operation<T>].doSomething()`. We will call receiver types `Rx` where `x` is the index.
3. **Binding** of extension function. Each of types `R` in receiver set is matched against the elements of context list from right to left, binding it to first match and thus creating a map `Rx to Cy`. If there is a bound pair for each of `Rx` then function is considered resolved and bound to context. Multiple `R` could be bound to the same context `C`, so it is possible to have just one context for multiple receiver types if it matches them both. 

The normalization step is probably could be avoided in this scheme. The results of this procedure are the decision about binding and binding map `Rx to Cy`.

**This** resolution is done via declared receiver types `Rx` which are then substituted by actual runtime objects representing `Cy`.

### Compatibility check

* `[A].f === A.f`.Extension function with single argument should work exactly like existing extension function. It seems like it does. It is resolved and bound to the closest context matching its type.

* [Match current member receivers strategy](https://discuss.kotlinlang.org/t/compound-extension/10722/17?u=darksnake). Seems to be working the same way. It resolves to the closest context even if this context implements both receivers.

altavir:
## Some additional thoughts
I was thinking a lot about this context-based resolution idea and I think that it is possible to go a little further (not immediately, mind you, it is the idea for the future). I am leaving this global context `G` everywhere in my schemes and it does not play any important role because it usually is empty. But in fact it could provide a lot of opportunities. 

### File level context
`G` is basically resolved to file  which does not have any type of its own. But if there was a way to bind a context or a set of contexts to the file itself (not proposing any syntax solution, but it should be quite easy). Then everywhere in my schemes above we will just replace `G` with `G, F1, F2`. Meaning that all classes and functions in this file will have additional implicit contexts, just like type-classes. Of-course, it means that `F`s could only be singletons in this case.

*This mechanism is in fact currently used in KEEP-75 for script implicit receivers. But of course, for code, it should be explicit and only singleton objects will do.*

### Extension classes and interfaces
We consider a class or interface to be top-level non-empty context. It seems to be not so hard to add a set of external contexts to the class context. It could look like `class [T1,T2].MyClass{}`. In this case the instance of this class could be created only in context bound to both `T1` and `T2` and all members of this class will have additional implicit contexts, meaning member of `MyClass` will be able to call members of `T1` without additional declarations. From the implementation point of view, the instances of `T1` and `T2` could be class constructor implicit parameters (or a single map parameter, which is probably better), it should not violate anything, even Java compatibility (you will have to pass instances of contexts to create class instance). The interfaces could work the same way, just provide synthetic (invisible) property which holds the `Map<KClass, Any>` of actual receiver.

**Note:** There will be some problem with class type parameter syntax here since unlike functions, type parameters in classes are declared after class name, but probably it could be solved.

This second proposal is in fact much more flexible than first one since we can use non-singleton contexts and define them explicitly on class creation. Also, both ideas probably could cover most of type-classes use-cases.

chuckjaz:
> I want to highlight some not covered/problematic places.
> 
> ### Syntax
> As @JakeWharton mentioned before, syntax `fun A.B.foo()` can has 2 different meaning: function with receiver `A.B` (where `B` is nested class of `A`) or function with two receivers: `A` and `B`.
> 
> Another problem here is that sometimes you want to use fully-qualified name for type i.e. `fun a.b.Foo.x.y.Bar.bas()` and in such cases resolution rules will be even worse (especially in package `a` we have class `b` and so on).

My intent here is not to change the rules of resolution but to add a meaning to a case that is currently an error. The proposal only kicks in, so to speak, when Kotlin would report an error. 

> Ambiguity can be resolved as you proposed (nested class wins), but user cannot force compiler use another meaning: two receivers.

They can by using a type alias or rename on import. This is the same case you run into if you are trying to use the type `Bar` from `x.y` in a package that defines its own `Bar`. One must be renamed or otherwise aliased.
 
> ### Resolution rules
> Consider the following example:
> 
> ```kotlin
> class A {
>     fun B.C.foo() {}
> }
> 
> class B
> class C
> 
> fun A.B.foo() {}
> 
> fun test() {
>     with(A()) {
>         with(B()) {
>             with(C()) {
>                 foo()
>             }
>         }
>     }
> }
> ```
>
> This example illustrates 2 different problems:
> 
> 1. what we should do if our candidates have different amoung of receivers? (We can consider that member extension function `foo` in class `A` have 3 receivers: `A`, `B`, `C`)

The first found is scope wins. In this case since the `fun B.C.foo()` in class `A` wins. The number of receivers is defined by the function being called. In this case, it requires 3, `A` (as `this`) `B` and `C` as receivers. Since `fun A.B.foo()` is not found the fact it takes 2 receivers is not relevant.

In other words, the lookup should behave consistently with removing the compound receiver and being left only the simple receiver. For example, because

```kotlin
class A {
    fun C.foo() {}
}
 
class B
class C
 
fun B.foo() {}
 
fun test() {
    with(A()) {
        with(B()) {
            with(C()) {
                foo()
            }
        }
    }
}
```

will result in `A`'s `fun C.foo()` being called as it is in scope prior to `fun B.foo()`. The previous example should result in `fun B.C.foo()` from class `A`, and, identically, because `fun B.foo()` would receive two receivers today (`A` as the `this` parameter and `B` as the receiver parameter) and the presence of `foo B.foo() {}` doesn't affect this, the function `fun B.C.foo()` would receiver three and is unaffected by `foo A.B.foo()` in scope.
 
> 2. should be members wins over extensions?

The current lookup rules for extensions are unmodified. A compound receiver function `foo A.B.foo()` should be thought of as extending the type `A` with a member `fun B.foo()`, which extends the type `B` with `fun foo()`. A direct member of `B` has precedence over all extensions.

> ### Receivers handling
> Disambiguation via `this@A` where `A` is type name introduce entirely new meaning for labels -- label by type name. I would try to avoid that. Instead of that maybe we can mark receiver explicitly if necessary.

I will add that as alternative. My intent was not to add a new type label but to have the type name imply a label.
 
> Related not covered theme -- possibility to explicitly pass all implicit receivers on call site:
> 
> ```kotlin
> class A
> class B
> 
> fun A.B.foo() {}
> 
> fun test(a: A, b: B) {
>     (a, b).foo() // one possible synthaxs for that
> }
> ```

This seems to be a related but separate proposal as it would apply to extensions in generate, not just compound extensions.

chuckjaz:
> Sometimes extensions win over members:
> 
> ```
> class A {
>     fun foo() {}
> }
> class B
> 
> fun B.foo() {}
> 
> fun test() {
>     with(A()) {
>         with(B()) {
>             foo() // resolved to B.foo
>         }
>     }
> }
> ```
> So particular rules should be clarified.

The proposal defines, given a scope, such as the scope for the body of the `with (B())`, how does one determine if `foo()` matches an extension function in scope. It does not affect how scopes are composed. The example above illustrates what happens when a member of `B` obscures a member of `A`. This proposal will have an effect on this case but only as far as this proposal affects how members are found in the scope of `B`.

fvasco:
In the section "Matching rules": "Report ambiguous calls If multiple valid candidates are possible the call is ambiguous and an error is reported", how to disambiguate and so fix the error?

I wish to use a compound extension to add two integers, should I write:

```kotlin
fun Int.Int.sum() = this@Int + this@Int
```

The section "Disambiguation of this" does not help for my task.

Using the above function, what is `x` value of the code:

```kotlin
val x =
 with(2) {
  with(3) {
   with(5) {
    sum()
   }
  }
 }
```

Is it an _ambiguous call_?

chuckjaz:
@fvasco This is not ambiguous, it is just obscure.

You can use a type alias to introduce a name to refer to the first receiver.

```kotlin
typealias MyInt = Int
fun MyInt.Int.sum() = this@MyInt + this@Int
```

The value of `x` above is a 8 as you would expect. It is not ambiguous as one and only one function would match `fun MyInt.Int.sum()`. The receivers are calculated as nearest where `MyInt` binds to the `Int` scope introduced by `with(3)` and `Int` binds to the receiver introduced by `with(5)`.

fvasco:
@altavir I agree with you, this is a missing point in the documentation.

If we consider the `this` scopes as a list then no permutation should be allowed (because the position is relevant, see "Matching rules"), otherwise if we consider the `this` scopes as a set then no duplication should be allowed (because multiple permutation matches).

Moreover in my opinion type alias is a bad solution, Kotlin already supports label so I consider a better approach:

```kotlin
fun A@Int.B@Int.sum() =this@A + this@B
```

altavir:
I also would like to remind everyone, that we want the feature to be gradually integrated in the language. That means restrict first, permit later approach. We need to restrict anything that could pose a problem in first implementation and think about relaxing the restriction.

And again, I do not like the dot syntax. It is confusing. It supposes that we have a fixed order, but we agreed that order of receivers should not matter. Also it will be really difficult to work with function types (you should also remember that types could have parameter).

By the way, if we are talking about parameters, we need to understand if two types with the same base and different parameter are allowed. I think that it should be restricted as well meaning that `[Box<A>, Box<B>]` (or `Box<A>.Box<B>` in dot notation) should be prohibited as well.

chuckjaz:
> @chuckjaz I am not sure that type alias is a good idea. I do not know how typealiases are resolved by compiler, but using them for _this_ resolution could bring additional level of complexity. 

The `this` scopes would exist, using a label just disambiguates them allowing you to access a possibly obscured member or the instance directly as in the example.

Currently the simple name of the type is allowable as a label such as in the case of referring to the outer class in a nested class.

```kotlin
class A {
  var foo = 1
  inner class B {
    var foo = 2
    fun addFoos() = this@A.foo + foo
  }
}
```

My proposal uses something similar to disambiguate the receivers just as the above.

> What will happen if ones switches order of `with` in example?

The order of the receiver parameters change.

BenWoodworth:
Should this proposal also allow for multiple receivers for lambdas?
E.g. `(T1, T2).() -> R`:
```
inline fun <T1, T2, R> with(receiver1: T1, receiver2: T2, block: (T1, T2).() -> R): R {
    with(receiver1) {
        with (receiver2) {
            return block()
        }
    }
}
```

altavir:
It was somewhere in discussion if not in the proposal itself. Yes, of course it does not make sense without multiple receiver function types.
Also, we definitely will want  `with(receiver2, receiver2)`.

pdvrieze:
Looking at this, I am not sure that it can work without a syntax for precise multi-receiver invocation. Currently you can do precise invocations as there are only 2 receivers and you can use implicit receiver vs explicit receiver. When multiple receivers  (>2) are present that no longer works. You probably also want to have both syntaxes be consistent.

altavir:
I am not sure how >2 case is different from case of double receiver. Of course, I am not talking about dot notation. We can easily have something like `val func: [A, B, C].()->` signature and invoke it with `with(b, c, a){func()}` notation. I believe that common consensus is that order should not matter.

pdvrieze:
@altavir At least in the current situation there are three different ways to invoke a multi-receiver function:

- `with(a) { b.foo() }`
- `with(b) { a.foo() }`
- `with(a) { with(b) { foo() } }`

These three have different resolution/priority rules that can be used, for example when `a` and `b` are of related types, and sometimes some of these variants do or do not work - cause errors.

In the case of compound extensions with 2 extension types the existing rules could/should be reused as there should not be a difference in static resolution of compound extension vs member extension. However in the case of more than 2 extension types it is no longer possible to reuse these rules and a new (more robust) is needed for those cases. In other words if I have an extension function that takes multiple types (`C:B` `B:A` and `A`) and is intended to have 3 receivers all of type C. how does the compiler determine the order of the receivers on invocation? And how does it even pick all three rather than just the innermost object as it satisfies the constraints.

I agree that in most cases the issue doesn't come up, but in case it does, it must be handled correctly. Currently `with` is a regular library function, so it could be syntactic sugar, but it would in its own code have the exact same problem as it is intended to solve. Making `with` a function-that-is-always-an-instruction-and-has-special-meaning is possible, but I don't like it as a solution as it also introduces a scope. I would want a way to basically call `(a,b,c).foo` and have that have the intended meaning (order for member extension functions would need to be specified).

The alternative, somehow depending on the order in which receiver candidates were introduced into the context is something that from my perspective too opaque. I'd much prefer the compiler to throw an error and make me specify in a clearly legible way what the intended receivers are, and what their order is. And casting variables to remove ambiguity is not really a desirable solution to me (although would be something that developers could not be stopped from doing).

altavir:
It is hard to focus right now, but we discussed this situation before the KEEP and I think current rules should work with it. My summary [is few posts up in this discussion](https://github.com/Kotlin/KEEP/pull/176#issuecomment-460147227). And the link in it goes to original discussion which has some additional details. If you look at discussion, there were proposals about ordered resolution, but we decided that it is not possible to match current behavior with it.

> Duplicate types or ambiguities throw compile error.

Meaning that if the declaration contains a type which is s sub-type of another type in declaration, this declaration won't compile. It is possible to construct a context with multiple receivers of the same type, but since binding is made from right to left, only the last context with matching type will be bound.

I believe that there will be some complications with parametric types like `<T> [T, B]`. In this construct T could be in fact of type `B` and we can't forbid it. But following binding procedure, it will be bound to the rightmost context in the context list matching condition, which in this case means the innermost context independent of type. Looks fine to me.

raulraja:
@chuckjaz @altavir Just wanted to point out that this proposal and https://youtrack.jetbrains.com/issue/KT-10468 overlap with some of the features proposed in KEEP-87.

https://github.com/Kotlin/KEEP/pull/87

KEEP-87 also supports an arbitrary number of injections that activate syntax over the scope but it's not restricted to extension functions. Instead associates a type with a list functions. Once the type is used in scope it makes the functions available. 

The fact there is already 3 proposals and the overwhelming traction in KEEP-87 shows these are features the community is interested in.

If you think there is an opportunity to collaborate among all three proposals we'd be happy to look into where the overlap is and adjust the proposal to reflect these needs.

altavir:
It was discussed thoroughly in community. Both approaches cover mostly the same use cases (specifically if we add file-level receivers), so in the end probably only one of them will be main-stream (or we will have two compiler plugins). 

The key difference is that while type-classes approach infer the receiver by implicit mechanism using declaration-site information, whereas multi-receiver requires explicit list of receiver objects to be passed at use-site. Both mechanisms have their own pros and cons. Type-classes do not require additional receiver declaration and always use the same context in all functions (for the same package). Multi-receiver approach could use a receiver with state (see `CoroutineScope` for example). Since it is explicit it allows much more maneuverability, like splitting receiver declaration:

```kotlin
fun [A,B].doBoth()

fun A.doB(b:B){
  // the important part here is that you can explicitly pass b as parameter 
  with(b){doBoth()}
}

```

The implicit inference  is the most controversial point of type-class proposal. If you remove it, we come to this proposal and certainly can work on improvements together. Also, the important part of this approach is that it requires almost no changes in the language (Java interop, reflections, etc) and best practices. In fact, a lot of libraries currently rely a lot on functions with receivers, and just adding a possibility of multiple receivers seems to be a logical next step.

proposals/compound-extensions.md
LouisCAD: @chuckjaz What about using the `+` character in place of the possibly ambiguous `.`?
The plus symbol on types at receiver declaration is currently not valid Kotlin, so it wouldn't break any code, and I think it'd be more clear than a dot.

LouisCAD: Example:
```kotlin
operator fun Body+String.unaryPlus(s: String) = escape(s)
```

chuckjaz: Can you create a pull request to my branch to incorporate this as an alternative syntax?

altavir: > What about using the `+` character in place of the possibly ambiguous `.`?
> The plus symbol on types at receiver declaration is currently not valid Kotlin, so it wouldn't break any code, and I think it'd be more clear than a dot.

It is better, than dot, but readability is still worse than brackets. You need to remember, that the same notation for functional types. And there should be place for parameters (and parameters in theory could also have multi-receivers).

Mishkun:
There is a plenty of discussion about ambiguities in resolution sequence and same type conflicts. Also many discussion about syntactical issues. I personally really like KEEP-87 way about thinking of that like a parameter that can be set implicitly from the enclosing scope. But the main thing I dislike in that proposal is that scope is created by the package structure or companion. Explicit scope declarations `with` and `run` are much more simple.

I propose to use soft keyword `with` before a parameter to declare that this parameter can be implicitly passed from the outer scope. Note that it says "implicitly passed from scope", not being called with receiver. Single-receiver extension functions can live the same way they live today.

```kotlin
fun A.doSomething(b: B, with c: C)

a.doSomething(b, c) // explicit pass

with(c) {
    a.doSomething(b) // implicit pass from scope
}

with(a, c) {
   doSomething(b) // extensions work normally
}

fun doSomething(b: B, with c: C)

c.doSomething(b) // ERROR with parameters can only be passed from scope

fun doSomething(with b: Int, with c: Int)

with(x, y) {
   doSomething(b = x, c = y) // ambiguities can (and must) be resolved explicitly
}
```

---
TOL: This also can make possible some interesting things like `vararg with` to capture several scopes at once.

alexdok:
Is it all about just sugaring `where` clause?

```kotlin
fun T.foo() where T: A, T: B, T: C
```

```kotlin
fun [A, B, C].foo()
```

altavir:
@Mishkun The idea has its merits, but I do not think it could work this way. First of all there is a problem with function literals. Multiple receivers could not work without them and in your proposal syntax for function literals looks rather strange: `A.(B, c with C)->Unit` with obligatory `c`. It could be done, but looks rather unnatural. 

The second problem is the precedence of receivers. In your proposal you have a primary receiver and secondary ones. As we discussed in initial thread, order-based receiver resolution conflict with the way receivers work right now. There will be a lot of confusion about how to resolve cases like `A.(b with B)` and `B.(a with A)`. Which one to take?

I also think that passing receives as parameters partially kills the idea. I really want for receivers to be contexts, not parameters. In this case I will get new behaviors from context in all functions with appropriate receivers, they will live in the micro-worlds, created by them, not only inject dependencies from environment.

We discussed `this` resolution ambiguity in initial thread, and I believe it could be done with `[A,B].()` syntax. For that you just need to add a new label to this resolution. For example like this:
```kotlin
with(a,b){ a: A, b: B->
  this@with[A]// could add special syntax like  this@with[a] for this case
  this@with[B]
}
```
or this:
```kotlin
fun [A,B].()->Unit = {
  this@with[A]
  this@with[B]
}
```

HKhademian:
I think this method of implementation is similar to current kotlin language and readable, also it's very similar how C# defines extension functions:
```
fun (Context).dp(with value: Int) = ...
```
or (but not recommended):
```
fun (Context).dp(this value: Int) = ...
```
now we can use these functions like before (`with (context) { val size = dip(16) }`) or as multiple receiver as `with(context) { val size = 16.dip() }`
I think we can even implement infix functions like  `with(context) { val size = 16 dip }`
Or multiple receiver variables like:
```
val (Context).dp get(with value: Int) = ...
// ...
// then
with(context) { val size = 16.dip }
with(context) { val size = dip(16) }

```

altavir:
@HKhademian I can't understand your message and not that familiar with C#. I do not see the second context (receiver) in your code and without it, it is hard to understand what do you mean in terms of multiple receivers.

HKhademian:
> @HKhademian I can't understand your message and not that familiar with C#. I do not see the second context (receiver) in your code and without it, it is hard to understand what do you mean in terms of multiple receivers.

I use `with` before second, third, ... context (receiver) parameters in function definition like we use `this` in c# (just for one context).

altavir:
Does it makes sense to make a primitive receiver?

There was a similar proposal from @Mishkun . And I already explained what is wrong with that. Anyway, you need to give some more detailed examples so we could see how it looks.

fatjoem:
@HKhademian is giving an example from Android development. 

In the Android API there is a function that converts device independent pixel values to physical device's pixel values. This function is an instance member of a class named Context (ambiguous in the context of this discussion).

It is commonly requested by developers to use "10.dip". This would require 2 receivers: the Context object that can do the conversion and the integer value itself.

altavir:
I have not developed for Android for a long time, but still this requirement seems  rather artificial to me. That is what I would have done:

```kotlin
class ScreenUnits(val dpSize: Int){
  val Int.dip: Int get() = this*dpSize
}
```
And then perform all screen operations in the context of this class. From architectural point of view, it would also be good to have all screen rendering features in a single class, not only density.

fatjoem:
Your first example would still require boilerplate that he wanted to get rid of.

Your last example is exactly what he wants to achieve. He just uses a different syntax for it. I am not familiar with that syntax, but apparently it comes from C€. I like your syntax more.

altavir:
The two examples are essentially the same. The difference is that you can use second one with existing classes, while first one requires some changes to the API. For example adding the function to Context.

fatjoem:
So you mean the requirement is artificial because the Android API is poorly designed? I am not sure whether that qualifies as artificial-ness, but getting off-topic...

fatjoem:
A somewhat related keep:

https://github.com/Kotlin/KEEP/pull/106

It proposes to add "extension values"  to classes (called "companion values" in the proposal). Such values would implicitly be provided as receivers to all member functions of the class.

hannespernpeintner:
I experimented a bit in #106 to implement companion vals in the compiler and i added a test there how to realize the compound receiver use case with it. It doesnt use the syntax using all receivers as prefix and i indeed think that using a single extension receiver is better, combined with a companion, or extension parameter, because semantically one wants to invoke a function/access a property on a single thing, the rest is just context. My test case also shows a slightly different way, modeling the compound receiver as an explicit class, having only its properties companions. This enables the nice dsl usage for multiple receivers, while it eliminates implicit parameter deduction from the context, because the instance must still be provided explicitly.

fatjoem:
I don't think that https://github.com/Kotlin/KEEP/pull/106 can replace compound receivers. Neither vice versa. They might have intersections but are mostly orthogonal features.

One of my main use cases for compound receivers would be top level functions. I cannot use https://github.com/Kotlin/KEEP/pull/106 for that.

In fact in my use case compound receivers would work very very well together with https://github.com/Kotlin/KEEP/pull/106 .

Right now in my project we are solving the use case with some boilerplate, and we look forward for both of these keeps. If there is interest, I can provide the use case in more detail.

hannespernpeintner:
Did you take a look at my prototype implementation? There is a test case implemented with top level function, although i extended the implementation's scope compared to the original proposal.

fatjoem:
Does your prototype allow providing extension values as the argument of top level functions? In that case it would be usable in my use case, however it would mean boilerplate at the call site.

fatjoem:
I invested the time to prepare a simplified version of my use case in which applying compound extensions as well as extension values ( https://github.com/Kotlin/KEEP/pull/106) would reduce boilerplate. Neither of both proposals would be sufficient for its own to reduce all boilerplate in this example.

```
interface I18nMessages {
   fun getMessage(key: String)
}

interface I18nContext {
    val messages: I18nMessages
    fun getMessage(key: String) = messages.getMessage(key)
}

interface UrlBuilder {
   fun getUrl(key: String)
}

interface UrlContext {
    val urls: UrlBuilder
    fun getUrl(key: String) = urls.getUrl(key)
}

class FooPageController(
    override val messages: I18nMessages,
    override val urls: UrlBuilder
) : I18nContext, UrlContext {
   
    fun handleRequest(request: Request) = FooModel(
        windowTitle = getMessage("foo.windowTitle"),
        fooHeader = getMessage("foo.headerLabel"),
        homeLink = homeLink()
    )
}

// a global function that is usable in every controller:
fun <T> T.homeLink() where T : I18nContext, T : UrlContext = Link(
    url = getUrl("urls.homePage"),
    label = getMessage("label.link.homePage")
) 

data class FooModel(
    val windowTitle: String,
    val fooHeader: String,
    val homeLink: Link
)
data class Link(val url: String, val label: String)
```

hannespernpeintner:
Thanks for the example! When interfaces only are involved, indeed everything is more simple. There wouldn't be a need for the exports feature from companion vals at all because one can use standard interface delegation.

But i think this proposal is about cases where you already have two objects and want to work with types that are not necessarily interfaces.

fatjoem:
This is a simplified version of real world example, in which we indeed have the same kind of interfaces.

The *Context interfaces exist only for the purpose shown in this example. They are interfaces because this allows to inherit code from multiple interfaces.

The I18nMessages and UrlBuilder types are interfaces because of the usual clean code considerations. Of course there are concrete implementations for them, but those concrete implementations are not known at compile time.

HKhademian:
I think a good language for today needs is a language that behave like all other todays usual tasks.

As our daily life experiences, when we are in a context, like a room,  we can do some actions on inner objects (like phone charger) that only possible if we be there, like connect it to power outlet. And we don't redefine charger when we change our context, we just define some common actions on some special objects according to its context.

I'm an Android Developer, I use Anko to build my ui layouts. when I use anko, I face some problems as simple as `dip(16)` or `view.lparams` and when I look at how anko provides some simple actions like them, I wish, kotlin has some simple solutions as always to solve it.

in Anko we have child for every common `ViewGroup`s. and why? to implement a simple function that create its special `LayoutParam` instance like `_LinearLayout { fun View.lparams(...) = LinearLayout.LayoutParams(...) }`, `_FrameLayout { fun View.lparams(...) = FrameLayout.LayoutParams}`, ...
(it's hard to write even these structural short versions, think about tens or even hundreds of classes to extends (if possible) to bring this single functionality)

there is no problem, until you define your own ui, or want to extends those `ViewGroup`s or even use a library that extends those. in that point there is no `_LinearLayout { fun View.lparam(...) }` or any other helper functions.

But if we implement this multi context receiver parameter in extension functions or even package level functions, I think we have better code, less code boilerplates, and better libraries even out of android ecosystem.

And as I mentioned before, I like these multi receiver context parameters to have their own names, because, because it can't solve some rare problems on same type receivers like `fun [View, View].pack() = ...`
Because I like C# (of course less than Kotlin), I think if we define them with some prefixes like `with` (better relation with kotlin current definition) or `this` (better meaning), we have more option to join this concept with other kotlin features like optional parameters, infix functions, ..

And for last words, I think #106 is a good idea to join kotlin community (but first need many polishes), but it's not a solution for all our pains (boilerplate codes, ...).

Think you need to mention `private companion val contextHelper = AnkoContextHelper` in every layout layers like:
```kotlin
verticalLayout {
    private companion val contextHelper1 = AnkoLinearLayoutContextHelper // <- it's an object
    // why I need to define a variable when I don't use it directly, I may discuss it on #106

    frameLayout {
        private companion val contextHelper2 = AnkoFrameLayoutContextHelper // <- it's an object too
        // what the point if we have code boilerplates again, I know, discuss in #106

        ....
    }
}
```

I recommend something like these:
```kotlin
fun Context.dip(this /* or with */ size: Number) = ...

fun ViewGroup.lparams(this view: View) = ...
fun FrameLayout.lparams(with view: View) = ...

// thinks about these beauties (There is no default context definition)
fun avatarize(this toolbar: ToolBar, this avatar: AvatarView = CircularAvatar()) = ...
// --- and use it like:
toolbar {
   
    avatarize()
    squareAvatar.avatarize()
}
// or
squareAvatar {
    toolbar.avatarize()
}

// try infix
infix fun hit(this player1: Player, this player2: Player) = ...
// --- and use it like:
mainPlayer.apply {
    hit enemy
}

// maybe triple or more receivers even vararg, of course it rare case, but who we are to ban it
fun cubeFromPoints(with vararg points: Point) = ...
```

HKhademian:
I read my codes above and I think, if **multi receiver context parameter extension functions** (what a like title for a tiny feature 😃 ) with no default receiver as `avatarize` function I defined above,has limits on how we call them, on overall codes has cleaner schema and easier usage:

new version (but not sure):
```kotlin
toolbar {
    avatarize()
    avatarize(squareAvatar)
    avatarize(newToolbar, squareAvatar)

    squareAvatar.avatarize()
    squareAvatar.avatarize(otherToolbar)
}

squareAvatar {
    toolbar.avatarize()
    toolbar.avatarize(someOtherAvatar)

    avatarize(toolbar)
    avatarize(toolbar, someOtherAvatar)

    someOtherAvatar.avatarize(toolbar)
    someOtherAvatar.avatarize()
}
```

I think we can have multi stage to implement #176 (multi receiver context parameter extension functions).
First add simple definitions, then add these extra features to it.

fatjoem:
I updated my example to take advantage of compound extension as well as companion objects:
https://github.com/Kotlin/KEEP/pull/176#issuecomment-516461789

fatjoem:
I believe that this keep together with compound extension functions ( https://github.com/Kotlin/KEEP/pull/106) could support many of the use cases that are given to motivate type classes ( https://github.com/Kotlin/KEEP/pull/87).

I added an example that illustrates this in that keep:
 https://github.com/Kotlin/KEEP/pull/87#issuecomment-516613955

netvl:
One example where this could be useful is extending certain parts of Gradle builds. For example:
```kotlin
// Suppose we want to abstract adding multiple dependencies of the same "kind"
// whose version is defined in the project properties, or otherwise derived from
// the `Project` instance
val groupVersion: String by project.extra
dependencies {
    listOf("name1", "name2", "name3").forEach {
        implementation("group", it, groupVersion)
    }
}

// It'd be nice to be able to abstract the above like this:

dependencies {
    configurations.implementation(groupDependencies)
    // while also allowing other configurations to be used:
    configurations.testRuntimeOnly(groupDependencies)
}

// Unfortunately, at this moment it is not possible to do this nicely without
// multi-receiver extensions, because constructing and adding a dependency
// would require three objects: `Project`, `DependencyHandler` and `Configuration`:

class Dependencies(scope: DependencyHandlerScope) {
    fun configure(project: Project, scope: DependencyHandler, configuration: Configuration)
}

val groupDependencies: Dependencies = object : Dependencies {
    override fun configure(project: Project, scope: DependencyHandler, configuration: Configuration) {
        val groupVersion: String by project.extra
        scope {
            listOf("name1", "name2", "name3").forEach {
                configuration("group", it, groupVersion)
            }
        }
    }
}

fun Project.DependencyHandler.Configuration.invoke(dependencies: Dependencies) {
    dependencies.configure(this@Project, this@DependencyHandler, this@Configuration)
}
```

There are certainly other ways to do this, but none (as far as I can see) would allow the "native" syntax like `<configuration>(<dependency details>)`.

altavir:
Oleg Yukhnevich just presented an example which is currently working, but would be broken on a non-ordered multi-receiver resolution:

```kotlin
object A {
    fun B.ext()= println("AB")
}

object B{
    fun A.ext() = println("BA")
}

fun main(){

    with(A) {
        B.ext()
    }
    
    with(B) {
        A.ext()
    }
}
```

There are two solutions to that problem:
* Just throw compile-time exception if function have an ambiguous binding as it stated in bounding rules. The given code is can't have a lot of usages since it is obviously bad.
* Add additional rule for special treatment of first receiver in case it is a member receiver. It could require a lot of rule tweaking.

I prefer the first variant.

proposals/compound-extensions.md
```diff
@@ -0,0 +1,380 @@
...
+val a: extension (A, B, C).(Int)->Int = ...
+```
+
+Using a pseudo keyword avoids the ambiguity of the `(` in the type expression. A pseudo-
+keyword could also be used with the `[`...`]` syntax to leave a unadorned `[`...`]` to
+mean a tuple in the future.
+
+### Matching rules
+
+1. **Create an ordered list of implicit receivers** in scope with types
```
DavidDTA, line 302: I would like to see support for explicit receivers for multi-receiver extensions.  Existing extensions can be called with an explicit or implicit receiver, and I think there are cases where explicit receivers could be important for multi-receiver extensions.

As noted below, receivers may be "explicitly" provided with the following syntax, but it extremely verbose and not obvious:
```
with (a) {
    with (b) {
        with (c) {
            doSomething()
        }
    }
}
```

proposals/compound-extensions.md
```diff
@@ -0,0 +1,380 @@
...
+    operator fun Body.String.unaryPlus() = escape(this)
+    ```
+
+    therefore allowing unary `-` to be provided generally by extension of `Body`,
+
+    ```kotlin
+    operator fun Body.String.unaryMinus() = underline { +this@unaryMinus }
+    ```
+
+## Proposal
```
DavidDTA, line 234: This proposal does not address limitations regarding two receiver types being related.  Would it be possible to define an extension with two receivers of the same type, and if so, how would they be disambiguated?

proposals/compound-extensions.md
pdvrieze: What about:
```
with (a,b,c) { doSomething() }
```
Where the receivers match the order of the receivers of `doSomething`

proposals/compound-extensions.md
altavir: This question is answered in the  discussion starting at https://github.com/Kotlin/KEEP/pull/176#issuecomment-460147227

altavir:
@DavidDTA There was a lot of discussions about the proposal and there are actually several different variants of it in the discussion. Sadly, the initial text is not updated anymore. And KEEP is now officially not the place for such discussion. It should continue at https://youtrack.jetbrains.com/issue/KT-42435.

elizarov:
Thanks a lot for a fruitful discussion on this issue. The use-case of writing this kind of "compound extensions" in Kotlin is important. 

However, this specific design proposal does not meet the basic requirements that were identified for this kind of feature in Kotlin. There is no specific design yet to address the use-cases that are covered here, but there is a YouTrack issue that tracks this problem and has some basic requirements that a design shall meet that were identified thus far: https://youtrack.jetbrains.com/issue/KT-10468 Please, vote for it, comment on your specific use-cases, and continue discussions there 👍 

With this, I'm closing this KEEP issue.

PR #177, opened 2019-01-13
Clarifying non obvious syntax with cryptic error

theaspect:


nikitabobko:
Thanks, merged manually

PR #182, opened 2019-02-08
KotlinX Dates and Times Library

UnknownJoe796:
Presents a possible API for a date/time library.
`proposals/extensions/dates-and-times.md`

### DIFF
--- a/proposals/extensions/dates-and-times.md
+++ b/proposals/extensions/dates-and-times.md
@@ -1,0 +1,216 @@
+# Dates and Times
+
+* **Type**: Standard Library API proposal
+* **Author**: Joseph Ivie (Lightning Kite)
+* **Status**: Submitted
+* **Prototype**: In progress
+
+
+## Summary
+
+This is intended to be a simple, not all-encompassing implementation of date and time functionality that leaves room for extension.  Only the most common elements of managing dates and times are written here, and additions to this proposal should be in other proposals.
+
+- `kotlin.time.Date`, representing a particular day, with no concept of time zones.  Internally an inline class holding the number of days since January 1st, 1970.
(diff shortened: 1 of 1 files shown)


Wasabi375:
Firstly, I don't think it's a good idea to introduce a system like this without adding real support for timezones at the same time, otherwise this will just create an API with too many problems. 
One example for this (already present) is that adding a duration to a `DateTime` without proper knowledge of timezones is not possible. One example for that is daylight saving time. `1:59 am + 1 minute` could be either `2:00 am` or `1:00am`.  
I believe that it would be better to design a proper API for dates, times, etc which supports everything needed in most programs, including time zones and probably even support for different calendar systems. 
I personally don't know Klock so I can't comment about whether it is any good, but if it is any effort for Kotlin Multiplatform Date/Time should be focused there. No need to create 2 rival open source projects solving the same problem when we can instead create one good solution.
If not I suggest we follow an API design like [Noda Time](https://nodatime.org/)
(Also can you add a link to it in the KEEP: https://github.com/korlibs/klock)
___
That all being said there is a great talk about this topic by Jon Skeet: https://youtu.be/saeKBuPewcU
He is the main developer of [Noda Time](https://nodatime.org/) a .Net Time library.
___
2 small things: The proposal is missing a representation of a time interval, which differs from a duration in that it contains the start and endpoint. 
And I don't understand the exact functionality of those 2 functions in `Date`
```kotlin
fun toNextDayOfWeek(value: DayOfWeek): Date
fun toDayInSameWeek(value: DayOfWeek): Date
```

gildor:
Completely aggree with @Wasabi375 just want to add that I would against adding this to support library, even if such library will be developed and supported by Kotlin Team I believe it should be a separate library, like kotlinx.time rather than addition to stdlib as it proposed

PR #185, opened 2019-04-11
KEEP rewording

JorgeCastilloPrz:
* [x] Renamed it to something more in the line of its final purpose.
* [x] Reworded everything to not include anything related to FP but refer to compile time extension resolution. Added more details and removed some sections not applicable anymore.
* [x] Included all details we had in our Kotlin fork PR description but not here, regarding how to test the keep or how error reporting looks like.
* [x] Added our names to the list.
* [ ] Add all caveats / todos to [the required section](https://github.com/Kotlin/KEEP/blob/c811a73b0dd5f09b73e4548665d06d8f8063f603/proposals/compile-time-dependency-resolution.md#whats-still-to-be-done). I've just added a single one we had on the keep already. Can you help on that @truizlop @raulraja ?


JorgeCastilloPrz:
I'm sorry, this is not the place where I wanted to open it. We'll update #87 with this info soon.

Issue #190, opened 2019-06-06
Duration and time measurement API


ilya-g:
This issue is to discuss the proposal to introduce [`Duration`, `Clock` and time measurement API](https://github.com/Kotlin/KEEP/blob/durations/proposals/stdlib/durations-and-clocks.md).

JakeWharton:
Can you detail why the stdlib is the target for this API rather than a standalone kotlinx library?

I don't think there's a clear delineation between when something should target the stdlib vs. kotlinx. Coroutines were split between the two, but leaned aggressively away from the stdlib in everything but the fundamental building blocks. Looking ahead, the multiplatform file/fs abstraction seems to be targeting kotlinx entirely. Nothing about the proposal makes it obvious why the stdlib is where this should live. I'm not opposed to it being there, to be clear, but I think it's worth capturing the arguments as to why something should live in the stdlib vs. kotlinx, applying them to this proposal, and seeing whether it still makes sense.

fluidsonic:
As far as I understand this KEEP is mostly about duration and timing related to measurement, timeouts etc. and not for real-world date/time application.

I suggest to be careful with the naming (e.g. `Duration`, `Clock`) because it could conflict with a potential date/time library in the future if it intends to implement these in a different way or with different functionality but will have to pick another name. This can become a source of confusion.

A `Duration` for measuring something short-term maybe work differently and have different functionality as a `Duration` used for representing real-world date/time differences. Same for `Clock` which may use completely different approaches and offer different functionality in each case.

Wasabi375:
Well some possible names to solve this kind of conflict would be `Stopwatch` and maybe `MeasuredDuration`, `MeasuredTimeInterval`. While I kind of like the name `Stopwatch` I think `MeasuredDuration` might be to verbose. 
But then how often is it necessary to both deal with monotonic time and real world time. It might be simply enough the different "timing" implementations in different packages (as they would be anyways). Worst case you can always add typealiases for the implementation you use less.

markov:
FWIW I made a little lib a while ago that implements a large part of this proposal for the `Duration` class: https://github.com/markov/lucid-time

The major difference is that I chose the internal value to be a `Long` not a `Double` - I feel like most applications do not need to represent time differences larger than 292 years. Also this is consistent to how java's own `TimeUnit` enum would handle unit transformations and I wanted to have 100% same results if someone chose to replace all usages of `TimeUnit` with my lib.

[...]

PR #191, opened 2019-06-06
Proposal: Duration and time measurement API

ilya-g:
Discussion goes in [KEEP-190](https://github.com/Kotlin/KEEP/issues/190)

### DIFF
--- a/proposals/stdlib/durations-and-clocks.md
+++ b/proposals/stdlib/durations-and-clocks.md
@@ -2,8 +2,8 @@
 
 * **Type**: Standard Library API proposal
 * **Author**: Ilya Gorbunov
-* **Status**: Experimental in Kotlin 1.3.x
-* **Prototype**: In progress
+* **Status**: Experimental in Kotlin 1.3.50
+* **Prototype**: Implemented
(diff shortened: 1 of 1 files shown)


proposals/stdlib/durations-and-clocks.md
```diff
@@ -0,0 +1,389 @@
...
+- `isNegative()`, `isFinite()`, `isInfinite()` functions
+
+#### Equality and comparison
+
+The equality of Duration is defined so that two duration values representing time intervals of the same length
+are equal, no matter which numeric types and duration units were used to construct these durations, e.g.:
+
+```kotlin
+1.days == 24.hours
+1.5.minutes = 90.seconds
```
pavlospt: Needs another = here I think!

ilya-g: Thanks, good catch. Fixed.

elizarov:
It would seem extremely useful to add `String.toDuration()` extension that can reverse both the results of `Duration.toString()` call and the result of `Duration.toIsoString()`.

ilya-g:
@elizarov Please use #190 for discussion. 
Is it ok that `Duration.toString().toDuration()` would be lossy conversion then?

PR #194, opened 2019-10-01
Named arguments in their own position

dzharkov:


fvasco:
What is the use case behind this KEEP?
What issue it solves?

> But in some cases, such limitation might be inconvenient.

Mixing positional and named arguments does not look really readable for me, the proposal examples look too abstract.
Why `f(x = 1, y = 2)` in the first paragraph isn't good?

dzharkov:
The main use is omitting names for subsequent arguments when the names are actually obvious on call-site like:
```kotlin
someFunction(runInTheSameThread=true, someOtherName)
```

It looks better and more clear to me than `someFunction(runInTheSameThread=true, someOtherName=someOtherName)` that becomes boilerplate

Issue #45, opened 2016-08-20
Explicit mode for library authoring 


ilya-g:
## Motivation

There were a couple of hot discussions about the default visibility before release:
https://discuss.kotlinlang.org/t/public-by-default-for-classes/110
https://discuss.kotlinlang.org/t/kotlins-default-visibility-should-be-internal/1400

The main concern against public-by-default visibility was that it becomes too easy for library authors to expose something accidentally, release it, and then have to make breaking changes to hide it back.

## The proposal

The proposal is to provide an 'Explicit API' mode in the compiler which helps library authoring. Such mode should prevent delivery of unintended public API/ABI to the clients by requiring explicit visibility modifier and explicit return types for public declarations.

See the full proposal text in https://github.com/Kotlin/KEEP/blob/master/proposals/explicit-api-mode.md

cypressious:
Have you considered including open/final in the explicit mode?

ilya-g:
@cypressious,
Current situation with final-by-default is safe from the point of library authoring, one have to be explicit to make a class more accessible/extensible. 
What benefits do you expect explicit final would bring?

cypressious:
The main benefit is choice. Maybe some library authors don't want final by default. See https://www.reddit.com/r/programming/comments/4yi4ch/kotlin_the_good_the_bad_and_the_ugly/ for a controversial discussion about this topic.

JakeWharton:
Are those library authors? Or are they application developers complaining
about libraries because they've never had to maintain one of non-trivial
popularity. I'm betting the latter.

ilya-g:
I think there should be exceptions even in the explicit `public` mode, that can have public omitted and still have it by default:
- properties, defined in primary constructor: `public class Foo(val bar: T)`
- constructors: `public class Foo() { constructor(...) {} }`
- companion objects: `public class Foo { companion object { } }`

What do you think?

mikehearn:
It might be nice if the compiler would emit a warning in "api mode" when you have something public which doesn't have a kdoc (or doesn't inherit one)

wyand:
There are good reasons mentioned here to enforce some constraints on code style -- but I think adding these directly to the compiler as options is a step in the wrong direction.

[...]

PR #197, opened 2019-10-16
Explicit api mode

sandwwraith:


proposals/explicit-api-mode.md
```diff
@@ -0,0 +1,108 @@
...
+
+Introduce 'Explicit API' compiler mode. Compilation in such mode differs from the default mode in the following aspects:
+
+* Compiler requires you to specify explicit visibility for a declaration when leaving default visibility would result in exposing that declaration to the public API surface.
+
+* Compiler requires you to specify the explicit type of property/function when it is exposed to the public API surface.
+
+* Compiler warns you when exposed to public API surface declaration does not have a KDoc.
+
+See the definition of public API [here](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/binary-compatibility-validator/ReadMe.md#what-constitutes-the-public-api).
```
elizarov: I'd suggest to give more details here. In particular I would clearly distinguish "Public API" vs "Published API" where "Published API" should be defined as "Public API" + "internal @PublishedApi" declarations. Why? Because "Published API" itself should not require KDocs. Only when its part of "Public API" then should require KDocs. 

Also, as a stylistic note the world "surface" looks redundant in text. I suggest to systematically replace phrases like "public API surface" with a simpler "public API"

swankjesse:
Some Kotlin changes are source-compatible but binary-incompatible. For example, adding a parameter with a default value is binary-incompatible on JVM. Is it appropriate for this new mode to help with such problems?

ilya-g:
@swankjesse This mode is intended to analyze a snapshot state of the code at the moment of compilation, so it cannot help with detecting source- or binary-incompatible changes. 
We usually solve this task by dumping public API after compilation and comparing it with the reference dump stored in a VCS beside the project.

Please use KEEP #45 for further discussion.

proposals/explicit-api-mode.md
```diff
@@ -0,0 +1,148 @@
...
+* **Type**: Design proposal
+* **Author**: Ilya Gorbunov
+* **Contributors**: Leonid Startsev, Roman Elizarov
+* **Status**: Prototype
+* **Original proposal and discussion**: [KEEP-45](https://github.com/Kotlin/KEEP/issues/45)
+
+## Synopsis
+
+Provide an 'Explicit API' mode in the compiler which helps library authoring.
+Such mode should prevent delivery of unintended pubic API/ABI to the clients by requiring explicit visibility and return types on declarations.
```
ilya-g:
```suggestion
Such mode should prevent delivery of unintended public API/ABI to the clients by requiring explicit visibility and return types on declarations.
```
that's a quite different kind of api :)

proposals/explicit-api-mode.md
```diff
@@ -0,0 +1,148 @@
...
+    While it is possible to make this functionality as a separate plugin, the lack of stability and consistency in the compiler plugin API (because there is no official public definition of plugin API/compiler surface, actually) would require a lot of efforts to support this relatively small feature.
+
+2. External tool
+
+    An external tool like a style checker may look like a proper solution for this problem, but it requires much more additional setup than one DSL parameter, which pushes adoption back. Another big problem is that such tools re-analyze whole code one more time, which leads to increased build times, big size of the tool, or incorrect results.
+
+## Compatibility
+
+The explicit mode doesn't change the semantics of the correct code and does not affect bytecode generation, so it's safe from the standpoint of compatibility.
+For example, if some code was compiled once in explicit mode and later without explicit mode, the result should be the same.
```
ilya-g:
```suggestion
For example, if some code was compiled once in the explicit mode and later without the explicit mode, the result should be the same.
```

proposals/explicit-api-mode.md
```diff
@@ -0,0 +1,148 @@
...
+
+However, this mode could make previously correct code to compile with errors (the code that had default visibility declarations and inferred types in public API).
```
ilya-g:
```suggestion
However, this mode could make previously correct code to compile with errors (the code that had the default visibility declarations and inferred types in public API).
```

proposals/explicit-api-mode.md
```diff
@@ -0,0 +1,154 @@
...
+Explicit mode enabled here would not affect test sources.
+
+The exact place of these methods (top-level `kotlinOptions`, or compilation, or particular source set) is TBA. Maven plugin option is TBD.
+
+Kotlin plugin in IntelliJ IDEA would recognize that explicit mode is enabled and would offer corresponding intentions and quick-fixes.
+
+The following **IDE** inspections would be disabled either because they're redundant or duplicated in the compiler:
+
+* 'Redundant visibility modifier' for public API declarations
+* 'Public API declaration has implicit return type' (would be replaced with compiler diagnostic)
```
diego-gomez-olvera: It would be good to relax this restriction for `const val` types, as they are only used for primitives and `String`. I created [an issue](https://youtrack.jetbrains.com/issue/KT-30512) about this topic some time ago.

proposals/explicit-api-mode.md
```diff
@@ -0,0 +1,154 @@
...
+
+Introduce 'Explicit API' compiler mode. Compilation in such mode differs from the default mode in the following aspects:
+
+* Compiler requires you to specify explicit visibility for a declaration when leaving default visibility would result in exposing that declaration to the public API.
+
+* Compiler requires you to specify the explicit type of property/function when it is exposed to the public/published API.
+
+* Compiler requires you to explicitly [propagate experimental status](https://kotlinlang.org/docs/reference/experimental.html#propagating-use) for functions which contain experimental types in the signature.
+
+* Compiler warns you when exposed to public API declaration does not have a KDoc.
```
elizarov: "exposed to public API declaration" -> "declaration exposed to public API"

proposals/explicit-api-mode.md
```diff
@@ -0,0 +1,154 @@
...
+
+A class is considered to be effectively public if all of the following conditions are met:
+
+ - it has one of the following visibilities in Kotlin:
+    - no visibility (means no Kotlin declaration corresponds to this compiled class)
+    - *public*
+    - *protected*
+ - it isn't a local class
+ - it contains at least one effectively public member, in case if the class corresponds
+   to a kotlin *file* with top-level members or a *multifile facade*
```
elizarov: I don't see how the item on faced classes is relevant to the KEEP. We don't enforce "explicit public" or "explicit KDocs" on _file facades_ here. I suggest removing this item.

proposals/explicit-api-mode.md
```diff
@@ -0,0 +1,154 @@
...
+* Compiler warns you when exposed to public API declaration does not have a KDoc.
+
+### Public API definition
+
+#### Classes
+
+A class is considered to be effectively public if all of the following conditions are met:
+
+ - it has one of the following visibilities in Kotlin:
+    - no visibility (means no Kotlin declaration corresponds to this compiled class)
```
elizarov: > (means no Kotlin declaration corresponds to this compiled class)

Let's rephrase or remove this part in parenthesis here and in the section on members below.

proposals/explicit-api-mode.md
```diff
@@ -0,0 +1,154 @@
...
+
+ - it has one of the following visibilities in Kotlin:
+    - no visibility (means no Kotlin declaration corresponds to this class member)
+    - *public*
+    - *protected*
+
+    > Note that Kotlin visibility of a field exposed by `lateinit` property is the visibility of its setter.
+
+ - in case if the member is protected, it is contained in *non-final* class
+ - it isn't a synthetic access method for a private field
```
elizarov: It is not relevant here since synthetics are not present in the source.

proposals/explicit-api-mode.md
```diff
@@ -0,0 +1,154 @@
...
+2. External tool
+
+    An external tool like a style checker may look like a proper solution for this problem, but it requires much more additional setup than one DSL parameter, which pushes adoption back. Another big problem is that such tools re-analyze whole code one more time, which leads to increased build times, big size of the tool, or incorrect results.
+
+## Compatibility
+
+The explicit mode doesn't change the semantics of the correct code and does not affect bytecode generation, so it's safe from the standpoint of compatibility.
+For example, if some code was compiled once in the explicit mode and later without explicit mode, the result should be the same.
+
+However, this mode could make previously correct code to compile with errors (the code that had the default visibility declarations and inferred types in public API).
```
elizarov: "could make previously correct code ~to~ compile with errors"  (remove "_to_" here)

Issue #95, opened 2017-12-27
Opt-in requirements (aka Experimental API support)


udalov:
Discussion of the proposal:
https://github.com/Kotlin/KEEP/blob/master/proposals/experimental.md

LouisCAD:
Looks good! I'm myself developing a few experimental libraries, [including one already released for BluetoothGatt relying on kotlinx.coroutines](https://github.com/Beepiz/BleGattCoroutines), and this would be really useful.

cy6erGn0m:
There is "Same module exemption" but what if my library consist of multiple modules? Is there way to avoid propagation requirement except suppressing an error?

robstoll:
It's a bit late, so dare with me if I overlooked a phrase. I couldn't find how *signature usage* and *body usage* differ in propagation

robstoll:
Another question to clear up propagation. The proposal does not mention which access levels need propagation. I assume it is public and protected (where protected makes only sense for an open type), right?

udalov:
@cy6erGn0m 
> There is "Same module exemption" but what if my library consist of multiple modules? Is there way to avoid propagation requirement except suppressing an error?

Generally, if you're using an experimental API, you must propagate it. Otherwise, your clients will not be aware of the fact that something may break, and may break unexpectedly. From this perspective, it's irrelevant whether two modules (where a module is a single self-contained compilation unit) comprise a library in any sense, or not.

udalov:
@robstoll 
> It's a bit late, so dare with me if I overlooked a phrase. I couldn't find how *signature usage* and *body usage* differ in propagation. Or lies the only difference in *same module exemption*?

There are two key differences between body and signature usages, highlighted in the proposal:
1) body usages of _compile-time_ experimental API need not be propagated
2) body usages need not be propagated in the same module

> Another question to clear up propagation. The proposal does not mention which access levels need propagation. I assume it is public and protected (where protected makes only sense for an open type), right?

Visibility modifiers have no effect on the propagation requirements. The reason is that even a private declaration that uses something experimental might be called indirectly from client code, via some public declaration:

```
// Library code

@ShinyNewAPI
fun foo() {}

// If we do not require propagation here...
private fun bar() {
    foo()
}

// ... then it's not required here...
fun baz() {
    bar()
}

// Usage

fun usage() {
    // ... and then we get effectively experimental behavior without any explicit opt-in from the user
    baz()
}
```

cy6erGn0m:
@udalov 
[...]

PR #201, opened 2019-12-16
Rename "Experimental API support" -> "Opt-in requirements" (#95)

udalov:
The annotations are renamed too: `Experimental` -> `RequiresOptIn`,
`UseExperimental` -> `OptIn`.

This is needed to facilitate usage of this mechanism in other related
areas, such as restricted, internal, obsolete APIs etc. See KT-26216.

Also, add message to RequiresOptIn, see KT-28872.


sellmair:
Very welcome change 😍

proposals/experimental.md
```diff
@@ -34,100 +36,113 @@ package kotlin
...
+Using `OptIn` with annotations that are *not* opt-in requirement markers has no effect and yields a compilation warning. (Note that this must not be an error because user code should not break once an annotation is no longer an opt-in requirement marker.) Using `OptIn` with no arguments has no effect and yields a warning as well.
+
+## Opt-in requirement message
+
+When using an API that requires opt-in without the said opt-in, the compiler reports a warning or an error depending on the specified level. It's possible to specify a custom message in `RequiresOptIn`, that will be reported by the compiler. If no message is given (i.e. if `message` is empty), the compiler will report that the API is *experimental*:
+
+```
+test.kt:22:9: error: this declaration is experimental and its usage must be marked with '@ShinyNewAPI' or '@OptIn(ShinyNewAPI::class)'
```
qwwdfsad: Why does the default message mention experimentality?
E.g. why not
>est.kt:22:9: error: this declaration required opt-in and its usage must be marked with '@ShinyNewAPI' or '@ OptIn(ShinyNewAPI::class)'

ilya-g: Perhaps, to make the migration from `@Experimental` more smooth. 
However we need to think about the message template, so that one who needs a custom message would not have to spell `marked with @Annotation or @OptIn(Annotation::class)` every time.
Let's discuss it later in the corresponding issue, before we start implementing the message support.

proposals/experimental.md
```diff
@@ -34,100 +36,113 @@ package kotlin
...
+Therefore, we will **require** each user of `RequiresOptIn` to provide at least one unstable (`-X...`) compiler argument, which would mean that the user is understanding the risks of using this experimental functionality. It can be either `-Xopt-in=...` with any opt-in requirement marker, or the magic predefined argument `-Xopt-in=kotlin.RequiresOptIn` which doesn't allow using any API by itself, yet merely allows using `RequiresOptIn` and `OptIn` in the source code. Unless one of these arguments is provided, the compiler will report a warning on each usage of `RequiresOptIn` or `OptIn` (but not on usages of the markers!).
```
ilya-g: > to provide at least one unstable (`-X...`) compiler argument

Is any `-X...` argument enough, e.g. `-Xinline-classes`?

proposals/experimental.md
udalov: @qwwdfsad I thought that experimental API is still the main use case and "is experimental" by default is a bit clearer for the user than "requires opt-in". But I'm not very sure about this and can be convinced otherwise.

@ilya-g As far as I remember, we discussed this and explicitly decided to fully replace the message (unlike in Deprecated). See the internal discussion [doc](https://jetbrains.quip.com/Ga4lA7rtaF5Z/DM-Apr-18-2019-Completion-customization-for-experimental-API#XHSACAiFJKj):
> * Add ability to customize diagnostic message in experimental annotations (like in `@Deprecated`). That message will fully replace default message

But if you have a proposal for a message template, let's discuss it.

proposals/experimental.md
udalov: No, it was poorly worded, fixed.

PR #211, opened 2020-04-14
Add a way to print to stderr

nkniep:
[Rendered](https://github.com/nkniep/KEEP/blob/stderr-print/proposals/stdlib/stderr-print.md)


CLOVIS-AI:
What would happen in JS when the function is called?

elizarov:
This request is tracked as [KT-6064](https://youtrack.jetbrains.com/issue/KT-6064) Introduce standard(platform independent) way to print stderr. I'm closing this PR and moving discussion there.

PR #213, opened 2020-05-09
Pattern matching

Cottand:
## Proposal
The goal of this proposal is to bring full blown pattern matching to Kotlin. 

Pattern matching is a tried and tested feature of many languages, such as Haskell, Rust, Scala, Ruby, and in the near future, Java.

In a nutshell, it is a quick and intuitive way of checking whether an object is of some type, and to extract and check its contents. The propsal aims to do this through the existing `is` and destructuring `componentN()` semantics.

Pattern matching is a major addition to a language, and as such several aspects of the proposal are up to debate. Discussion and contributions are welcome. 

Here is a short example of what Kotlin with pattern matching could look like:

```kotlin
sealed class Download
data class App(val name: String, val developer: Person) : Download()
data class Movie(val title: String, val director: Person) : Download()
val download: Download = // ...

// Without pattern matching:
val result = when(download) {
  is App -> {
    val (name, dev) = download
    when(dev) {
      is Person -> 
        if(dev.name == "Alice") "Alice's app $name" else "Not by Alice"
      else -> "Not by Alice"
    }
  }
  is Movie -> {
    val (title, diretor) = download
    if (director.name == "Alice") {
      "Alice's movie $title"
    } else {
      "Not by Alice"
    }
  }
}

// With pattern matching:
val result = when(download) {
  is App(name, Person("Alice", _)) -> "Alice's app $name"
  is Movie(title, Person("Alice", _)) -> "Alice's movie $title"
  is App, Movie -> "Not by Alice"
}
```

For more details, you can read a rendered version of the proposal [here](https://github.com/Cottand/KEEP/blob/pattern-matching/proposals/pattern-matching.md).


hepin1989:
With `case` not `is`, or `match`

Cottand:
I purposefully aimed to not introduce a new keyword (like `match` ), because that would be a breaking change

rnett:
A few opinions/options:
- Use `val`/`var` to declare a new variable.  E.g. `is Person("Bob", val age)`.

- Guards can be done relatively easily using `continue` in an `is` block, I don't see the need to add them to the matching statement.  E.g. `is Person("Bob", val age) -> if(age < 18) continue else ...`.  Its even a one liner.

- Where I do think guards would be nice is "inline".  Something like `is Person(guard{ it.matches("...") }, val age guard{ it >= 18 })`.  Preferably for captured and non-captured values.  This would also take care of the reasons for the explicit syntax, although `guard{ it <= 3 }` is a bit more verbose.  For things like regex, case-insensitive strings, `startsWith`, `% K == 0`, etc, you could end up with a lot of guards for different variables, and something like this is more readable than doing them all in a guard at the end.  It also helps with collections: `is Person(contacts: guard{ it.all{ ... } })` where `contacts` is a List.  Syntax magic to make it look like `is Person(it.matches("..."), val age >= 18)` would be cool, but could end up making things more ambiguous (and also seems like it would be hard to implement).

- Named fields: allow things like `is Person(name: "Bob", age: val x)`.  Would be limited to occurring after the numbered/`componentN` fields like named arguments.  The names are just the names of fields, anything that would resolve after a dot works.  This allows for using pattern matching for things that don't have `componentN` functions (including collection sizes).

Miha-x64:
'guards' could use the same syntax that `when (subj)` uses on the left side of ->:
`is Person(…, val age in 18..Int.MAX_VALUE) ->`. Another option is to allow `if` as Rust does: `Person(…, age) if { age >= 18 } =>`

hepin1989:
> 'guards' could use the same syntax that `when (subj)` uses on the left side of ->:
> `is Person(…, val age in 18..Int.MAX_VALUE) ->`. Another option is to allow `if` as Rust does: `Person(…, age) if { age >= 18 } =>`
```when (x) if patten```
```when (x) match patten```
```when (x) case patten```
I think anyway, it's not `is`.

**Comment by Cottand** [comment]
> * Use `val`/`var` to declare a new variable.  E.g. `is Person("Bob", val age)`.

I did consider using `val`, and it would certainly solve the anbiguity problem. But if you try to write a long match with several destructured arguments, you will find it actually adds _a lot_ of verbosity. I can still try to add it in the proposal though maybe making declaration more explicit is a better idea after all.

> *Guards can be done relatively easily using `continue` in an `is` block, I don't see the need to add them to the matching statement.  E.g. `is Person("Bob", val age) -> if(age < 18) continue else ...`.  Its even a one liner.

I agree in that using one-line `if`s looks a lot like guards, but I do not think it is an option. In particular, it leads to a nested `if` inside the `when`, which is exactly what this proposal aims to avoid.

> * Where I do think guards would be nice is "inline".  Something like `is Person(guard{ it.matches("...") }, val age guard{ it >= 18 })`.  Preferably for captured and non-captured values.  This would also take care of the reasons for the explicit syntax, although `guard{ it <= 3 }` is a bit more verbose.  For things like regex, case-insensitive strings, `startsWith`, `% K == 0`, etc, you could end up with a lot of guards for different variables, and something like this is more readable than doing them all in a guard at the end.  It also helps with collections: `is Person(contacts: guard{ it.all{ ... } })` where `contacts` is a List.  Syntax magic to make it look like `is Person(it.matches("..."), val age >= 18)` would be cool, but could end up making things more ambiguous (and also seems like it would be hard to implement).  

That is a neat idea, but it does introduce a new keyword. How would it look like for more complex nested patterns? You could make a PR into the proposal under the Guards section, where you discuss this possible alternative :).

> * Named fields: allow things like `is Person(name: "Bob", age: val x)`.  Would be limited to occurring after the numbered/`componentN` fields like named arguments.  The names are just the names of fields, anything that would resolve after a dot works.  This allows for using pattern matching for things that don't have `componentN` functions (including collection sizes).

Another neat idea, and I did think about this one writing the proposal. I really like the idea of matching by name on classes that may or may not have `componentN()`. At the same time though, I cannot come up with a clear syntax. I think the one you suggested is a bit inappropriate in that it uses `:`, which in Kotlin has always been a 'is of type' relationship. I thought about using the same syntax as named parameter passing (what you did but using `=` instead of `:`) but `name = val x` just looks weird in Kotlin too. Doing it the other way around would mean using the named parameter on the RHS which also looks weird. I also think that it makes an already complex feature more convoluted still! But it does not 'break' the already proposed syntax, so it could be added later down the line. You could add it as an idea under the Beyond the Proposal section.

**Comment by Cottand** [comment]
> 'guards' could use the same syntax that `when (subj)` uses on the left side of ->:
> `is Person(…, val age in 18..Int.MAX_VALUE) ->`. 

Could you elaborate on how this would look if the guard was not a `contains` call? Like just using the `>` operator.

>Another option is to allow `if` as Rust does: `Person(…, age) if { age >= 18 } =>`

This semantically the same as the current proposal, but with `if` instead of `where`.

**Comment by Cottand** [comment]
> `when (x) if patten`
> `when (x) match patten`
> `when (x) case patten`
> I think anyway, it's not `is`.

`if` is largely not appropriate since it should be followed by a `Boolean` as it is in the rest of the language. Otherwise that is plain confusing.
Your other alternatives, again, introduce new keywords, which is a breaking change.

**Comment by fvasco** [comment]
Hi @Cottand,

> I did consider using val, and it would certainly solve the anbiguity problem. But if you try to write a long match with several destructured arguments, you will find it actually adds a lot of verbosity.

To solve ambiguity, we do not add verbosity, we add clarity.
A short code, hard to read, is the goal for a regular expression pattern, but I hope this isn't a good metric in the Kotlin language.

I suggest to mark every new value declaration with `val`, as `when` does.
Moreover, we should avoid any distinction between identifier and constant. In `is Success(x)`, the `x` must always be a valid identified: parameter, attribute, variable or a `const val` one; this code have to be clear even without read the entire kt file and the imported ones.

> An idiom in Haskell or Scala is to pattern match on collections.

Who cares?
How "Matching on collections" should help a Kotlin developer?
In your example, `mySum*` isn't `tailrec` and can be efficently replaced with a `fold`.

Finally the references to JEP 375 and "Pattern Matching for Java" are wrong for me. Those cover the Kotlin's smart cast.
Java does not plan to implement this kind of pattern mathing.

**Comment by rnett** [comment]
> > * Use `val`/`var` to declare a new variable.  E.g. `is Person("Bob", val age)`.
> 
> I did consider using `val`, and it would certainly solve the anbiguity problem. But if you try to write a long match with several destructured arguments, you will find it actually adds _a lot_ of verbosity. I can still try to add it in the proposal though maybe making declaration more explicit is a better idea after all.

I'd argue in favor of it (obviously).  I think its at least worth adding to the design decisions.  I can make a PR if you want.

> I agree in that using one-line `if`s looks a lot like guards, but I do not think it is an option. In particular, it leads to a nested `if` inside the `when`, which is exactly what this proposal aims to avoid.

I agree its not ideal, but I think its better than adding an entirely new keyword just to avoid it.  Also `where` is already used for [multiple upper bounds](https://kotlinlang.org/docs/reference/generics.html#upper-bounds), you probably want to choose a different keyword.

> That is a neat idea, but it does introduce a new keyword. How would it look like for more complex nested patterns? You could make a PR into the proposal under the Guards section, where you discuss this possible alternative :).

I'll make a PR for that.

> Another neat idea, and I did think about this one writing the proposal. I really like the idea of matching by name on classes that may or may not have `componentN()`. At the same time though, I cannot come up with a clear syntax. I think the one you suggested is a bit inappropriate in that it uses `:`, which in Kotlin has always been a 'is of type' relationship. I thought about using the same syntax as named parameter passing (what you did but using `=` instead of `:`) but `name = val x` just looks weird in Kotlin too. Doing it the other way around would mean using the named parameter on the RHS which also looks weird. I also think that it makes an already complex feature more convoluted still! But it does not 'break' the already proposed syntax, so it could be added later down the line. You could add it as an idea under the Beyond the Proposal section.

I'll make a PR for this too.  I agree about the syntax, there's not a lot of good options.

**Comment by Cottand** [comment]
Thanks for your input @fvasco and @rnett .
> I suggest to mark every new value declaration with val, as when does.
> Moreover, we should avoid any distinction between identifier and constant. In is Success(x), the x must always be a valid identified: parameter, attribute, variable or a const val one; this code have to be clear even without read the entire kt file and the imported ones.

I fully agree that there should be no ambiguity when matching, which is why the proposal addresses the problem in a number of ways, like explicit matching and guards. Of the two, explicit matching is the most 'simple' in that it fully tells apart whether we are assigning a new identifier or not (like using `val` everywhere does, except it is the less common case, and thus makes matching _a lot_ shorter). Personally, I think guards also solve the issue and are even more powerful, but they increase the scope of the proposal a lot which is why they are not the default solution.
I think you will agree that getting rid of the verbosity Java introduced is in the spirit of Kotlin, as long as that preserves clarity.

> >An idiom in Haskell or Scala is to pattern match on collections.
>
>Who cares?
>How "Matching on collections" should help a Kotlin developer?

I agree that Kotlin does not need matching on collections, which is why it is clearly stated that _'matching on collecitons is **not** the aim of the proposal'_ in the same section you quoted. It is just an issue that has been discussed before as a possible motivation for pattern matching in the language. But again, this proposal does not address that.

>In your example, mySum* isn't tailrec and can be efficently replaced with a fold.

It is a minimal example to show how it _could_ be done on Kotlin collections. `fold` only covers some of what recursion typically addresses. In particular, you could not write the example `mySum` for a Tree, as the recursion pattern is not the same.

> Finally the references to JEP 375 and "Pattern Matching for Java" are wrong for me. Those cover the Kotlin's smart cast. 

The idiom does a lot more than smart casts. In particular, it allows nested pattern matching and equality checking. I suggest you watch [this bit](https://www.youtube.com/watch?v=te3OU9fxC8U&feature=youtu.be&t=2528) of KotlinConf '19 which I use as one of the examples of the proposal.

> Java does not plan to implement this kind of pattern mathing.

Actually, I am aiming to implement exactly that kind of pattern matching! With some possible additions, like guards or explicit matching. Again, Jake Wharton's example in the proposal is actually taken from a Java 14 vs Kotlin comparison.

Now, to address @rnett .

>I'd argue in favor of it (obviously). I think its at least worth adding to the design decisions. I can make a PR if you want.

Yes, that would be great, thanks. I agree that a new subsection under Design decisions is a good place. Thanks for your other PRs too! You can also add yourself as a contributor.

>I agree its not ideal, but I think its better than adding an entirely new keyword just to avoid it. Also where is already used for multiple upper bounds, you probably want to choose a different keyword.

Fair point. What do you think of still keeping `where` precisely because it is already a reserved keyword? Again, this would avoid introducing (even) more keywords. Given context is completely different, this could be fine? Much like how `@` is used for both annotations and labels.

**Comment by rnett** [comment]
> Fair point. What do you think of still keeping where precisely because it is already a reserved keyword? Again, this would avoid introducing (even) more keywords. Given context is completely different, this could be fine? Much like how @ is used for both annotations and labels.

I'm not sure which is better tbh, they both have drawbacks.  Its probably a question that would have go to the Kotlin devs.

**Comment by hepin1989** [comment]
> > `when (x) if patten`
> > `when (x) match patten`
> > `when (x) case patten`
> > I think anyway, it's not `is`.
> 
> `if` is largely not appropriate since it should be followed by a `Boolean` as it is in the rest of the language. Otherwise that is plain confusing.
> Your other alternatives, again, introduce new keywords, which is a breaking change.

image
or just the rust's style

**Comment by whyoleg** [comment]
I want to just mention one thing about nested matching. In current proposal if we will take a look on [one of examples](https://github.com/Cottand/KEEP/blob/pattern-matching/proposals/pattern-matching.md#from-jake-wharton-at-kotlinconf-19) and adapt it a little:
```kotlin
val result = when(download) {
  is App(name, Person("Alice", "some other")) -> "Alice's app $name"
  is Movie(title, Person("Alice", "some other")) -> "Alice's movie $title"
  is App, Movie -> "Not by Alice"
}
```
Here `Person("Alice", "some other")` can be interpreted differently:
- constructor invocation + comparing by `equals`
- matching by class `Person` + comparing 2 arguments of it

The result of two such cases will be different.

One other thing, that we can't check just for instance of in nested patterns. We can take the same example. Then, f.e., we can't easy check that `App.developer` (second property of `App` class) is just `Person` class, now we can only match it parameters.

So to overcome those to issues, I propose to use `is` also inside nested patterns with `val` for property assignments as it was proposed upper. It's of course become more wordy, but then it will be more explicit (as Kotlin is all about explicit).

Then cases which I worried about will look like this:
- Calling function/constructor:
```kotlin
val result = when(download) {
  is App(name, Person("Alice", "Some other")) -> "Alice's app $name" //checking by .equals
  is App, Movie -> "Not by Alice"
}
```
- Matching properties:
```kotlin
val result = when(download) {
  is App(name, is Person("Alice", "Some other")) -> "Alice's app $name" //checking that class is Person and properties match
  is App, Movie -> "Not by Alice"
}
```
- Matching properties + assignment:
```kotlin
val result = when(download) {
  is App(name, is Person("Alice", val prop)) -> "Alice's app $name" //checking that class is Person and properties match + assign prop
  is App, Movie -> "Not by Alice"
}
```
- Matching just class (not sure about syntax here)
```kotlin
val result = when(download) {
  is App(name, val person is Person) -> "Alice's app $name" //checking that class is Person and assign person
  is App, Movie -> "Not by Alice"
}
```

Cottand:
>Here Person("Alice", "some other") can be interpreted differently:?
>
>constructor invocation + comparing by equals
>matching by class Person + comparing 2 arguments of it

Well spotted. I think adding `is` in every match (including the nested ones) makes for some less natural looking stuff. In particular, your 'matching just class' example (aside from the obvious sacrifice of having to write out `is` every time). Do you think it would be reasonable to restrict matching only to literals and identifiers (so no constructors, no function calls, etc)? It could be a solution since it would also forbid werid matches like this from the language:
```kotlin
val result = when(download) {
  is App(name, DB.getExpectedPerson(someParam)) -> "Alice's app $name" //checking that class is a person we wanted
  is App, Movie -> "Not by Alice"
}
```
Which is confusing, because it looks like we might be destructuring `DB.getExpected...` and even extracting `someParam` from it. Making that illegal would force the programmer to write instead:
```kotlin
val expected = DB.getExpectedPerson(someParam)
val result = when(download) {
  is App(name, person) where person == expected -> "Alice's app $name" //checking that class is a person we wanted
  is App, Movie -> "Not by Alice"
}
```
... where we use the guards solution, or alternatively
```kotlin
val expected = DB.getExpectedPerson(someParam)
val result = when(download) {
  is App(val name, expected) -> "Alice's app $name" //checking that class is a person we wanted
  is App, Movie -> "Not by Alice"
}
```
...where we use the explicit declaration solution.

Kroppeb:
I'm in favour of requiring `val`. Grammar wise I would do it like this:
Each element in the destructuring is either
* `'val' variableDeclaration`
* `whenCondition`: This would allow for nested pattern matching.
This would make the proposal mostly like how @whyoleg suggested. except for the case:
`is App(name, val person is Person)` becoming `is App(name, val person:Person)`

Cottand:
Thanks, @Kroppeb .
Then would a nested match look like this? (do correct me otherwise!)
```kotlin
val result = when(download) {
  is App(val name, val person: Person ("Alice", _)) -> "Alice's app $name"
  is App, Movie -> "Not by Alice"
}
```
A couple of things bug me a bit about this:
- It looks as if `person` is of type `Person("Alice", _)`.  I don't think that's desirable either.
- You are introducing pattern binders (because we can now 'declare' nested matches' names). I argue against this at the end of the proposal, but in this case it seems it's just a side effect of your suggestion. So potentially there could be no problem. But it does kind of replace smart casts since we could just do `download.developer` now that `download` is smart casted to an `App`. This overlap is one if the main reasons people have argued pattern matching in Kotlin is just unncessary (see fvasco 's comment this morning, but it was not the first time it was argued to me). Again in your case it's just a side effect though.

Cottand:
I once again feel like guards would just solve all this and I am tempted to include them in the main proposal (right now they are there just for discussing the possibility of them). Along with a literals-only restriction, they would solve [the issue](https://github.com/Kotlin/KEEP/pull/213#issuecomment-626307049) @whyoleg rightly pointed out and the [matching existing identifiers](https://github.com/rnett/KEEP/blob/pattern-matching-require-val/proposals/pattern-matching.md#-matching-existing-identifiers) amiguity.

fvasco:
The @whyoleg's proposal looks really interesting to me, but we have to consider the current `when` behaviour.

The second match in the example:

```kotlin
val result = when(download) {
  is App(name, is Person) -> "Alice's app $name"
  is App, Movie -> "Not by Alice"
}
```

Looks like `download is App || download == Movie`, so it should be `is App, is Movie`.

> Do you think it would be reasonable to restrict matching only to literals and identifiers (so no constructors, no function calls, etc)?

A context sensitive behaviour can increase confusion in the reader.

@Cottand, maybe it is a trivial concern, but I don't understand how JEP 375 can implement this KEEP in Java.  I suggest you to drop any reference of pattern matching in Java and the ineffective section "Matching on collections".

whyoleg:
Guards are of course good, but TBH I think that they are to wordy for simple cases.
In my opinion, solution with `val` and `is` inside can be better it 2 ways:
- explicit declaration of everything, IMHO it's one of the best 'features' in Kotlin - so you, know, what you do, and it isn't possible to mean one construction in two different code fragments. I mean case like this, where from first look, you can't know, if `name` is new variable or declared before, or even it's from some receiver:
```kotlin
val result = when(download) {
  is App(name, _) -> "Alice's app $name"
  is App, Movie -> "Not by Alice"
}
```
So here it will be much better to use `val` for declare properties, it will also look like just nested `when`

IMO, it can be even added to language in several steps:
1. Support `is Class("to check", "to check 2")` - extending `when` (and even `if`) statements/expressions with support of `is` check and destructuring check. Without support of nested patterns.
It will replace such patterns:
```kotlin
val result = when(val d = download) {
  is App -> if(d.name == "Oleg") println("Cool name") else println("Not Oleg but: ${d.name}")
}
```
with 
```kotlin
val result = when(val d = download) {
  is App("Oleg") -> println("Cool name")
  is App -> println("Not Oleg but: ${d.name}")
}
```
It seems will not break old semantics, but add one more
2. Support variable assignments inside destructuring with `val` keyword. Even now, it's not really a problem, because you can just get property by smartcasting (as in example upper).
So it will replace:
```kotlin
val result = when(val d = download) {
  is App("Oleg") -> println("Cool name")
  is App -> println("Not Oleg but: ${d.name}")
}
```
with this:
```kotlin
val result = when(val d = download) {
  is App("Oleg") -> println("Cool name")
  is App(val name) -> println("Not Oleg but: $name")
}
```
Also, new semantic, not breaking old code
3. Support nested `when` conditions. It will be possible to allow all existing conditions like with ordinary when. It will look like this:
- equality check is already implemented in step 1
- type check
```kotlin
val result = when(download) {
  is App("Oleg", is Person) -> ...
}
```
- range check (let third property of `Person` will be an Int)
```kotlin
val result = when(download) {
  is App("Oleg", _, in 0..10) -> ...
}
```
- type check + destructuring
```kotlin
val result = when(download) {
  is App("Oleg", is Person("name", val cost)) -> ...
}
```
And so on with other nested destructuring. It will use same semantics.
One thing about type check with assignment. What I propose before is:
4. Adding guards as an additional check statements, f.e. checking related properties. Im really more like just `where` statement after condition. It looks like a `post condition` and is easier to read, then `guard` that are in proposal, even if they are more powerfull.

So, examples with guard will be:
```kotlin
val result = when(download) {
  is App("Oleg", is Person("name", val cost)) -> ...
}
```

whyoleg:
> Looks like `download is App || download == Movie`, so it should be `is App, is Movie`.
@fvasco right! it should be double `is`, as in current `when`. I was just copy it from proposal..

Kroppeb:
> Thanks, @Kroppeb .
> Then would a nested match look like this? (do correct me otherwise!)
> 
> ```kotlin
> val result = when(download) {
>   is App(val name, val person: Person ("Alice", _)) -> "Alice's app $name"
>   is App, Movie -> "Not by Alice"
> }
> ```
> 
> A couple of things bug me a bit about this:
> 
> * It looks as if `person` is of type `Person("Alice", _)`.  I don't think that's desirable either.
> * You are introducing pattern binders (because we can now 'declare' nested matches' names). I argue against this at the end of the proposal, but in this case it seems it's just a side effect of your suggestion. So potentially there could be no problem. But it does kind of replace smart casts since we could just do `download.developer` now that `download` is smart casted to an `App`. This overlap is one if the main reasons people have argued pattern matching in Kotlin is just unncessary (see fvasco 's comment this morning, but it was not the first time it was argued to me). Again in your case it's just a side effect though.

Actually, my proposal wouldn't add that. What I wrote would only allow simple type checks. I wasn't gonna add them, but I saw whyoleg suggesting to have a simple type check available. Idk don't think this would be used a lot so it might be better to only allow `'var' simpleIdentifier` indeed.

Cottand:
> The second match in the example:
>  [...]
> Looks like `download is App || download == Movie`, so it should be `is App, is Movie`.

Well spotted, @fvasco . That would introduce breakage with existing code I am afraid.

> > Do you think it would be reasonable to restrict matching only to literals and identifiers (so no constructors, no function calls, etc)?
> 
> A context sensitive behaviour can increase confusion in the reader.

I don't think behaviour is context sensitive. You can only match on literals and identifiers, and that's it. If this rule is enforced (along with guards) it becomes very readable, as you know that whatever is in the match _has_ to be a new variable.


> @Cottand, maybe it is a trivial concern, but I don't understand how JEP 375 can implement this KEEP in Java. I suggest you to drop any reference of pattern matching in Java and the ineffective section "Matching on collections".

JEP 375 and this KEEP are different things. I just used Goetz' proposal as inspiration for the syntax. When pattern matching in Java does come, it will not be compatible with Kotlin classes due to the fact that (as of now) Kotlin does not generate bytecode corresponding to records. This is because that is JVM14 bytecode. At the same time, whether we'll be able to destructure Java records on our end depends on how the Kotlin team decides to define interop with records, independently of this KEEP.

As for collections, it is under the Limitations section, precisely because we are not aiming for that. I can drop the bit on defining an extension function to be able to match, if you want, since that is not part of the proposal anyway.

Addressing @whyoleg ,
Matching nested `when` actually looks pretty good. I'll show that new syntax in comparison with the original proposal for comparison, with the added equality check:
```kotlin
// Original proposal:
val expected = "Alice"
val result = when(download) {
  is App(name, Person(dev, _)) where dev == expected -> "$expected's app $name"
  is Movie(title, Person(dir, _))  where dir == expected-> "$expected's movie $title"
  is App, Movie -> "Not by Alice"
}
```
```kotlin
// Alternative
val expected = "Alice"
val result = when(download) {
  is App(val name, is Person(expected, _)) -> "$expected's app $name"
  is Movie(val title, is Person(expected, _)) -> "$expected's movie $title"
  is App, Movie -> "Not by Alice"
}
```
I am personally ok the new syntax, although it is more verbose, which will be noticeable in bigger matches. I want to ask whether omitting a class name is still valid code? Eg, is this ok, from the Arrow example:
```kotlin
fun <A> Kind<ForOption, A>.eqK(other: Kind<ForOption, A>, EQ: Eq<A>) =
    when(this.fix() to other.fix()) {
        is (is Some(val a), is Some(val b)) -> EQ.run { a.eqv(b) }
        is (is None, is None) -> true
        else -> false
    }
```
Matching on known `Pair`s looks useful when comparing patterns in two objects (notice x6 `is` for 2 matches of 2 objects),

whyoleg:
Also a little thought about [tuple syntax](https://github.com/Cottand/KEEP/blob/pattern-matching/proposals/pattern-matching.md#-destructuring-tuples-syntax).
I think it's not needed at all, as we have already Destructuring Declarations. So f.e.
```kotlin
when(person) {
  is ("Alice", age) -> // ...
}
```
can be replaces with: 
```kotlin
val (alice, age) = person
when(alice) {
  "Alice" -> // ..
}
```
or for consistency
when(person) {
  is Pair("Alice", age) -> // ...
}
```
IMO construction `is (.., ..)` looks not intuitive

**Comment by Cottand** [comment]
Using your alternative would then stop us from some pretty idiomatic tuple checking on other languages though. See the Arrow example I just posted

**Comment by whyoleg** [comment]
@Cottand as for now, kotlin doesn't have a `tuples` in language. So I think, we don't need a shortcut for `tuple` destructuring. Why not to just use `Pair` class?
So your example will look like: 
```kotlin
fun <A> Kind<ForOption, A>.eqK(other: Kind<ForOption, A>, EQ: Eq<A>) =
    when(this.fix() to other.fix()) {
        is Pair(is Some(val a), is Some(val b)) -> EQ.run { a.eqv(b) }
        is Pair(is None, is None) -> true
        else -> false
    }
```
Even, if doesn't check for type, it use property destructuring. IMO, it's better to keep consistency over `shortcut for 4 symbols`.

Cottand:
Kotlin does not have a `tuple` type, but in practice any data class we use is a tuple of any size because we can destructure it.

I'd like to highlight that
```kotlin
fun <A> Kind<ForOption, A>.eqK(other: Kind<ForOption, A>, EQ: Eq<A>) =
    when(this.fix() to other.fix()) {
        is (Some(a), Some(b)) -> EQ.run { a.eqv(b) }
        is (None, None) -> true
        else -> false
    }
```
... which was a pretty short and simple example, has grown into:
```kotlin
fun <A> Kind<ForOption, A>.eqK(other: Kind<ForOption, A>, EQ: Eq<A>) =
    when(this.fix() to other.fix()) {
        is Pair(is Some(val a), is Some(val b)) -> EQ.run { a.eqv(b) }
        is Pair(is None, is None) -> true
        else -> false
    }
```

The `is (_, _)` syntax would only be used with additional matches, and is optional. I do not see harm in letting it be, especially because it is pretty redundant to write a type check for a class that we know to be a `Pair` for every match that we want to perform on it. I'd say it is in the type inference spirit of the language.

whyoleg:
@Cottand Yes, example is grown, but now you can interpret it in one way. You see, that you double check for instance, and create new variables.
Before it was:
- do it call function, or it type check
- do it create variable, or it check for existing variable which comes from upper scope, or receiver
I will mention it one more time: in such `core` feature, explicit declaration of everything will be much better, that implicit, when you can understand in different ways

Cottand:
@whyoleg I see that the issue before was that the semantics could be ambiguous. I will update the proposal later today so that is not the case anymore, and add the idea of recursive `when` conditions.

I definitely agree that such a core feature should not be ambiguous. At the same time (and preciesely because it's a core feature) it's something that people might be writing all the time! It's hard to find that compromise.

Cottand:
Consequently, I will make this PR into a draft until I have properly updated the proposal and then we can resume discussion :) thanks for your inputs people!

Kroppeb:
About the nested pattern matching. Are `where` clauses only allowed on outside, or would something like this be legal too?
```kotlin
val result = when(entity) {
  is Player(
    is Person(val name, val age) where age >= 18,
    val nickName
  ) -> ...
}
```
or should we only allow
```kotlin
val result = when(entity) {
  is Player(
    is Person(val name, val age),
    val nickName
  )  where age >= 18 -> ...
}
```

Cottand:
To answer @Kroppeb , I personally think nesting guards would easily complicate `when`s very easily but at the same time they provide a very expressive way of supplying samll conditions if combined with lambdas. @rnett suggested something similar which you will be able to find in the 'Beyond' section of the proposal.

LifeIsStrange:
Just saying that the proposal should:

1) Mention and compare its choices with the Java ones in https://cr.openjdk.java.net/~briangoetz/amber/pattern-match.html
The design should be at least as clear and expressive as the Java one 

2) show that deconstruction can be used in if () too (a consequence of 1))

3) hint that the implementation will consider making use of this performance optimization when available (not too old jvm) -> quoting the JEP
*The implementation will likely make use of Dynamic Constants in the JVM.*
https://bugs.openjdk.java.net/browse/JDK-8177279
https://openjdk.java.net/jeps/8213076
This could allow to reduce/eliminate the potential opt-in cost of compile time
(again a consequence of 1))

But we can probably do even better than Java in the following ways ->

4) be engeenered to be forward-compatible with future enhancements like more powerful patterns such as rust or patterns (for example) https://github.com/rust-lang/rfcs/blob/master/text/2535-or-patterns.md

5) be as well designed as other kotlin features / syntaxes 
Kotlin is one of the very few languages to have simultaneously great semantics, and excellent syntaxes. This is made possible by being able to study in depth the prior art, the past design of the other programming languages in order to create the best mix of ideas from others while avoiding their historical errors/deficiencies. 
This is why the kotlin architects have a *duty* of studying prior art in order to design one of the best pattern matching syntax/semantics. 

Hence some references to rigorously study ->
See above the Java design 
See here the current design of Scala pattern matching 
https://docs.scala-lang.org/tour/pattern-matching.html
See here the future of Scala pattern matching 
https://docs.scala-lang.org/sips/case-if.html
https://dotty.epfl.ch/docs/reference/changed-features/pattern-matching.html

The Rust design 
https://doc.rust-lang.org/book/ch18-00-patterns.html
https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html

Cottand:
Thanks for yout input @LifeIsStrange . 

As a side note, I would greatly appreciate it if people wishing to participate in this discussion would read the proposal before commenting. That way you can provide comments that relate what you expect to what is currently planned, instead of only commenting on what you expect. The motivations behind all the decisions are also discussed, and for most of them alternatives are offered. Just a note for a couple people so far :) thank you!

> Just saying that the proposal should:
> 
> 1. Mention and compare its choices with the Java ones in https://cr.openjdk.java.net/~briangoetz/amber/pattern-match.html
>    The design should be at least as clear and expressive as the Java one

Brian Goetz's post on the pattern matching coming to Java has actually heavily influenced this KEEP. The form of pattern matching we propose is _more_ expressive than Java's, and (in my personal opinion) also clearer. This is due to the fact that we match on nested arguments just the same, without having to declare their types (or `var`). The syntax is otherwise very similar and the semantics are slightly different in that in Kotlin you would be able to pattern match only against literals (to test for equality) to avoid weird inlined function calls that would be harder to read. We cover that use case through [Guards](https://github.com/Cottand/KEEP/blob/pattern-matching/proposals/pattern-matching.md#guards-) which also offer extra expressiveness through predicates.

> 2. show that deconstruction can be used in if () too (a consequence of 1))

Java supports this but that does not mean this KEEP should, as the design space is very different. As the author of the KEEP I see no need for this (use case is mostly covered through smart casts and a guard would not be of any use) and it has only been suggested by 1 person so far. While I certainly think we should consider it if the community judges it is something that would bring a lot of value, I do not think we should include it in the proposal just because Java is doing it.

> 3. hint that the implementation will consider making use of this performance optimization when available (not too old jvm) -> quoting the JEP
>    _The implementation will likely make use of Dynamic Constants in the JVM._
>    https://bugs.openjdk.java.net/browse/JDK-8177279
>    https://openjdk.java.net/jeps/8213076
>    This could allow to reduce/eliminate the potential opt-in cost on compile time
>    (again a consequence of 1))
> 

This optimisation uses dynamic class file constants, which is something that has to be engineered on the class being matched rather that on the logic doing the pattern matching. It basically means having a whole new constant pool in JVM classes, afaik and Kotlin does not currently do this (as it is compatible with Java 6). When (and if) Kotlin classes start doing this (data classes _might_ at some point) this KEEP can be adjusted accordingly. Until then, this implementation mostly focuses on using existing semantics (like `componentN` functions) to exhibit expected/predictable behaviour.

tl;dr, using a new dynamic constant pool in the JVM is up to the Kotlin team.

> But we can probably do even better than Java in the following ways ->
> 
> 1. be engeenered to be forward-compatible with future enhancements like more powerful patterns such as rust or patterns (for example) https://github.com/rust-lang/rfcs/blob/master/text/2535-or-patterns.md

Please correct me if I am mistaken, because I have only played around with rust a bit. but *rust does not offer guards*. It offers membership checking and disjunctions (at least in the RFC you linked) which are just Boolean checks that can be offered through a guard. Additioanally, a guard offers any check the user can write, but also functions! This enables libraries and users to offer their own matches. I think this makes for very powerful matches.

As for forward compatibily that largely depends on the scope of future improvements.

> 2. be as well designed as other kotlin features / syntaxes
>    Kotlin is one of the very few languages to have simultaneously great semantics, and excellent syntaxes. This is made possible by being able to study in depth the prior art, the past design of the other programming languages in order to create the best mix of ideas from others while avoiding their historical errors/deficiencies.
>    This is why the kotlin architects have a _duty_ of studying prior art in order to design one of the best pattern matching syntax/semantics.

Frankly, we are doing our best to make this a well-designed feature! But that is very personal so it requires inspiration from existing semantics, and careful study of the alternatives. Again, if you read this proposal you will find it discusses features available in other languages. Some it decides to bring to Kotlin (like guards) that are not available in other languages (like Java); but also some that are available in other languages (like as-patterns or binders) that it is judged are already covered by the existing semantics and thus would only clog down the proposal.

But at the end of they day, it is up to the Kotlin team (experts at designing a language!) to decide what is well-designed, and as such, this proposal is only a careful suggestion of what could be done and is open to changes.

> 
> Hence some references to rigorously study ->
> See above the Java design
> See here the current design of Scala pattern matching
> https://docs.scala-lang.org/tour/pattern-matching.html
> See here the future of Scala pattern matching
> https://docs.scala-lang.org/sips/case-if.html
> https://dotty.epfl.ch/docs/reference/changed-features/pattern-matching.html
> 
> The Rust design
> https://doc.rust-lang.org/book/ch18-00-patterns.html
> https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html
> 
> @elizarov

All the references you quoted were studied in the making of this proposal, and some are even quoted as its references. If you are interested in our design decisions as opposed to some of scala's, I can elaborate further on that!

I hope this addresses all your concerns

edrd-f:
Some considerations:

* The comparison of code from Baldung's "Binary Trees" is not fair, since it's comparing `if / else` code with the pattern matching equivalent. It should compare the current `when` implementation with the proposed pattern matching implementation:
    ```kotlin
    private fun removeNoChildNode(node: Node, parent: Node?) {
      when {
        parent == null ->
          throw IllegalStateException("Can not remove the root node without child nodes")
        node == parent.left -> parent.left = null
        node == parent.right -> parent.right = null
      }
    }
    ```

    versus

    ```kotlin
    private fun removeNoChildNode(node: Node, parent: Node?) {
      when (node to parent) {
        is (_, null) ->
          throw IllegalStateException("Can not remove the root node without child nodes")
        is (n, Node(n, _)) -> parent.left = null
        is (n, Node(_, n)) -> parent.right = null
      }
    }
    ```
* Quoting the argument against `val` syntax for value extraction:
    > It is extremely cumbersome to write for nested patterns of more than one level, or where there is a lot of inspection of destructured elements. Consider how using is and val every time would look for the Arrow example

    "Extremely cumbersome" is subjective. I suggest saying it "implies writing additional code" (which is objective) and putting the equivalent version of the Arrow example. This will help people make their minds based on the code.

* Except for the Arrow and Jake Warton's examples, other comparisons show very few differences in the amount of code necessary to implement such cases. Considering Jake Warton's example is an ad-hoc code made up for the presentation, I wouldn't take it as an example of code that could be improved by pattern matching because there are better ways of writing it, such as with the usage of polymorphism. 
My suggestion is to include more real-world examples that show how the code can be improved by the proposal. Not because I don't think pattern matching is useful, but because such examples are the best argument, in my opinion, to why a feature should or should not be included in the language.

Cottand:
Hi @edrd-f 

> * The comparison of code from Baldung's "Binary Trees" is not fair, since it's comparing `if / else` code with the pattern matching equivalent. It should compare the current `when` implementation with the proposed pattern matching implementation:

While I see your point, I just aimed to compare code found in the wild. Chained `if`s and `when`s a are interchangeable in most situations (and IntelliJ even offers a quick action to swap between them). Baeldung chose `if`s so that is what is used. If the snippet was modified before comparing it, then it would hardly be a comparison betweem code from a tutorial and the alternative this proposal offers.

>   "Extremely cumbersome" is subjective. I suggest saying it "implies writing additional code" (which is objective) and putting the equivalent version of the Arrow example. This will help people make their minds based on the code.

I agree. I changed the wording and added the example.

> * Except for the Arrow and Jake Warton's examples, other comparisons show very few differences in the amount of code necessary to implement such cases. Considering Jake Warton's example is an ad-hoc code made up for the presentation, I wouldn't take it as an example of code that could be improved by pattern matching because there are better ways of writing it, such as with the usage of polymorphism.
>   My suggestion is to include more real-world examples that show how the code can be improved by the proposal. Not because I don't think pattern matching is useful, but because such examples are the best argument, in my opinion, to why a feature should or should not be included in the language.

Finding "Real world" examples as in more examples from existing open source codebases is not that easy, since I need to go through repos finding places where just a few lines could be replaced with a pattern match. If functions are used for nested checking, for example, it cannot be an exmple for the proposal (even though it would be a good example for pattern matching!).
I will still try to come up with nice examples of my own for comparison.

As for how some of the examples show very few differences in the amount of code written, I would like to add that pattern matching also offers readability (and not just briefness). I think the [couroutines](https://github.com/Cottand/KEEP/blob/pattern-matching/proposals/pattern-matching.md#-from-kotlinxcoroutines) and [Arrow](https://github.com/Cottand/KEEP/blob/pattern-matching/proposals/pattern-matching.md#from-the-arrow-library-) examples show this nicely.

fvasco:
I have to share the point of view of @edrd-f,
this KEEP might contain a good idea, but I think the comparison is not fair.

Your labelled your example "Without pattern matching", but it should be "code found in the wild".
You cherry picked some code snipped, but there is no evidence that this proposal can help Baeldung.
A developer write some code for specific reasons, this proposal may not cover these use case.

> JEP 375 and this KEEP are different things. I just used Goetz' proposal as inspiration for the syntax. 

I cannot see any evidence. JEP 375 looks really extraneous.

Cottand:
@fvasco , for every single example shown, there is a link to where the original code 'without pattern matching' was found. This is true for Baeldung's but also for the other open source libraries. If you think the example is not fair because Baeldung chose to use nested `if`s, or that Baeldung would not benefit from this proposal, I can always replace that example with one of my own.

> I cannot see any evidence. JEP 375 looks really extraneous.

Please see this [more detailed explanation](https://cr.openjdk.java.net/~briangoetz/amber/pattern-match.html), which is what was used as inspiration. Aside from referencing it as a source, I cannot really provide any further evidence.

Additionally, I found [a couple more snippets](https://github.com/Cottand/KEEP/blob/pattern-matching/proposals/pattern-matching.md#from-tornadofx) where pattern matching could be of use (7108c931b80161706b70bf20c3740cc7b25c9cd2) in order to provide even more useful examples.

fvasco:
Hi @Cottand,

> If you think the example is not fair because Baeldung chose to use nested ifs, or that Baeldung would not benefit from this proposal, I can always replace that example with one of my own.

I will use the examples of real code exposed in this KEEP, assuming those are the "best reason in favour of pattern matching".

I am joking in this reply, however please consider my statements.

### From the Arrow library: 

With pattern matching:

```
fun <A> Kind<ForOption, A>.eqK(other: Kind<ForOption, A>, EQ: Eq<A>) =
    when(this.fix() to other.fix()) {
        is (Some(a), Some(b)) -> EQ.run { a.eqv(b) }
        is (None, None) -> true
        else -> false
    }


fun Ior<L, R>.eqv(other: Ior<L, R>): Boolean = when (this to other) {
    is (Ior.Left(a), Ior.Left(b)) -> EQL().run { a.eqv(b) }
    is (Ior.Both(al, ar), Ior.Both(bl, br)) -> EQL().run { al.eqv(bl) } && EQR().run { ar.eqv(br) }
    is (Ior.Right(a), Ior.Right(b)) -> EQR().run { a.eqv(b) }
    else -> false
}
```

Without pattern matching:

```
fun <A> Kind<ForOption, A>.eqK(other: Kind<ForOption, A>, EQ: Eq<A>) =
    (this.fix() to other.fix()).let { (a, b) ->
      if (a is Some && b is Some) EQ.run { a.t.eqv(b.t) }
      else a is None && b is None
    }

fun Ior<L, R>.eqv(b: Ior<L, R>): Boolean = when (this) {
    is Ior.Left -> b is Ior.Left && EQL().run { value.eqv(b.value) }
    is Ior.Both -> b is Ior.Both && EQL().run { leftValue.eqv(b.leftValue) } && EQR().run { rightValue.eqv(b.rightValue) }
    is Ior.Right -> b is Ior.Right && EQR().run { value.eqv(b.value) }
  }
}
```

In such case, removing the pattern matching avoids the variable proliferation, reduces the cyclomatic complexity and the total amount of lines of code, increasing the readability.

### From kotlinx.coroutines:

With pattern matching:

```
public val value: E get() {
    _state.loop { state ->
        when (state) {
            is Closed(valueException) -> throw valueException
            is State<*>(value) where { value == UNDEFINED } -> throw IllegalStateException("No value")
            is State<*>(value) -> return value as E
            else -> error("Invalid state $state")
        }
    }
}
```

Without pattern matching:

```
public val value: E get() {
    _state.loop { state ->
        when (state) {
            is Closed -> throw state.valueException
            is State<*> -> {
                check(state.value !== UNDEFINED) {"No value"}
                return state.value as E
            }
            else -> error("Invalid state $state")
        }
    }
}
```

In such case, pattern matching increases the number of type check in a performance critical library, without provide any reasonable enhancements.

### From JetBrains' Exposed:

With pattern matching:

```
infix fun Expression<Boolean>.and(op: Expression<Boolean>): Op<Boolean> = when(this to op) {
  is (AndOp, AndOp(opExpres)) -> AndOp(expressions + opExpres)
  is (AndOp, _) -> AndOp(expressions + op)
  is (_, AndOp(opExpres)) -> AndOp(ArrayList<Expression<Boolean>>(opExpres + 1).also {
        it.add(this)
        it.addAll(opExpres)
    })
  else -> AndOp(listOf(this, op))
}
```

Without pattern matching:

```
infix fun Expression<Boolean>.and(op: Expression<Boolean>): Op<Boolean> = when {
    this is AndOp && op is AndOp -> AndOp(expressions + op.expressions)
    this is AndOp -> AndOp(expressions + op)
    op is AndOp -> AndOp(ArrayList<Expression<Boolean>>(op.expressions.size + 1).also {
        it.add(this)
        it.addAll(op.expressions)
    })
    else -> AndOp(listOf(this, op))
}
```

In such case, pattern matching requires a `Pair` extra allocation, puts new variable on the table without reduce the amount of code.

### From TornadoFX:

With pattern matching:

```
return when (val json = Json.createReader(StringReader(content)).use { it.read() }) {
    is JsonArray where { it.isEmpty() } -> Json.createObjectBuilder().build()
    is JsonArray -> json.getJsonObject(0)
    is JsonObject -> json
    else -> throw IllegalArgumentException("Unknown json result value")
```

Without pattern matching:

```
return when (val json = Json.createReader(StringReader(content)).use { it.read() }) {
    is JsonArray && json.isEmpty()-> Json.createObjectBuilder().build()
    is JsonArray -> json.getJsonObject(0)
    is JsonObject -> json
    else -> throw IllegalArgumentException("Unknown json result value")
```

Should this KEEP help TornadoFX? How?

### From dokka:

With pattern matching:

```
fun processSampleBody(psiElement: PsiElement): String = when (psiElement) {
    is KtDeclarationWithBody(KtBlockExpression(text)) -> text.removeSurrounding("{", "}")
    is KtDeclarationWithBody(body) -> body!!.text
    else -> psiElement.text
}
```

Without pattern matching:

```
fun processSampleBody(psiElement: PsiElement): String {
    if (psiElement !is KtDeclarationWithBody) return psiElement.text
    val bodyExpression = psiElement.bodyExpression as? KtBlockExpression ?: return psiElement.bodyExpression!!.text
    return psiElement.bodyExpression.text.removeSurrounding("{", "}")
}
```

In such case there this KEEP increases the verbosity.

JakeWharton:
For what it's worth, my example isn't really contrived, it's was just simplified and moved into a domain that people would be familiar with (rather than having to explain both the domain and the code). I don't really understand the polymorphism hand-waving-solution anyway.

Cottand:
> I will use the examples of real code exposed in this KEEP, assuming those are the "best reason in favour of pattern matching".

I would not make this assumption. I could have provided many different code snippets that would have shown more obvious benefits, but chose not to in favour of real world snippets from codebases I knew.

The most illustrative example is [Jake Wharton's](https://github.com/Cottand/KEEP/blob/pattern-matching/proposals/pattern-matching.md#from-jake-wharton-at-kotlinconf-19). I specifically chose to not only display 'clones' of it. 

> In such case, removing the pattern matching avoids the variable proliferation, reduces the cyclomatic complexity and the total amount of lines of code, increasing the readability.

Whether it increases readability is arguable, as it is harder to read both cases in parallel (and you only shave off 1 line!).

> In such case, pattern matching increases the number of type check in a performance critical library, without provide any reasonable enhancements.

The number of type checks only increases with a naive implementaiton that does not 'cache' results of previous succesful `instanceof` calls. 

The performance impact of this KEEP (or any other for that matter) can only be evaluated once an implementation that allows us to reason about its complexity and benchmark it exists.

Additionally, if the feature does turn out to have a performance impact, a user can always opt out by not writing a pattern match. This trade-off already exists when choosing to use an API through DSLs for example. They require a number of extra allocations that are a price to pay for readabilty and perhaps type safety. A good example is present in the Kotlin documentation on [type safe builders](https://kotlinlang.org/docs/reference/type-safe-builders.html#a-type-safe-builder-example), even though the most efficient way of writing HTML is unarguably by writing a single string literal.

> ##### From TornadoFX:
> Without pattern matching:
> 
> ```
> return when (val json = Json.createReader(StringReader(content)).use { it.read() }) {
>     is JsonArray && json.isEmpty()-> Json.createObjectBuilder().build()
>     is JsonArray -> json.getJsonObject(0)
>     is JsonObject -> json
>     else -> throw IllegalArgumentException("Unknown json result value")
> ```

Please correct me if I am wrong but this is not even valid Kotlin, as you did a type check followed by `&& ...`. Essentially, you tried to use a guard!

In order to one-line the type check along with `.isEmpty()` (and thus avoiding the nested `if`) you would need to use `when` without a receiver and declare a new variable `json` outside of the when.


> Without pattern matching:
> ```
> fun processSampleBody(psiElement: PsiElement): String {
>     if (psiElement !is KtDeclarationWithBody) return psiElement.text
>     val bodyExpression = psiElement.bodyExpression as? KtBlockExpression ?: return psiElement.bodyExpression!!.text
>     return psiElement.bodyExpression.text.removeSurrounding("{", "}")
> }
> ```
> 
> In such case there this KEEP increases the verbosity.

... but is arguably more readable.

For all the examples you have listed, those libraries chose to use their nested constructs over your alternative. Probably because they were simply more intuitive (especially in the case of dokka!) and this proposal just suggests more work in that direction with through a more powerful construct already tried and tested in other languages.

Moreover, any example of pattern matching is definitely achievable with chained/nested `if`s and `when`s. This proposal simply argues abstracting that away.

Over the next couple days I can add more examples if you are still not convinced!

fvasco:
Thank you, @Cottand,
I fixed my joke.

> Over the next couple days I can add more examples if you are still not convinced!

I don't have to convince me, but we have to provide a valuable proposal, together.

edrd-f:
Hey, @Cottand 

> While I see your point, I just aimed to compare code found in the wild. Chained ifs and whens a are interchangeable in most situations (and IntelliJ even offers a quick action to swap between them). Baeldung chose ifs so that is what is used. If the snippet was modified before comparing it, then it would hardly be a comparison betweem code from a tutorial and the alternative this proposal offers.

Even if `when`s and `if`s are interchangeable, I think it's still misleading because it's not code written in a "Kotlin way", so it shouldn't be used as an example of what pattern matching can improve. An improvement for this case would be to simply refactor the code with `when`. 
I understand your reasoning for not changing the original code, but in this case I would then remove the code or at least add an observation.  

> The most illustrative example is Jake Wharton's. I specifically chose to not only display 'clones' of it.

I'd really like to see the real version of this code. @JakeWharton since your example is considered the most illustrative, could you explain the domain and also post code closer to the real version? 

Also, what I've said about polymorphism is not hand-waving. It's a very old and known technique (https://refactoring.com/catalog/replaceConditionalWithPolymorphism.html). 
Apart from the cyclomatic complexity, knowledge about `App`, `Movie`, `Person` and `Developer` at the same time also suggests a violation of the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) principle (`download.developer.name` -> too much knowledge), so even if this code could be less verbose with pattern matching, it would still be referencing elements that shouldn't have knowledge of each other, making code harder to reason about.

In short, what I want to say is that when we want to improve our code, language features are not the only option. Design patterns have been helping people solve complex problems in very limited languages for decades, and that's why I'm asking for real-world examples here: we should analyze possibilities within language constructs *and* code design as a whole, and for that we need full context.

> As for how some of the examples show very few differences in the amount of code written, I would like to add that pattern matching also offers readability (and not just briefness). I think the couroutines and Arrow examples show this nicely.

I think in some cases it helps readability and in others it makes code too *dense*. Arrow's `eqv` example for me is one that is hard to understand in its current version and even harder with the pattern matching syntax. Maybe it's easy for people familiar with FP, but I would say the best readability improvement here would be to simply extract parts of the code to functions with descriptive names. 

Lastly, I want to thank you for including the Dokka and TornadoFX examples. I tried searching for usages of `when` on GitHub and know that's a hard task.

JakeWharton:
Yeah sorry I don't see how that applies in any way to my example.

edrd-f:
@JakeWharton alright, here's the refactored version of your example using polymorphism:

```kotlin
sealed class Download
data class App(val name: String, val developer: Person) : Download()
data class Movie(val title: String, val director: Person) : Download()
val download: Download = // ...

interface Titleable {
    fun title(): String
}

class AppView(private val app: App) : Titleable {
    override fun title() = when {
        // Prevents Law of Demeter violation 
        app.isDevelopedBy("Alice") -> "Alice's app ${app.name}"
        else -> "Not by Alice"
    }
}

class MovieView(private val movie: Movie) : Titleable {
    override fun title() = when {
        // Prevents Law of Demeter violation 
        movie.isDirectedBy("Alice") -> "Alice's movie ${movie.title}"
        else -> "Not by Alice"
    }
} 

val view: Titleable = when (download) {
    is App -> AppView(download)
    is Movie -> MovieView(download)
}

val result = view.title()
```

It eliminates the need for nested `when`s and destructuring.

edrd-f:
Actually, the link I posted has nothing besides code. This one is better:
https://refactoring.guru/replace-conditional-with-polymorphism

JakeWharton:
I see. This is, of course, at the exorbitant expense of an explosion of code authoring and long-term maintenance. While there may be value in such abstraction for some parts of a codebase, it's somewhat ridiculous to think that you would replace the otherwise simple and (most importantly) localized pattern match with such a thing in the general case. You're basically doing to work of a compiler pattern match yourself.

The real domain is a payment modeling.

edrd-f:
If you look at this short piece of code as the entire app, for sure it's not worth it. Now for long-term maintenance, I think the opposite: it's much better to have classes with well-defined responsibilities (SRP) than a `when` statement that deals with numerous unrelated entities in the system and mixes domain and view logic.

Cottand:
Thanks @edrd-f 
>[...] An improvement for this case would be to simply refactor the code with `when`.
> I understand your reasoning for not changing the original code, but in this case I would then remove the code or at least add an observation.

Fair enough. I will write some Kotlin-FP-style tree snippet or try to find one in the wild.

> Apart from the cyclomatic complexity, knowledge about `App`, `Movie`, `Person` and `Developer` at the same time also suggests a violation of the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) principle (`download.developer.name` -> too much knowledge) [...]

I see why you would think that pattern matching would tend to violate the law of Demeter. It is a good point and I think this comes from FP vs OOP in a broader context, where FP navigates from a component in a structure to another (sometimes even using deep recursion). Instead, OOP 'delegates' tasks to close neighboors (through 'tell, don't ask') embracing mutability. Thus, I think the LoD (in this context) is mainly there to prevent corruption of state, something an FP programmer reasons about in very different ways. I just found [this article](https://enterprisecraftsmanship.com/posts/law-of-demeter-and-immutability/) thought it was an interesting read. At any rate, should it not always be the user's choice when it comes to whether to abide by a programming paradigm's principles?

> I think in some cases it helps readability and in others it makes code too _dense_. Arrow's `eqv` example for me is one that is hard to understand in its current version and even harder with the pattern matching syntax. Maybe it's easy for people familiar with FP, but I would say the best readability improvement here would be to simply extract parts of the code to functions with descriptive names.

I personally would find extracting a simple pattern match into separate functions very hard to read as we would have to make jumps, and overall it would result in so much more code. I agree that not being familiar with the construct, it can look like a lot of things are happening at once. 
 
> Lastly, I want to thank you for including the Dokka and TornadoFX examples. I tried searching for usages of `when` on GitHub and know that's a hard task.

You're welcome! :) while using real world examples makes it harder to make vanilla Kotlin look 'worse', I think they show nice use cases.

proposals/pattern-matching.md
```diff
@@ -0,0 +1,714 @@
...
+- encourages guards
+
+...using `val` and `is` would make for a pattern mathcing that
+  - is more verbose (but offers some clarity)
+  - allows inlining for equality checking
+  - allows recursive `when` conditions (like `in 0..18`)
+
+### Restricting matching to data classes only
+
+A possibilty suggested during the conception of this proposal was to restrict pattern matching to data classes. The argument would be to start with a samller size of 'matchable' elements in order to keep the inital proposal and feature as simple as possible, as it could be extended later down the line.
```
cortinico, line 527:
```suggestion
A possibility suggested during the conception of this proposal was to restrict pattern matching to data classes. The argument would be to start with a smaller size of 'matchable' elements in order to keep the initial proposal and feature as simple as possible, as it could be extended later down the line.
```

proposals/pattern-matching.md
```diff
@@ -0,0 +1,714 @@
...
+
+#### Conclusion
+
+There are two competing alternative syntax possibilites for pattern matching
+in Kotlin. While this proposal argues for one that
+- is slightly more restrictive (no nested `when` conditions)
+- is less verbose
+- encourages guards
+
+...using `val` and `is` would make for a pattern mathcing that
```
cortinico, line 520:
```suggestion
...using `val` and `is` would make for a pattern matching that
```

proposals/pattern-matching.md
```diff
@@ -0,0 +1,714 @@
...
+    Mail.send(addr, "Thanks for choosing us, $name!")
+
+  is Prospect(addr, true) ->
+    Mail.send(addr, "Please consider our product...")
+}
+```
+
+The syntax proposed uses the already existing `is` construct to check the
+type of the subject, but adds what semantically looks a lot like a
+destructuring delcaration with added equality checks. This approach is
```
cortinico, line 53:
```suggestion
destructuring declaration with added equality checks. This approach is
```

proposals/pattern-matching.md
```diff
@@ -0,0 +1,714 @@
...
+on the matched pattern 'changing' depending on the state of the collection.
+Because this proposal aims to use `componentN()` for destructuring, such a
+thing would not be possible in Kotlin as `componentN()` returns the Nth
+element of the collection (instead of its tail for some `componentN()`).
+
+This limitation is due to the fact that in Haskell, a list is represented
+more similarly to how sealed classes work in Kotlin (and we can match on
+those).
+
+**Pattern mathcing on collections is not the aim of this proposal**, but such
```
cortinico, line 545:
```suggestion
**Pattern matching on collections is not the aim of this proposal**, but such
```

Kroppeb:
I want to bring up an issue. Given the following code
```kotlin
sealed class Test<T>{
	data class FailState<T>(val errorCode:Int):Test<T>()
	data class MutableState<T>(var state:T):Test<T>()
}

fun getTest(): Test<Test<Int>> = TODO()

// without matching
fun getItem() = when(val test = getTest()){
	is Test.FailState -> test.errorCode
	is Test.MutableState -> when(val state = test.state){
		is Test.FailState -> state.errorCode
		is Test.MutableState -> state.state
	}
}

// with matching?
fun getItem() = when(getTest()){
	is Test.FailState(val errorCode) -> errorCode
	is Test.MutableState(is Test.FailState(val errorCode)) -> errorCode
	// however the state is mutable, so if the state now switches from Mutable to Fail,
	// none of these branches would trigger
	is Test.MutableState(is Test.MutableState(val result)) -> result
}
```
Should this just not be allowed or should each deconstructor only be called once and cached?

Cottand:
Good remark. I think the expected behaviour would be to only call each deconstructor once. The main case for this is avoiding an extra function call every time and benefitting from smart casts as usual.

I don't see a scenario where the alternative would be more desirable.

elizarov:
Here's a really short personal review of this proposal (disclaimer: it does not necessarily represent the opinion of the Kotlin team on the pattern matching):

👍 Being able to combine type test and guard expression in `when` expression (e.g. `is JsonArray where { it.isEmpty() }` is great. I'm not a fan of this specific syntax (and it is breaking, since `where` is not a hard keyword), but it is a good start anyway. It is a simple, helpful, and nice feature that can be moved to a separate proposal.

👉 Use of positional destructuring needs a careful review from a tooling perspective. `is Customer(name, age, addr)` looks clear, but we already have complains about the existing positional destructuring in Kotlin that it can cause subtle problems when refactoring the order of parameters. These concerns have to be addressed before doubling down on position destructuring and making it even more widely spread in the language.

👎 Introducing a new `age` declaration by `Person(age)` when `age` is "an undefined identifier" sets a dangerous precedent. Nowhere in Kotlin declaration/usage depends on such non-local information as whether a name "is already defined is scope" (whatever it means). Scopes in Kotlin can contain tons of available declarations coming from a plethora of different files. While it is definitely easy for the compiler to track, it would be impossible for a human reading the code to figure out what this piece of code _does_ (check the person has a previously computed age or destruct a person's age into a newly introduced name).

Kroppeb:
> 👍 Being able to combine type test and guard expression in `when` expression (e.g. `is JsonArray where { it.isEmpty() }` is great. I'm not a fan of this specific syntax (and it is breaking, since `where` is not a hard keyword), but it is a good start anyway. It is a simple, helpful, and nice feature that can be moved to a separate proposal.

I would suggest to go with `if` instead of `where` and also allow it to be used in a case such as this
```kotlin
when(val item = getItem()){
	is Book -> ...
	is Bottle -> ...
	if(item.canBreak()) -> ...
	// or potentially
	if item.canBreak ->
	else -> ...
}
```
Now, if you have a long list of `is Type` in your when, and realise you need a slightly different boolean expression, we need to remove our subject and test individually. This syntax would allow adding arbitrary boolean expressions to a `when` with subject

> 👎 Introducing a new `age` declaration by `Person(age)` when `age` is "an undefined identifier" sets a dangerous precedent. Nowhere in Kotlin declaration/usage depends on such non-local information as whether a name "is already defined is scope" (whatever it means). Scopes in Kotlin can contain tons of available declarations coming from a plethora of different files. While it is definitely easy for the compiler to track, it would be impossible for a human reading the code to figure out what this piece of code _does_ (check the person has a previously computed age or destruct a person's age into a newly introduced name).

Yeah, this would be weird behaviour. I thought we were planning to use `Person(val age)` to extract a value, but idk whether that hasn't been updated in the proposal yet, or if this needed more discussion.

Cottand:
Thanks for your inputs, @elizarov , @Kroppeb !

>  Being able to combine type test and guard expression in `when` expression (e.g. `is JsonArray where { it.isEmpty() }` is great. I'm not a fan of this specific syntax (and it is breaking, since `where` is not a hard keyword), but it is a good start anyway. It is a simple, helpful, and nice feature that can be moved to a separate proposal.

I am not married to this syntax, and would also be completely fine with moving it into a new proposal. I do believe though that both guards and pattern matching in general should be designed closely in consideration of each other so they are nice to use if both ever make it into the language.

>  Use of positional destructuring needs a careful review from a tooling perspective. `is Customer(name, age, addr)` looks clear, but we already have complains about the existing positional destructuring in Kotlin that it can cause subtle problems when refactoring the order of parameters. These concerns have to be addressed before doubling down on position destructuring and making it even more widely spread in the language.

I agree, and it was also discussed in Slack that Kotlin could potentially suffer from matches like `is (_, _, _, _, age)` for large data classes (some see this as a problem in Scala).

Potentially, this could be solved by matching on named properties. The proposal [quickly takes a look at the idea](https://github.com/Cottand/KEEP/blob/pattern-matching/proposals/pattern-matching.md#named-components-) but could not come up with nice syntax. Suggestions are welcome!

>  Introducing a new `age` declaration by `Person(age)` when `age` is "an undefined identifier" sets a dangerous precedent. Nowhere in Kotlin declaration/usage depends on such non-local information as whether a name "is already defined is scope" (whatever it means). Scopes in Kotlin can contain tons of available declarations coming from a plethora of different files. While it is definitely easy for the compiler to track, it would be impossible for a human reading the code to figure out what this piece of code _does_ (check the person has a previously computed age or destruct a person's age into a newly introduced name).

> Yeah, this would be weird behaviour. I thought we were planning to use `Person(val age)` to extract a value, but idk whether that hasn't been updated in the proposal yet, or if this needed more discussion.


This concern was raised early when this proposal was submitted, and I think 2 solutions exist: explicitly declaring new variables through `val` (like @Kroppeb suggests) or just not allow matching on defined identifiers (as in that file would not compile at all) thus only matching on literals (this is what Haskell does).

This is discussed in the Design Decisions section (in particular [here](https://github.com/Cottand/KEEP/blob/pattern-matching/proposals/pattern-matching.md#conclusion)). I am personally against having to type `val` every time in favour of equality checks on identifiers with guards only, but again, I am not married to the syntax and think it does need further discussion.

In examples:
```kotlin
val age = 2
when (person) {
is Person(_, age) -> ...
...
```
would **not** be valid, much like one cannot write:

```kotlin
val age = 3
...
val age = 2.3
```
I would like to note that this is at present valid Kotlin:
```kotlin
val a : String.() -> Unit = {
  val length = 3.0f
  val x : Float = length
}
```
In short, the destructuring this proposal suggests would be similar to the semantics of delcaring something new through `val`.
> 
> I would suggest to go with `if` instead of `where` and also allow it to be used in a case such as this
> 
> ```kotlin
> when(val item = getItem()){
> 	is Book -> ...
> 	is Bottle -> ...
> 	if(item.canBreak()) -> ...
> 	// or potentially
> 	if item.canBreak ->
> 	else -> ...
> }
> ```

I fail to see how this syntax combines a type test and a guard. Could you please provide another example? For example, an equivalent of:
```kotlin
when(val x = getSomeObject()) {
    is Bottle where { it.canBreak() } -> // x is a breakable Bottle
    is Bottle -> // x is an unbreakable Bottle
    is Book -> // x is not a bottle at all
    else ->
}
```
where only a `Bottle` has `canBreak()` (and not a `Book`).

Or did I misunderstand and you meant something along the lines of:
(using my example above)
```kotlin
when(val x = getSomeObject()) {
    is Bottle if (x.canBreak()) -> // x is a breakable Bottle
                  else ->    // x is an unbreakable Bottle
    is Book -> // x is not a bottle at all
    else -> 
}
```
If this is the case, I do personally like not having to write the type test for each match. It also acts more like Haskell pattern matching: a type match is performed, then a series of guards, then the next type match.

Kroppeb:
> I fail to see how this syntax combines a type test and a guard. Could you please provide another example? For example, an equivalent of:
> 
> ```kotlin
> when(val x = getSomeObject()) {
>     is Bottle where { it.canBreak() } -> // x is a breakable Bottle
>     is Bottle -> // x is an unbreakable Bottle
>     is Book -> // x is not a bottle at all
>     else ->
> }
> ```
> 
> where only a `Bottle` has `canBreak()` (and not a `Book`).
> 
> Or did I misunderstand and you meant something along the lines of:
> (using my example above)
> 
> ```kotlin
> when(val x = getSomeObject()) {
>     is Bottle if (x.canBreak()) -> // x is a breakable Bottle
>                   else ->    // x is an unbreakable Bottle
>     is Book -> // x is not a bottle at all
>     else -> 
> }
> ```
> 
> If this is the case, I do personally like not having to write the type test for each match. It also acts more like Haskell pattern matching: a type match is performed, then a series of guards, then the next type match.

The example I write was not about pattern matching.
My suggested syntax would be 
```kotlin
when(val x = getSomeObject()) {
     is Bottle if (x.canBreak()) -> // x is a breakable Bottle
     is Bottle -> // x is an unbreakable Bottle
     is Book -> // x is not a bottle at all
     else ->
 }
```
However what I was talking about was having the type test optional and just having a guard alone also be valid, so you could rewrite the following
```kotlin
val x = getSomeObject()
when{
	x is Book -> // use book
	isTrash(x) -> // throw away
	x is Chair -> // use chair
	x is Table -> // use table
}
```
into
```kotlin
when(val x = getSomeObject()){
	is Book -> // use book
	if(isTrash(x)) -> // throw away
	is Chair -> // use chair
	is Table -> // use table
}
```

Cottand:
As per @elizarov 's and @Kroppeb 's suggestions, I decided to include an alternative `if`
syntax for guards in the proposal. If it proves popular or someone from the Kotlin team
backs it, I am happy with replacing the `where` syntax altoghether and change all exmaples
accordingly.

The same goes for other 'alternatives' of the proposal (like writing nested `is` and 
explicitly writing `val`, etc).

Personally, I am happy with this suggestion as it allows chained guards, and looks more
like the already familiar `if` structure.

> However what I was talking about was having the type test optional and just having a guard alone also be valid

Additionally, I think this proposal is mainly concerned with pattern matching and this suggestion
could deserve its own separate discussion. But again, if it proves popular I could inlcude it.

Wlodas:
Relating to the original proposal:
```kotlin
val result = when(download) {
  is App(name, Person("Alice", _)) -> "Alice's app $name"
  is Movie(title, Person("Alice", _)) -> "Alice's movie $title"
  is App, Movie -> "Not by Alice"
}
```
This probably won't work as expected, because **name** and **title** variables are just read from outer scope. Full example:
```kotlin
val name = "someName"
val title = "someTitle"

val result = when(download) {
  is App(name, Person("Alice", _)) -> "Alice's app $name"
  is Movie(title, Person("Alice", _)) -> "Alice's movie $title"
  is App, Movie -> "Not by Alice"
}
```
To indicate that we want to extract data and not just match variable from outer scope we should somehow declare "new" variables. For example like this:
```kotlin
val result = when(download) {
  is App(val name, Person("Alice", _)) -> "Alice's app $name"
  is Movie(val title, Person("Alice", _)) -> "Alice's movie $title"
  is App, Movie -> "Not by Alice"
}
```
Another thing I don't like in current imlemetation is how declared subject variable is just smart-casted, but you can't declare new variable name for each case. This is how I would like it to work more less:
```kotlin
val result = when(download) {
  is val app: App -> "App: $app"
  is val movie: Movie -> "Movie: $movie"
}
```
or without **val** keyword:
```kotlin
val result = when(download) {
  is app: App -> "App: $app"
  is movie: Movie -> "Movie: $movie"
}
```

Cottand:
Hi @Wlodas !
All the remarks you make are addressed in the proposal already, so please read carefully before posting. I'll go through them anyway:

> This probably won't work as expected, because name and title variables are just read from outer scope

As defind in the [Semantics](https://github.com/Cottand/KEEP/blob/pattern-matching/proposals/pattern-matching.md#semantics-) section, this is a compile error (conflicting declarations). It's clear and will avoid unexpected mistakes.

> To indicate that we want to extract data and not just match variable from outer scope we should somehow declare "new" variables.

Discussed [as an alternative](https://github.com/Cottand/KEEP/blob/pattern-matching/proposals/pattern-matching.md#specifying-extraction-with-val-) to current syntax. 

> Another thing I don't like in current implementation is how declared subject variable is just smart-casted, but you can't declare new variable name for each case.

I assume that waht you want is to 'name' a match. This is briefly discussed [here](https://github.com/Cottand/KEEP/blob/pattern-matching/proposals/pattern-matching.md#alternatives), and is called an as-pattern or binder. In the example you provide, you don't even extract new variables, you simply rename the `download` variable. I personally think smart casting covers this use case, and that the specific example you give is not particularily tied to pattern matching.

LifeIsStrange:
The Java pattern matching proposal has recently evolved, if you're interested for inspiration:
https://mail.openjdk.java.net/pipermail/amber-spec-experts/2020-August/002318.html

hfhbd:
FYI: This would be the Swift code, which also uses `let/val` as variable declaration. But Swift has no auto casting, so you will need a `title`, `name` variable in the case. https://docs.swift.org/swift-book/LanguageGuide/ControlFlow.html#ID129
```swift
struct Person: Equatable {
    let name: String
}

enum Download: Equatable {
    case app(name: String, developer: Person)
    case movie(title: String, director: Person)
}

let download: Download = .app(name: "App", developer: Person(name: "Alice"))
        
switch download {
    case .app(let name, Person(name: "Alice")):
        print("Alice's app \(name)")
    case .movie(let title, Person(name: "Alice")):
        print("Alice's movie \(title)")
    default:
        print("Not by Alice")
}
```
You can use `where` and move `let` to the front too: 
```swift
case .movie(let title, let person):
    print("\(person.name)'s movie \(title)")    
// you can move the let to the front
case let .movie(title, person):              
    print("\(person.name)'s movie \(title)")
// using with another where clause
case let .movie(title, person) where title.count <= 3: 
    print("\(person.name)'s movie with a short title: \(title)")
```

elizarov:
I've reopened https://youtrack.jetbrains.com/issue/KT-186 language design request for Pattern Matching and added a "wish list" for features and qualities that a good Kotlin-style design for pattern matching should strive to achieve. That issue might be a better place to discuss what pattern matching in Kotlin should and should not do since it does not propose any particular design, but just establishes some basics and calls for a more open-ended discussion, consistently with the [Contribution Guidelines for the Kotlin language](https://github.com/Kotlin/KEEP#contributing-use-cases-and-specific-enhancement-proposals) as spelled out in this KEEP's readme.

Cottand:
Thanks for re-opening @elizarov 

I had been meaning to discuss this proposal's shortcomings (in particular nominal destructuring). I posted over at https://youtrack.jetbrains.com/issue/KT-186

LifeIsStrange:
@elizarov Java 17 is officially getting pattern matching support:
https://mail.openjdk.java.net/pipermail/jdk-dev/2021-May/005579.html

elizarov:
Let's move all further discussion to https://youtrack.jetbrains.com/issue/KT-186 Closing this one.

PR #220, opened 2020-09-02
Update serialization KEEP with actual for 1.4 and 1.0.0 information

sandwwraith:


proposals/extensions/serialization.md
```diff
@@ -67,38 +68,38 @@ The following interfaces are provided in the library to be implemented by the se
...
 interface SerialDescriptor {
-    val name: String
+    val serialName: String
     val kind: SerialKind
 
     fun getElementName(index: Int): String
     fun getElementIndex(name: String): Int
 
-    fun getEntityAnnotations(): List<Annotation>
+    val annotations: List<Annotation>
```
qwwdfsad, line 102: let's move it to other vals then

proposals/extensions/serialization.md
```diff
@@ -67,38 +68,38 @@ The following interfaces are provided in the library to be implemented by the se
 
 ```kotlin
 interface SerializationStrategy<in T> {
-    fun serialize(output: Encoder, obj : T)
+    fun serialize(encoder: Encoder, value: T)
     val descriptor: SerialDescriptor
```
qwwdfsad, line 72: Here it also would be nice to comply our own code style and have vals before functions

proposals/extensions/serialization.md
qwwdfsad: also, `isNullable` and `elementsCount`

proposals/extensions/serialization.md
```diff
@@ -168,54 +173,55 @@ interface Encoder {
...
+    fun <T : Any?> encodeSerializableValue(serializer: SerializationStrategy<T>, value: T) =
         strategy.serialize(this, value)
-    
+
     // more methods to follow ...
 }
 ```
 
-However, this interface is not able to write composite data like classes, collections, or unions, because there is no way to express delimiters and metadata about the entity itself. 
+However, this interface is not able to write composite data like classes, collections, or unions, because there is no way to express delimiters and metadata about the entity itself.
```
qwwdfsad, line 186: We no longer have a notion of union in here

proposals/extensions/serialization.md
```diff
@@ -168,54 +173,55 @@ interface Encoder {
...
 }
 ```
 
 > Note, that this interface lacks functions similar to `encodeNotNullMark` and `encodeNull` by design.
-In general, generated class serializers support only non-nullable values of the corresponding type, 
-which is default in the idiomatic Kotlin code. Nullable types are handled by wrapping the 
-original serializer and are encoded via invocation of `encodeSerializableElement`. 
+In general, generated class serializers support only non-nullable values of the corresponding type,
+which is default in the idiomatic Kotlin code. Nullable types are handled by wrapping the
+original serializer and are encoded via invocation of `encodeSerializableElement`.
```
qwwdfsad, line 219: or `encodeNullableSerializableElement` as an optimization?

proposals/extensions/serialization.md
```diff
@@ -235,56 +240,44 @@ interface Decoder {
...
+    fun <T : Any?> decodeSerializableValue(deserializer: DeserializationStrategy<T>): T =
         strategy.deserialize(this)
-    
+
     // more methods to follow ...
 }
 ```
 
-However, an opposite interface for `ElementEncoder` would not be very useful. Primitives in the input stream can go in an arbitrary order, so in addition to methods opposite to `encodeXxxElement`, there is a `decodeElementIndex` method, which is called to determine the current position in the structure.
+However, an opposite interface for `ElementEncoder` would not be very useful. Primitives in the input stream can go in an arbitrary order, so in addition to methods opposite to `encodeXxxElement`, there is a `decodeElementIndex` method, which is called to determine the current position in the structure. The index returned from this method should be used as an argument to `decodeXxxElement` to insert value in the correct position in the structure.
```
qwwdfsad, line 253: `ElementEncoder` -> `CompositeEncoder`?

proposals/extensions/serialization.md
```diff
@@ -356,24 +355,18 @@ When generating the implementation methods, the compiler plugin needs to choose
...
-Currently, this package contains serializers for the following types: `Array`, `(Mutable)List`, `ArrayList`, `(Mutable)Set`, `LinkedHashSet`, `(Mutable)Map`, `LinkedHashMap`, `Pair`, `Triple`, and all primitive types.
-This package is a part of `kotlinx.serialization` runtime library.
+### Serializers for the standard library types
 
-### Auto-discovering user-defined external serializers
-
-To give the compiler plugin a hint about an external serializer, the annotation `@Serializable(with: KClass<*>)` can be applied on a property. However, it can be boilerplate-ish to annotate every property if you have a large number of domain classes which have to use, for example, an external serializer for `java.util.Date`.
-For this purpose, the annotation `@Serializers(vararg s: Serializer<*>)` is introduced. It can be applied to a class or to a file, and adds the specified serializers to the scope of the compiler plugin.
+Currently, this package `kotlinx.serialization.internal` contains serializers for the following types: `Array`, `(Mutable)List`, `ArrayList`, `(Mutable)Set`, `LinkedHashSet`, `(Mutable)Map`, `LinkedHashMap`, `Pair`, `Triple`, and all primitive types.
```
qwwdfsad, line 361: Also, `Map.Entry`

proposals/extensions/serialization.md
```diff
@@ -356,24 +355,18 @@ When generating the implementation methods, the compiler plugin needs to choose
...
+This package is a part of `kotlinx.serialization` runtime library.
```
qwwdfsad, line 362: Here and everywhere `runtime library` -> `core library` or `companion library`?

qwwdfsad (changes requested):

proposals/extensions/serialization.md
```diff
@@ -2,8 +2,7 @@
 
 * **Type**: Design proposal
 * **Authors**: Leonid Startsev, Roman Elizarov, Vsevolod Tolstopyatov
-* **Status**: Draft, Submitted
-* **Prototype**: Partially implemented, bundled as a separate plugin in Kotlin 1.3
+* **Status**: Implemented in 1.4 as a compiler plugin.
```
qwwdfsad, line 5:
```suggestion
* **Status**: Implemented in 1.4 as a compiler plugin and standalone library.
```

qwwdfsad (approved):

PR #222, opened 2020-10-04
Fix a note about lambdas and suspending functions in KEEP

vadimsemenov:
I assume _suspending_ lambdas may invoke suspending functions whenever they want.


elizarov:
I fact, this is a note about _regular lambda_. If non-local return is allowed, then suspension is allowed.

vadimsemenov:
Ok, clarified that it's about _regular_ lambda.

elizarov (approved):
Thanks

PR #229, opened 2020-12-08
Simple collection literals

HughG:
A proposal for a hopefully simple approach to collection literals which would be flexible in just the right Kotlin-y sort of way.  Please discuss under the related issue #230, as described by the KEEP README.md -- at least, that's what I think it's asking for.

### DIFF
--- a/proposals/simple-collection-literals.md
+++ b/proposals/simple-collection-literals.md
@@ -2,7 +2,7 @@
 
 * **Type**: Design proposal
 * **Author**: [Hugh Greene](https://github.com/HughG)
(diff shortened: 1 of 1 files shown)


elizarov:
I'm closing this issue and the corresponding PR in favor of moving the discussion from the subject of a particular design to the discussion of use-cases and requirements that such a design shall meet. So far, the proposed design does not meet all the basic requirements that were identified for support of collection literals in Kotlin.

Please, follow the YouTrack issue https://youtrack.jetbrains.com/issue/KT-43871 for further details and cast your votes there 👍

PR #243, opened 2021-03-13
Fix typo in value-classes.md

naftalmm:


elizarov:
That is not a typo. That's the whole idea to compile `VArray<Color>` to `int[]` as opposed to a `Color[]` which contains boxed version.

PR #244, opened 2021-03-19
Revise Result type for Kotlin 1.5, remove restrictions

elizarov:
* **Kotlin 1.5**
  * Allow returning the `Result` type from functions.
  * Allow Kotlin null-safety operators `?.`, `?:` and `!!` on both nullable and non-null `Result` types.
  * Text updated to replace `inline class` with `@JvmInline value class`.

### DIFF
--- a/proposals/stdlib/result.md
+++ b/proposals/stdlib/result.md
@@ -749,5 +749,5 @@ that provide the corresponding utilities.
 
 * **Kotlin 1.5**
   * Allow returning the `Result` type from functions.
-  * Allow Kotlin null-safety operators `.?`, `?:` and `!!` on both nullable and non-null `Result` types.
+  * Allow Kotlin null-safety operators `?.`, `?:` and `!!` on both nullable and non-null `Result` types.
   * Text updated to replace `inline class` with `@JvmInline value class`.
--- a/proposals/stdlib/result.md
+++ b/proposals/stdlib/result.md
@@ -565,21 +565,21 @@ and all of them are purely tentative.
 
 ### Representing as a sealed class
 
-Kotlin `inline` classes cannot be currently used with `sealed class` construct. 
+Kotlin `@JvmInline value` classes cannot be currently used with `sealed class` construct. 
 If that is supported in the future, then we could change implementation of 
 `Result` without affecting its public APIs and binary interfaces in the following way:
 
 ```kotlin
-sealed inline class Result<T> {
-    inline class Success<T>(val value: T) : Result<T>()
+@JvmInline sealed value class Result<T> {
+    @JvmInline class Success<T>(val value: T) : Result<T>()
     class Failure<T>(val exception: Throwable) : Result<T>()
 }
 ``` 
 
-> Notice, that only `Success` case is marked with `inline` modifier here. That is the case that should be
-represented without boxing. In general, if `inline sealed` classes are allowed in the future,
-then Kotlin compiler could only support `inline` modifier on a set of subclasses with pairwise non-intersecting 
-types of their primary constructor properties. In particular, both `Success` and `Failure` cannot be `inline` 
+> Notice, that only `Success` case is marked with `@JvmInline` annotation here. That is the case that should be
+represented without boxing. In general, if `@JvmInline sealed value` classes are allowed in the future,
+then Kotlin compiler could only support `@JvmInline` annotation on a set of subclasses with pairwise non-intersecting 
+types of their primary constructor properties. In particular, both `Success` and `Failure` cannot be `@JvmInline` 
 at the same time, since we would not be able to distinguish `Success(Exception(...))` from 
 `Failure(Exception(...))` at run time.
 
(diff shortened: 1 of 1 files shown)


proposals/stdlib/result.md
```diff
@@ -347,38 +349,40 @@ Functions that declare generic result types may, in fact, return values of `Resu
...
 
-Private and local properties of `Result`  type are allowed as long as they don't have custom getters:
+Private and local properties of `Result`  type were allowed as long as they did not have custom getters:
 
 ```kotlin
 private var foo: Result<Int> // Ok
 ```
 
-The use of Kotlin null-safety operators `.?`, `?:` and `!!` is not allowed on both nullable and non-null `Result` types:
+The use of Kotlin null-safety operators `.?`, `?:` and `!!` was not allowed on both nullable and non-null `Result` types:
```
udalov, line 358:
```suggestion
The use of Kotlin null-safety operators `?.`, `?:` and `!!` was not allowed on both nullable and non-null `Result` types:
```

proposals/stdlib/result.md
```diff
@@ -726,3 +744,10 @@ if that is needed for some business function
 In cases when you need to distinguish between different kinds of failures and these approaches do not work for you, 
 you are welcome to write your own utility libraries or use libraries like [Arrow](https://arrow-kt.io) 
 that provide the corresponding utilities.    
+
+## Revision history
+
+* **Kotlin 1.5**
+  * Allow returning the `Result` type from functions.
+  * Allow Kotlin null-safety operators `.?`, `?:` and `!!` on both nullable and non-null `Result` types.
```
udalov, line 752:
```suggestion
  * Allow Kotlin null-safety operators `?.`, `?:` and `!!` on both nullable and non-null `Result` types.
```

ilya-g (approved):
+1 to @udalov suggestions.

arhont375:
Just read this Keep as updated in this PR and there still places where:
1. Said that `However, writing functions that return Result is not allowed in Kotlin` in the Appendix seciton.
2. `sealed inline class` used in the future improvements section, should it be `sealed value class`?

I'm not sure if it's intended or not, but such inconsistency confused me during read

elizarov:
@arhont375 

> 1. Said that `However, writing functions that return Result is not allowed in Kotlin` in the Appendix seciton.
> 2. `sealed inline class` used in the future improvements section, should it be `sealed value class`?

Good catch. Thanks. Fixed.

LouisCAD:
`runCatching` would catch any `Throwable`, including `CancellationException`, which becomes a problem when your code rely on them to be thrown to be cancellable.

To alleviate that significant problem, I'm suggesting to make the lambda of all the `runCatching` functions `crossinline`, so one cannot run suspending, and likely cancellable code into a construct that would eat what powers the cancellation mechanism.

That safety measure would limit usage of `runCatching`, but with the benefit of making it impossible to misuse, ans encouraging people to define their own, domain specific error types, until possible new facilities come into the language or libraries to make it less effort to implement.

ilmirus:
@LouisCAD fixing `runCatching` does not fix other problems with `CancellationException`. To fix the issue, we should make structured concurrency a language feature instead of library one. I.e. treat cancellation as we treat suspension. We do not have `SuspendedException`, after all.

ilmirus (approved):

LouisCAD:
Sure @ilmirus , but in the meantime, it'd avoid a lot of bugs, and that crossinline can always be dropped later without affecting anyone.

Is structured concurrency as a language feature something we can expect in the not too distant future?

gmk57:
> `runCatching` would catch any `Throwable`, including `CancellationException`, which becomes a problem when your code rely on them to be thrown to be cancellable.
> 
> To alleviate that significant problem, I'm suggesting to make the lambda of all the `runCatching` functions `crossinline`, so one cannot run suspending, and likely cancellable code into a construct that would eat what powers the cancellation mechanism.

I've hit this too, but current behavior makes sense if you regard `runCatching` just as a fancy `try/catch`.

Personally, I often use `runCatching` as a top-level wrapper around a whole bunch of suspendable work. There is no heavy lifting after it, so non-propagating cancellation is not a real issue. Making lambda `crossinline` would break this use case.

Now that `Result` is non-restricted we can freely make our own variants of `myRunCatching`, e.g. with
`if (e is CancellationException) throw e`

And in some (rare) cases it is preferrable to treat `CancellationException` exactly like any other exception, e.g. in some Repository:
`.onFailure { dataLoading = false }`

camhashemi:
I'm looking for a linked discussion or proposal for why these restrictions were lifted. I've just been catching up on [the original keep](https://github.com/Kotlin/KEEP/blob/master/proposals/stdlib/result.md) and [discussion](https://github.com/Kotlin/KEEP/issues/127), but I don't see anything around this restriction being removed. So there's a gap for me between "Result is only allowed for restricted use cases" and "Result is only encouraged for restricted use cases".

elizarov:
>@camhashemi is there a discussion or proposal we can link to for why these restrictions were lifted?

The original keep mentioned some potential future enhancements to support null-related operators like `?.` and `?:` to have a special effect for result types. We've tried to design this approach in more detail and did not find any feasible way forward. With this potential enhancement off the table, there was no need to restrict usages of the result type anymore.

PR #247, opened 2021-04-09
Introduce an explicit `copy var` notation

elizarov:
Taking into account use-cases where a value class is used as an immutable handle into a mutable data structure and the need to gradually teach developers new features where properties of immutable classes are somehow updated, we've decided to give an explicit name to the corresponding concept -- a "copyable property", and introduce a corresponding `copy` modifier both for "copyable properties" (`copy var`) and for copying functions (`copy fun`).

The text was updated to consistently use the corresponding terminology and the `copy` modifier. Motivation for the change was added to the text, too.

Restriction on the use of `var` properties for value classes is removed.

### DIFF
--- a/notes/value-classes.md
+++ b/notes/value-classes.md
@@ -20,31 +20,31 @@ language design rule.  Any new feature in the language has to clear a high bar o
   * [Built-in primitive value classes](#built-in-primitive-value-classes)
   * [Inline classes are user-defined value classes](#inline-classes-are-user-defined-value-classes)
   * [Deep immutability vs shallow immutability](#deep-immutability-vs-shallow-immutability)
-  * [Compiling value classes: the single-field restriction](#compiling-value-classes-the-single-field-restriction)
+  * [Compiling value classes: the single-field restriction](#compiling-value-classes:-the-single-field-restriction)
   * [Project Valhalla](#project-valhalla)
   * [Multifield value classes before Valhalla](#multifield-value-classes-before-valhalla)
   * [Value classes vs structs](#value-classes-vs-structs)
 * [Immutability and value classes](#immutability-and-value-classes)
   * [Updating immutable classes](#updating-immutable-classes)
-  * [Mutable properties of immutable types](#mutable-properties-of-immutable-types)
-  * [Compiling mutable properties on JVM](#compiling-mutable-properties-on-jvm)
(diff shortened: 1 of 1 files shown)


notes/value-classes.md
```diff
@@ -422,37 +444,40 @@ fun State.updatedNow(): State {
 }
 ```
 
-This gives no improvement and is still far cry from the simplicity of working with mutable classes. So, members and extensions of value classes should support a mechanism to declare a function with a mutable receiver, that is, a function that exposes `this` receiver reference as `var` and implicitly returns the updated value of `this`. The best way to achieve it is via a modifier for a function declaration:
+This gives no improvement and is still far cry from the simplicity of working with mutable classes. So, members and extensions of value classes should support a mechanism to declare a function with a copyable receiver, that is, a function that exposes `this` receiver reference as `copy var` and implicitly returns the updated value of `this`. The best way to achieve it is via a modifier for a function declaration:
 
 ```kotlin
-mutating fun State.updateNow() { // implicitly returns new State
+copy fun State.updateNow() { // implicitly returns new State
```
fvasco, line 450: Can a `copy fun` have a return type?
I suppose no.

benjishults: But see [Mutating functions returning a value](https://github.com/Kotlin/KEEP/blob/master/notes/value-classes.md#mutating-functions-returning-a-value)

fvasco: So integration with JVM should be a different than described in "Compiling copyable properties on JVM".

elizarov: Returning a value requires a different JVM impl strategy. Either returning a pair or accepting an additional "ref" parameter. The former can be implemented efficiently with Valhalla, so that is what we'll prefer to do if we end up implementing it prior to Valhalla (since it will be easier to migrate to Valhalla later)

kyay10: Maybe you could actually do both though? like you could possibly introduce it as an annotation so that for example if you decide that returning a pair is the default one can tell the compiler to compile it as a ref instead (or compile it as both even) with an annotation (or a compiler argument) because I feel like maybe refs could be easier to optimize away with a good reducer like R8 while returned Pairs are probably harder to reason about (although I'm just guessing here with no real evidence lol)

PR #251, opened 2021-04-14
YouTrack links to the new issues

zhelenskiy:
As following [issues](https://github.com/Kotlin/KEEP/issues/237#issuecomment-818954944) are created, it may be better to mention them in the text.


nikitabobko:
There are a lot of different immutability issues scattered around the YouTrack, I'm not sure that we want to reference specifically these

PR #256, opened 2021-05-25
Fix typo

vincentlauvlwj:
Correct the conclusion text in the type mapping section - inline classes over not-null reference types.


elizarov:
The original text is actually correct.

vincentlauvlwj:
image

@elizarov If the original text is correct, what's the sense of giving the example of `bar` function here?

elizarov (approved):
Ah yes, thanks a lot. I've misread it, too. You are absolutely right.

PR #264, opened 2021-06-24
Context receivers KEEP: Improve "Referencing specific receiver" section

shadrina:


proposals/context-receivers.md
```diff
@@ -278,11 +278,12 @@ signature of the functional type replicates the textual order in which every arg
   }
   ```
 
-### Referencing specific context receiver
+### Referencing specific receiver
```
elizarov, line 281: Table of contents needs to be fixed, too. I'm using knit to generate it -> Clone https://github.com/Kotlin/kotlinx-knit, open it in IDEA, num Knit main file, pass it it the path to the file that needs processing.

elizarov (changes requested):

proposals/context-receivers.md
shadrina: Thanks, updated

PR #267, opened 2021-07-06
Instantiation of annotation classes

sandwwraith:
Allow calling constructors of annotation classes in arbitrary code
to obtain an instance of a particular annotation class.

### DIFF
--- a/proposals/annotation-instantiation.md
+++ b/proposals/annotation-instantiation.md
@@ -1,0 +1,66 @@
+# Instantiation of annotation classes
+
+* **Type**: Design proposal
+* **Author**: Leonid Startsev
+* **Status**: In review
+* **Prototype**: In progress
+* **Issue:** [KT-45395](https://youtrack.jetbrains.com/issue/KT-45395)
+
+## Synopsis
+
+Allow calling constructors of annotation classes in arbitrary code
+to obtain an instance of a particular annotation class.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


fvasco:
It is not clear if two instances of the annotation class `A` of different modules, ie:

```kotlin
// module a
annotation class A

// module1
val a1 = A()

// module2
val a2 = A()
```

the following statements are always true

```kotlin
a1::class == a2::class
a1::class.java == a2::class.java
a1.javaClass === a2.javaClass
```

Violating these assumptions can be misleading.

JakeWharton:
None of those statements are always true today when dealing with annotation instances from JVM reflection or when implementing instances manually (or using AutoAnnotation). The only safe way to compare the types (on the JVM) is by calling `annotationType()` and doing a comparison.

fvasco:
@JakeWharton, I understand that and I agree with you.
I think that this new feature put an unexpected behavior in the Kotlin language due to implementation/platform details.
Moreover, this behavior is unspecified and can be different between JVM and native.

I wish to suggest adding this "open issue"/"current limitation"/"implementation detail" in the documentation as a further reference.

It is possible to define a shared implementation of the annotation `A` in the "module a" to partially fix this issue, at least for a pure Kotlin project.
There are good reasons to avoid that and I agree with it, however, these and their consequences are unmentioned.

elizarov (approved):
👍 Great job

PR #269, opened 2021-07-19
Add definitely-non-nullable-types.md

dzharkov:


udalov:
Minor, but the feature should probably be named "definitely **non**-nullable types". "Not" isn't used as a part in hyphenated words in English AFAIK. Another alternative is "definitely not nullable types" but that sounds weirder ("- nullable types? Definitely not!").

dzharkov:
@udalov Thanks, will fix it!

elizarov (approved):

Issue #257, opened 2021-05-27
Repeatable annotations


udalov:
The goal of this proposal is to extend the existing Kotlin feature of repeatable annotations to allow binary- and runtime-retained repeatable annotations, while making it fully interoperable with repeatable annotations in Java.

Text: https://github.com/Kotlin/KEEP/blob/master/proposals/repeatable-annotations.md

zarechenskiy:
Closing the issue as the feature was graduated to Stable in Kotlin 1.6

PR #270, opened 2021-07-25
Update repeatable annotations KEEP (#257) informed by implementation

udalov:


proposals/repeatable-annotations.md
```diff
@@ -130,19 +134,19 @@ annotation class Tag(val name: String) {
     // @Target(CLASS, FUNCTION)
     // @Retention(BINARY)
     // @kotlin.jvm.internal.RepeatableContainer
-    // /* synthetic */ annotation class Container(val value: Array<Tag>)
+    // public annotation class Container(val value: Array<Tag>)
 }
 ```
 
-Another error is introduced in case the container annotation is applied manually at the same time as the contained annotation:
+Another error is introduced in case the container annotation is applied manually at the same time as the contained annotation, if the latter is repeated:
```
elizarov, line 141: But is it legal in Java to have both a container and a single copy of annotation? Is it available via Java reflection's `getAnnotationsByType`? If we support it just for the rare case of backwards compatibility, then we should have a migration plan to get rid of it.

udalov: Yes, apparently it is legal in Java, and `getAnnotationsByType` returns both annotations. (The previous text about `getAnnotationsByType` in the proposal was incorrect.)
```
@Tag("1")
@Tags(@Tag("2"))
public class A {
    public static void main(String[] args) {
        System.out.println(Arrays.asList(A.class.getAnnotationsByType(Tag.class)));
    }
}
```
->
```
[@test.Tag(value="1"), @test.Tag(value="2")]
```

With this in mind, I don't see much value in prohibiting this case. It's not especially confusing either, just a bit weird...

elizarov: Ok.

elizarov (approved):

PR #271, opened 2021-07-30
OptIn updates in 1.5.30 / 1.6.0

mglukhikh:


proposals/opt-in.md
```diff
@@ -0,0 +1,289 @@
...
+Both opt-in mechanisms allow to use the API for the selected markers anywhere in the parse tree lexically under the annotated element.
+
+Using `OptIn` with annotations that are *not* opt-in requirement markers has no effect and yields a compilation warning. (Note that this must not be an error because user code should not break once an annotation is no longer an opt-in requirement marker.) Using `OptIn` with no arguments has no effect and yields a warning as well.
+
+## Opt-in requirement message
+
+When using an API that requires opt-in without the said opt-in, the compiler reports a warning or an error depending on the specified level. It's possible to specify a custom message in `RequiresOptIn`, that will be reported by the compiler. If no message is given (i.e. if `message` is empty), the compiler will report that the API is *experimental*:
+
+```
+test.kt:22:9: error: this declaration is experimental and its usage must be marked with '@ShinyNewAPI' or '@OptIn(ShinyNewAPI::class)'
```
elizarov, line 110: There's a separate discussion in https://youtrack.jetbrains.com/issue/KT-47925 on making a better message and it's great to correct it in this KEEP, too.

proposals/opt-in.md
```diff
@@ -0,0 +1,289 @@
...
+
+// Usage:
+
+// Function useFooAndBar uses API marked with ShinyNewAPI and thus
+// is required to opt in to that marker. Here, we choose a propagating opt-in,
+// because the API is used in the function signature (parameter foo's type).
+// (But this is not enforced by the compiler in any way, we could've chosen
+// a non-propagating opt-in as well.)
+@ShinyNewAPI
+fun useFooAndBar(foo: Foo) {
```
elizarov, line 85: This is an outdated example and comment that was made obsolete with the recent changes. I suggest expanding an example section:

1. More this `useFooAndBar` to the end of the examples, mark it with `@OptIn` and explain that, even though it is an `@OptIn`, every usage will still get a warning or error, because it references an opt-in type.

2. Change `@ShinyNewAPI`-marked example here, so that it is similar to the `doSomething` below (does not have any parameters) but its author chose to mark it with `@ShinyNewAPI`. Name it something like `shinyNewFeature` and explain that this function is also, conceptually, a part of `ShinyNewAPI` and so it is marked as such. For contrast, `doSomething` example's comment can explain that this is an old function that is was always a part of the library and the fact that it is now using shiny new API is just its implementation detail.

proposals/opt-in.md
```diff
@@ -0,0 +1,289 @@
...
+```
+
+## OptIn marker contagiousness
+
+In Kotlin 1.5.30 we introduced contagiousness rules based on type usages.
+As a rule of thumb, all places which break if some experimental type disappear
+from a library should receive opt-in usage warning/error, 
+even if this experimental type is used implicitly.
+
+Some type is considered "requiring opt-in marker" if its constructor
```
elizarov, line 236: The word "constructor" is confusing here. I'd suggest rephrasing it clearer.

elizarov (changes requested):

proposals/opt-in.md
```diff
@@ -0,0 +1,289 @@
...
+    val s = foo()
+    bar()
+    null.baz()
+}
+```
+
+All three calls `foo()`, `bar()` and `baz()` inside `fun use()` use `Shiny` type implicitly,
+thus all three should receive opt-in usage warning/error. We can negate it using either `@ShinyNewAPI`
+or `@OptIn(ShinyNewAPI::class)` at `fun use()` itself. 
+Pay attention that `@OptIN(ShinyNewAPI::class)` at `foo`, `bar`, `baz` declarations does not help here.
```
shadrina, line 268:
```suggestion
Pay attention that `@OptIn(ShinyNewAPI::class)` at `foo`, `bar`, `baz` declarations does not help here.
```

proposals/opt-in.md
mglukhikh: I'd just delete the first example, because contagiousness is described at the end of the document separately.

mglukhikh: The second example was split into two different as you wish

elizarov (approved):
Good job

PR #273, opened 2021-08-10
scripting-support: language annotations for code blocks

xeruf:


elizarov (approved):
Thanks. Even though I have not figured out why `kt` language works. Github languages file https://github.com/github/linguist/blob/master/lib/linguist/languages.yml only lists `kotlin` as a name for Kotlin and defines no additional aliases.

xeruf:
I guess it also recognizes file extensions - `md` also works for Markdown etc.

PR #274, opened 2021-08-11
Context receivers: fix functional types sample

shadrina:
The current sample for contextual functional types seems incorrect since we do not support and do not propose the support of callable references for member extensions. I believe the new one shows the same and compiles.


elizarov (approved):
Great!

PR #277, opened 2021-09-15
Add support for explicit backing fields

lunakoly:
Please, see [explicit-backing-fields.md](https://github.com/lunakoly/KEEP/blob/explicit-backing-fields/proposals/explicit-backing-fields.md) for the full text.

The corresponding [issue](https://github.com/Kotlin/KEEP/issues/278).

### DIFF
--- a/proposals/explicit-backing-fields.md
+++ b/proposals/explicit-backing-fields.md
@@ -1,0 +1,273 @@
+# Explicit Backing Fields
+
+- **Type**: Design Proposal
+- **Author**: ?
+- **Contributors**: Svetlana Isakova, Kirill Rakhman, Dmitry Petrov, Roman Elizarov, Ben Leggiero, Matej Drobnič, Mikhail Glukhikh, Nikolay Lunyak
+- **Status**: Implemented in FIR
+- **Prototype**: Implemented
+- **Initial YouTrack Issue**: [KT-14663](https://youtrack.jetbrains.com/issue/KT-14663)
+- **Initial Proposal**: [private_public_property_types#122](https://github.com/Kotlin/KEEP/pull/122)
+
+## Summary
+
+**Note**: initial proposal contents have been partially copied down here for convenience. Despite the different approach, the already shown use cases are still relevant.
+
+>  Common pattern in java is having full type accessible as private property and then only exposing required interface in the public getter:
+>
+> ```java
+> class MyClass {
+> 	// Use full type for private access
+> 	private final ArrayList<String> data = new ArrayList<>();
+> 	
+> 	// Only expose what is needed in public getter
+> 	public List<String> getData() {
+> 		return data;
+> 	}
+> }
+> ```
+>
+> This pattern allows easy hiding of implementation details and allows for easy clean external interfaces to the classes. But there is no idiomatic Kotlin way to achieve this pattern. Best thing you can do is define two separate properties to mimic how Java does it:
+>
+> ```kotlin
+> class MyClass {
+>     private val _data = ArrayList<String>()
+>     val data: List<String>
+>         get() = _data
+> }
+> ```
+
+With the proposed syntax in mind, the above code snipped could be rewritten as follows:
+
+```kotlin
+class MyClass {
+    val data: List<String>
+        field = ArrayList()
+}
+```
+
+## Use Cases
+
+### Motivation
+
+> - There is no clean idiomatic way to do this pattern in Kotlin
+> - Current best approach throws away all benefits of Kotlin properties and forces developer to write Java-like code with separate private field and public getter
+> - Current best approach forces developer to assign two different names to single property (or pad private property, for example adding `_` prefix and then using this prefix everywhere in code)
+> - This is another place where Java pattern could be introduced into Kotlin with less boilerplate
+
+### Read-Only from Outside
+
+> We often do not want our data structures to be modified from outside. Unlike Java, this can be easily achieved in Kotlin by just exposing read-only `List`. But as already ilustrated in above example, exposing different type of a property is a bit messy.
+
+```kotlin
+private val _items = mutableListOf<Item>()
+val item : List<Item> by _items
+```
+
+And the new syntax allows us to write:
+
+```kotlin
+val items: List<Item>
+	private field = mutableListOf()
+```
+
+### Android Architecture Components
+
+>  Proper way to do Architecture components is to use `MutableLiveData` (`LiveData` implementation that allows caller to change its value) privately inside View Model classes and then only expose read-only `LiveData` objects outside.
+
+Sample code [from an Android app](https://github.com/elpassion/crweather/blob/9c3e3cb803b7e4fffbb010ff085ac56645c9774d/app/src/main/java/com/elpassion/crweather/MainModel.kt#L14):
+
+```kotlin
+private val mutableCity = MutableLiveData<String>().apply { value = "" }
+private val mutableCharts = MutableLiveData<List<Chart>>().apply { value = emptyList() }
+private val mutableLoading = MutableLiveData<Boolean>().apply { value = false }
+private val mutableMessage = MutableLiveData<String>()
+
+val city: LiveData<String> = mutableCity
+val charts: LiveData<List<Chart>> = mutableCharts
+val loading: LiveData<Boolean> = mutableLoading
+val message: LiveData<String> = mutableMessage
+```
+
+Becomes:
+
+```kotlin
+val city: LiveData<String>
+	field = MutableLiveData().apply { value = "" }
+val charts: LiveData<List<Chart>>
+	field = MutableLiveData().apply { value = emptyList() }
+val loading: LiveData<Boolean>
+	field = MutableLiveData().apply { value = false }
+val message: LiveData<String>
+	field = MutableLiveData()
+```
+
+### RX Observable and Subjects
+
+> Data can often be pushed into reactive world using subjects. However, exposing `Subject` would allow consumer of the class to push its own data into it. That is why it is good idea to expose all subjects as read-only `Observable`:
+
+```kotlin
+class MyClass {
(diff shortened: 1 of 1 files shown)


proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,317 @@
...
+	[visibility] field[: B] = initializer
+```
+
+The above `field` declaration is referred to as an _explicit backing field declaration_.
+
+#### Accessors
+
+- if `A :> B`, the compiler can provide a default getter
+- if `A <: B`, the compiler can provide a default setter
+- If an accessor is required, it must be provided either by the compiler or the user
```
mglukhikh, line 142: A bit unclear. In which situation "accessor is required and provided by compiler"? And what's the difference with situation when "accessor is not required"?

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,317 @@
...
+public val flow: SharedFlow
+    field: MutableSharedFlow? = null
+    get() {
+        return field ?: run { ... }
+    }
+```
+
+#### Visibility
+
+Only the `private` and the `internal` visibilities are allowed for explicit backing fields now. The default visibility is `private`.
```
mglukhikh, line 154: May be mention reasoning why don't we want `protected` one.

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,317 @@
...
+    lateinit field: String
+    get() = field.length
+    set(value) {
+        field = value.toString()
+    }
+```
+
+#### Initializer
+
+If there is an explicit backing field, it must have an initializer, and the property must not declare an initializer.
```
mglukhikh, line 178: > it must have an initializer

With an exception of `lateinit` case above

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,317 @@
...
+
+## Design
+
+The proposed design consists of two new ideas.
+
+### Explicit Backing Fields
+
+```kotlin
+val it: A
+	[visibility] field[: B] = initializer
```
mglukhikh, line 133: Not sure here but may be it's better to rename `A` to `P` (property) and `B` to `F` (field). This allows easier understanding of these abbreviations, especially in text far below.

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,317 @@
...
+```
+
+## Alternatives
+
+### Initial Proposal
+
+Initially, the following syntax was suggested:
+
+```kotlin
+public val items = mutableListOf<Item>()
```
mglukhikh, line 238: Initial example was reversed:
```
    private val items = mutableListOf<Item>()
        public get(): List<Item>
```

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,317 @@
...
+- **Initial YouTrack Issue**: [KT-14663](https://youtrack.jetbrains.com/issue/KT-14663)
+- **Initial Proposal**: [private_public_property_types#122](https://github.com/Kotlin/KEEP/pull/122)
+
+## Summary
+
+**Note**: initial proposal contents have been partially copied down here for convenience. Despite the different approach, the already shown use cases are still relevant.
+
+>  Common pattern in java is having full type accessible as private property and then only exposing required interface in the public getter:
+>
+> ```java
```
mglukhikh, line 17: Personally I think that Java example is good, but may be it shouldn't be in the title section (put it along with "Android architecture components"). About this place, I'd try to explain this pattern in pure Kotlin without direct mentioning of Java. @elizarov , do you have some other ideas here?

proposals/explicit-backing-fields.md
lunakoly: By "required" I meant "`val`s need getters, and `var`s need getters and setters", but ok, I'll paraphrase it

proposals/explicit-backing-fields.md
lunakoly: The thing is, we put `protected` away simply to let ourselves concentrate on a smaller task. Since STN is forbidden for non-`final` properties, there should be no ambiguity as to which backing field to use.

proposals/explicit-backing-fields.md
elizarov: +1. We should cater to Kotlin developer here. Frankly, I don't think that comparison with Java is actually useful in this document at all. Kotlin already has all the concepts: properties and backing field. Moreover, there is already an existing pattern of a "backing properties" that this design is striving to improve https://kotlinlang.org/docs/coding-conventions.html#names-for-backing-properties

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,322 @@
...
+```kotlin
+private val _items = mutableListOf<Item>()
+val item : List<Item> by _items
+```
+
+And the new syntax allows us to write:
+
+```kotlin
+val items: List<Item>
+	private field = mutableListOf()
```
elizarov, line 70: What's `private` is supposed to mean here before `field`? Also, from here and down the document something is wrong with the coding style (indentation suddenly becomes 8 spaces instead of 4).

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,322 @@
...
+	field = MutableLiveData().apply { value = "" }
+val charts: LiveData<List<Chart>>
+	field = MutableLiveData().apply { value = emptyList() }
+val loading: LiveData<Boolean>
+	field = MutableLiveData().apply { value = false }
+val message: LiveData<String>
+	field = MutableLiveData()
+```
+
+### RX Observable and Subjects
```
elizarov, line 104: Since it is already the year 2021 (and also taking into account readers from the future), it is way better to give this example using Kotlin Flow instead of Rx, e.g. with `MutableStateFlow<String>` that is exposed as `StateFlow<String>`.

proposals/explicit-backing-fields.md
elizarov: Yes. Also, for Kotlin/Native protected fields might be a challenge to implement.

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,322 @@
...
+```kotlin
+var someStrangeExample: Int
+    lateinit field: String
+    get() = field.length
+    set(value) {
+        field = value.toString()
+    }
+```
+
+#### Initializer
```
elizarov, line 179: This and the following chapters of the "design" section are too small. They should be merged into a single chapter and be better presented as separate paragraphs or maybe even as a bullted-list.

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,322 @@
...
+
+In fact, the above syntax brings in an incorrect mental model: it says _'There is a private property `it` with some *part* that declares its public behavior'_.
+
+Attempt to add support for the above syntax led to multiple redundant complications (see the [problems section](https://github.com/matejdro/KEEP/blob/private_public_property_types/proposals/private_public_property_types.md#questions-to-consider) + unclear override mechanics).
+
+## Future Enhancements
+
+### Direct Backing Field Access
+
+It might be handy to allow the direct access to the property backing field via syntax like `myProperty#field`.
```
elizarov, line 257: It is worth explicitly noting here that this is not the actually proposed syntax.

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,322 @@
...
+
+#### Accessors
+
+- if `P :> F`, the compiler can provide a default getter
+- if `P <: F`, the compiler can provide a default setter
+
+If the compiler can not provide a getter, the user must declare it explicitly. The same applies to setters in case of `var` properties.
+
+```kotlin
+public val flow: SharedFlow
```
elizarov, line 148: If you use really existing types in the example, then it is much better to use the real code in the example or some users might get confused that your feature also somehow supports the omission of type parameters. `SharedFlow` is a generic type, so it should be something like `SharedFlow<String>`. Alternatively, you can use imagined types in the examples.

proposals/explicit-backing-fields.md
lunakoly: It's true that `private` here is redundant. I thought, specifying it explicitly here would make the example look more consistent with the sample code above. It's like "look, we've had a `private` backing property, and now here is a `private` backing field"

proposals/explicit-backing-fields.md
lunakoly: This example would, indeed, illustrate the use of an explicit backing field, but since it's typical to use `asStateFlow()`, the smart type narrowing wouldn't work, and the example would be a bit misleading (apart from the STN there's currently no way to access the field value from outside the accessors)

proposals/explicit-backing-fields.md
lunakoly: Changed `private` to `internal` in this example

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,326 @@
...
+2. the type of that instance is compatible with the property type
+3. the backing field visibility allows a hypothetical direct access
+4. there is no ambiguity in such an access
+
+it can then narrow down the returned instance type at the call site without loss of functionality. This is what is meant by the words _smart type narrowing_.
+
+The formal checks corresponding to the above rules are:
+
+1. the property does not have a custom getter, only the default one
+2. `A :> B`
```
mglukhikh, line 215: Please replace with `P :> F` (to be honest, I suggested this rename mainly because of this place)

proposals/explicit-backing-fields.md
elizarov: Resolved in #289

proposals/explicit-backing-fields.md
elizarov: Fixed in #289

elizarov:
Superseded by #289. Please, add any further comments on the proposed text there.

PR #279, opened 2021-09-22
complete multi platform stdlib

smallshen:
need more discussion


gildor:
I don't think that this is a proper KEEP, please check this project README, I don't think this PR follows requirements, even this PR stands "need more discussion", which sounds right

The scope of the proposal is so huge and doesn't follow the existing development approach for stdlib, so I think it would be better do not to create this proposal, but instead discuss the general ideas on Kotlin Slack (I would suggest #language-evolution channel) or create a topic on [Kotlin forum](https://discuss.kotlinlang.org/).

smallshen:
agree, this might be too big as a proposal

PR #282, opened 2021-12-02
Actualize annotation instantiation proposal

sandwwraith:


sandwwraith:
Closed in favor of #324

PR #284, opened 2021-12-13
 Decommission `Enum.values()` and replace it with `Enum.entries`

qwwdfsad:


proposals/enum-entries.md
```diff
@@ -0,0 +1,174 @@
...
+  * Introduction of `values` property adds an unfixable method reference ambiguity for all the callers of `E::values`.
+* Java language [refers](https://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.9.1) to enum elements as `constants`. 
+  * The risks of introducing widely-used API named `constants` with the potential clashes with [Kotlin constant evaluation](https://youtrack.jetbrains.com/issue/KT-14652) outweighs the benefits of the straightforward name.
+* `valuesList` and `entryList` suffer the same disease -- excessive verbosity of the name.
+* `components` is already reserved in Kotlin for positional deconstructors.
+* Kotlin specification and documentation already refer to enum elements as entries, making it our final choice.
+
+### Translation strategy for Kotlin enums
+
+For source-compiled enums, a new static property with a JVM signature `public static List<E> getEntries()` is added,
```
elizarov, line 87: `List<E>` -> `EnumEntries<E>`

proposals/enum-entries.md
```diff
@@ -0,0 +1,174 @@
...
+enum class MyEnum {
+    A;
+}
+```
+
+has the following form (all irrelevant parts omitted):
+```java
+enum MyEnum extends Enum<MyEnum> {
+    private static final synthetic MyEnum[] $VALUES
+    private static final synthetic List<MyEnum> $ENTRIES;
```
elizarov, line 111: `List<MyEnum>` -> `EnumEntries<MyEnum>`

proposals/enum-entries.md
```diff
@@ -0,0 +1,174 @@
...
+        $VALUES = $values();
+        Supplier<MyEnum[]> supplier = #invokedynamic ..args.. values;
+        $ENTRIES = new EnumEntriesList(supplier);
+    }
+
+    public static MyEnum[] values() {
+        return $VALUES.clone();
+    }
+  
+    public static List<MyEnum> getEntries() {
```
elizarov, line 124: `List<MyEnum>` -> `EnumEntries<MyEnum>`

CLOVIS-AI:
What are the benefits of `EnumEntries<E>` being a `List<E>` instead of a `Set<E>`? As an immutable set, it could reuse the same strategies as Java's `EnumMap` for a fast `contains()` implementation.

gildor:
@CLOVIS-AI Set is an unordered collection, but the order is very important for Enum values in many cases, also you cannot access values by index (ordinal) in set, which is one more important case for collection of enum values

elizarov:
> @CLOVIS-AI What are the benefits of `EnumEntries<E>` being a `List<E>` instead of a `Set<E>`? As an immutable set, it could reuse the same strategies as Java's `EnumMap` for a fast `contains()` implementation.

To add to the answer by @[gildor](https://github.com/gildor), having `EnumEntries<E>` as `List<E>` does not prevent us from reusing the strategy for a fast `contains()` implementation, but it also makes possible for a fast retrieval of an entry by its ordinal (index) using `get(Int)` method which a `Set` does not have.

proposals/enum-entries.md
```diff
@@ -0,0 +1,174 @@
...
+* `values` will be de-prioritized and eventually removed from IDE auto-completion.
+* Soft warning with an intention to replace call to `values()` with call to `entries` will be introduced.
+* All the corresponding materials, such as Kotlin guides, J2K and tutorials will be adjusted to use `entries` API.
+* Eventually, we are going to re-visit this decision and decide on the further deprecation of `values` API.
+
+### `EnumEntries<E>` type
+
+Effectively, `entries` represents an immutable list of enum entries and can be represented as type `List<E>`.
+To have an ability to further extend Enum's API in a non-intrusive manner that does not involve code-generation
+for each enum, we expose a direct subtype of `List<E>` named `EnumEntries`.
```
ilya-g, line 65: Makes sense to quote its full signature `sealed interface EnumEntries<E : Enum<E>> : List<E>` and explain why it's sealed.

proposals/enum-entries.md
```diff
@@ -0,0 +1,174 @@
...
+```java
+enum MyEnum extends Enum<MyEnum> {
+    private static final synthetic MyEnum[] $VALUES
+    private static final synthetic EnumEntries<MyEnum> $ENTRIES;
+   
+    <clinit> {
+        A = new MyEnum("A", 0);
+        $VALUES = $values();
+        Supplier<MyEnum[]> supplier = #invokedynamic ..args.. values;
+        $ENTRIES = new EnumEntriesList(supplier);
```
ilya-g, line 117: Now it's a call to the `enumEntries` function instead of the constructor of `EnumEntries`

proposals/enum-entries.md
```diff
@@ -0,0 +1,174 @@
...
+
+For already compiled Kotlin enums and Java enums, a separate synthetic mapping classfile will be introduced 
+for each callsite of `Enum.entries`. The mapping class acts as a storage for a lazily-initialized and pre-allocated list of
+enum entries.
+
+## Risks and assumptions
+
+The proposal has two main risks:
+
+The potential source-compatibility issue for `entries` property in the companion objects within a Enum.
```
ilya-g, line 144:
```suggestion
The first is the potential source-compatibility issue for `entries` property in the companion object within an Enum.
```

proposals/enum-entries.md
```diff
@@ -0,0 +1,174 @@
...
+    companion object {
+        val entries: Any? = ...
+    }
+}
+
+MyEnum.entries // <- member has a higher priority than a companion member
+```
+
+To mitigate that, a separate compiler-assistant deprecation cycle is introduced, along with the preservation of 
+the resolve to `entries` from companion object.
```
ilya-g, line 160: Quite unclear what is "preservation of the resolve".

proposals/enum-entries.md
```diff
@@ -0,0 +1,174 @@
...
+Taking into account the additional deprecation cycle and non-trivial setup for the problem to reproduce, we do not 
+consider it a serious threat to compatibility.
+
+The second risk is the education disturbance and a new name for developers to get familiar with -- `entries`, opposed to already well-known `values`.
```
ilya-g, line 164:
```suggestion
The second risk is the education disturbance and a new name for developers to get familiar with — `entries`, opposed to already well-known `values`.
```

ilya-g: Or " ... familiar with, namely `entries`, opposed to ..."

ilya-g (commented):
Overall good text. 
Also, since it proposes to introduce new API in a new package, it may make sense to explain that choice: why new package, and why it is named like that.

proposals/enum-entries.md
```diff
@@ -0,0 +1,174 @@
...
+* All the corresponding materials, such as Kotlin guides, J2K and tutorials will be adjusted to use `entries` API.
+* Eventually, we are going to re-visit this decision and decide on the further deprecation of `values` API.
+
+### `EnumEntries<E>` type
+
+Effectively, `entries` represents an immutable list of enum entries and can be represented as type `List<E>`.
+To have an ability to further extend Enum's API in a non-intrusive manner that does not involve code-generation
+for each enum, we expose a direct subtype of `List<E>` named `EnumEntries`.
+
+All potential extensions, such as `valueOfOrNull(String)`, `hasValue(String)` can be implemented on the standard library
```
luanpotter, line 67: Thanks for this great proposal! Could we consider adding `valueOfOrNull(String)` as part/collateral of this instead of a followup? It would be great to have a standard centralized function from the start on the SDK to avoid each project having to define its own extension function and then at some point later having to replace those.
Rationale: https://youtrack.jetbrains.com/issue/KT-42619/Add-a-static-valueOfOrNull-method-to-Enums-similar-to-valueOf

qwwdfsad: It will definitely not be part of the initial 1.8.0 because we do not know a proper solution for this problem, but I consider this a natural continuation of this KEEP.

We have two options:

* Invest more into static interfaces/static extensions (https://youtrack.jetbrains.com/issue/KT-11968/) and then introduce a bunch of convenience extensions on all enums, such as `valueOfOrNull` etc.
* Introduce extensions/members on the proposed `EnumEntries` interface. It is a piece of cake and can be added at any moment after this KEEP, but it has a downside of being a bit too verbose: instead of `MyEnum.valueOfOrNull` one has to write `MyEnum.entries.valueOfOrNull`. Such a solution cures the symptoms, not the disease (inability to extend enums statically), or at least can be considered as such.

qwwdfsad: (The option "generate valueOfOrNull right into enum body like values/entries" is out of the question though as it bloats the code and is neither extensible nor future-proof for such a simple task)

fvasco: It may be possible to get an `Enum` value by index or by name, defining: `operator EnumEntries.get(name: String) : E?'

```kotlin
MyEnum.entries[index]
MyEnum.entries[name]
```

elizarov (approved):

proposals/enum-entries.md
qwwdfsad: This is an interesting suggestion, written it down in the corresponding KT issue

elizarov (approved):

proposals/enum-entries.md
luanpotter: @qwwdfsad thanks for the explanation and rationale :)
@fvasco I really like that suggestion as well!

PR #286, opened 2022-01-04
Fix package name that contains `Continuation` def

ZheyuJin:
`kotlinx.coroutines` -> `kotlin.coroutines`


elizarov (approved):
Great catch! Thanks.

PR #289, opened 2022-01-24
Updated explicit backing fields proposal

elizarov:
Updates to #277:

* Use-cases are orthogonalized (similar) use-cases are grouped together.
* Additional use-cases are explained to cover all the major pieces of the proposed design.
* Design is worked out in more detail, with grammar and various restrictions more explicitly spelled out.
* This list of future enhancements is narrowed down to the ones that are feasible in the near future.
* Table of Contents is added for reference.

### DIFF
--- a/proposals/explicit-backing-fields.md
+++ b/proposals/explicit-backing-fields.md
@@ -1,0 +1,273 @@
+# Explicit Backing Fields
+
+- **Type**: Design Proposal
+- **Author**: ?
+- **Contributors**: Svetlana Isakova, Kirill Rakhman, Dmitry Petrov, Roman Elizarov, Ben Leggiero, Matej Drobnič, Mikhail Glukhikh, Nikolay Lunyak
+- **Status**: Implemented in FIR
+- **Prototype**: Implemented
+- **Initial YouTrack Issue**: [KT-14663](https://youtrack.jetbrains.com/issue/KT-14663)
+- **Initial Proposal**: [private_public_property_types#122](https://github.com/Kotlin/KEEP/pull/122)
+
+## Summary
+
+**Note**: initial proposal contents have been partially copied down here for convenience. Despite the different approach, the already shown use cases are still relevant.
+
+>  Common pattern in java is having full type accessible as private property and then only exposing required interface in the public getter:
+>
+> ```java
+> class MyClass {
+> 	// Use full type for private access
+> 	private final ArrayList<String> data = new ArrayList<>();
+> 	
+> 	// Only expose what is needed in public getter
+> 	public List<String> getData() {
+> 		return data;
+> 	}
+> }
+> ```
+>
+> This pattern allows easy hiding of implementation details and allows for easy clean external interfaces to the classes. But there is no idiomatic Kotlin way to achieve this pattern. Best thing you can do is define two separate properties to mimic how Java does it:
+>
+> ```kotlin
+> class MyClass {
+>     private val _data = ArrayList<String>()
+>     val data: List<String>
+>         get() = _data
+> }
+> ```
+
+With the proposed syntax in mind, the above code snipped could be rewritten as follows:
+
+```kotlin
+class MyClass {
+    val data: List<String>
+        field = ArrayList()
+}
+```
+
+## Use Cases
+
+### Motivation
+
+> - There is no clean idiomatic way to do this pattern in Kotlin
+> - Current best approach throws away all benefits of Kotlin properties and forces developer to write Java-like code with separate private field and public getter
+> - Current best approach forces developer to assign two different names to single property (or pad private property, for example adding `_` prefix and then using this prefix everywhere in code)
+> - This is another place where Java pattern could be introduced into Kotlin with less boilerplate
+
+### Read-Only from Outside
+
+> We often do not want our data structures to be modified from outside. Unlike Java, this can be easily achieved in Kotlin by just exposing read-only `List`. But as already ilustrated in above example, exposing different type of a property is a bit messy.
+
+```kotlin
+private val _items = mutableListOf<Item>()
+val item : List<Item> by _items
+```
+
+And the new syntax allows us to write:
+
+```kotlin
+val items: List<Item>
+	private field = mutableListOf()
+```
+
+### Android Architecture Components
+
+>  Proper way to do Architecture components is to use `MutableLiveData` (`LiveData` implementation that allows caller to change its value) privately inside View Model classes and then only expose read-only `LiveData` objects outside.
+
+Sample code [from an Android app](https://github.com/elpassion/crweather/blob/9c3e3cb803b7e4fffbb010ff085ac56645c9774d/app/src/main/java/com/elpassion/crweather/MainModel.kt#L14):
+
+```kotlin
+private val mutableCity = MutableLiveData<String>().apply { value = "" }
+private val mutableCharts = MutableLiveData<List<Chart>>().apply { value = emptyList() }
+private val mutableLoading = MutableLiveData<Boolean>().apply { value = false }
+private val mutableMessage = MutableLiveData<String>()
+
+val city: LiveData<String> = mutableCity
+val charts: LiveData<List<Chart>> = mutableCharts
+val loading: LiveData<Boolean> = mutableLoading
+val message: LiveData<String> = mutableMessage
+```
+
+Becomes:
+
+```kotlin
+val city: LiveData<String>
+	field = MutableLiveData().apply { value = "" }
+val charts: LiveData<List<Chart>>
+	field = MutableLiveData().apply { value = emptyList() }
+val loading: LiveData<Boolean>
+	field = MutableLiveData().apply { value = false }
+val message: LiveData<String>
+	field = MutableLiveData()
+```
+
+### RX Observable and Subjects
+
+> Data can often be pushed into reactive world using subjects. However, exposing `Subject` would allow consumer of the class to push its own data into it. That is why it is good idea to expose all subjects as read-only `Observable`:
+
+```kotlin
+class MyClass {
+    private val _dataStream = PublishSubject.create<String>()
+    val dataStream: Observable<String>
+        get() = _dataStream
+}
+```
+
+Turns into:
+
+```kotlin
+class MyClass {
+    val dataStream: Observable<String>
+    	field = PublishSubject.create()
+}
+```
+
+## Design
+
+The proposed design consists of two new ideas.
+
+### Explicit Backing Fields
+
+```kotlin
+val it: A
+	[visibility] field[: B] = initializer
+```
+
+The above `field` declaration is referred to as an _explicit backing field declaration_.
+
+#### Accessors
+
+- if `A :> B`, the compiler can provide a default getter
+- if `A <: B`, the compiler can provide a default setter
+- If an accessor is required, it must be provided either by the compiler or the user
+
+```kotlin
+public val flow: SharedFlow
+    field: MutableSharedFlow? = null
+    get() {
+        return field ?: run { ... }
+    }
+```
+
+#### Visibility
+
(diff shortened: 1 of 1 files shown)


proposals/explicit-backing-fields.md
```diff
@@ -80,100 +109,155 @@ val loading: LiveData<Boolean>
...
+This proposal is designed with an idea to provide an explicit syntax to access a property's field from anywhere
+inside the corresponding class when the field is declared explicitly:
 
 ```kotlin
-class MyClass {
-    val dataStream: Observable<String>
-        field = PublishSubject.create()
+class Component {
+    var status: Status
+        field // explicit field declaration
```
mglukhikh, line 159: This line is redundant as it is

mglukhikh: Also, there is no initializer here

elizarov: Let's discuss this part of design here. Do we automatically make backing fields of _all properties_ available inside the class via TBD syntax or only if you explicitly declare a backing field for them?

mglukhikh: I think this is better

> we automatically make backing fields of all properties available inside the class via TBD syntax

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,355 @@
...
+        set(value) {
+            field = value
+            notifyStatusChanged()
+        }
+}
+```
+
+This way, all the code inside the class can change the field of the property directly, without invoking the setter .
```
ilya-g, line 167:
```suggestion
This way, all the code inside the class can change the field of the property directly, without invoking the setter. 
```

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,355 @@
...
+In some application it is desirable to expose not just a read-only subtype, but a specially constructed read-only
+view that protects the data structure from casting into mutable type. For example a `MutableStateFlow` has
+[`asStateFlow`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/as-state-flow.html)
+extension for that purpose.
+
+With the proposed syntax it is possible to declare a custom getter:
+
+```kotlin
+val state: StateFlow<State> 
+    get() = field.asStateFlow()
```
ilya-g, line 144: This creates a view on each getter invocation which may be unwanted.

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,355 @@
...
+```kotlin
+val city: LiveData<String>
+    field = MutableLiveData().apply { value = "" }
+val charts: LiveData<List<Chart>>
+    field = MutableLiveData().apply { value = emptyList() }
+val loading: LiveData<Boolean>
+    field = MutableLiveData().apply { value = false }
+val message: LiveData<String>
+    field = MutableLiveData()
+```
```
ilya-g, line 111: It can be noted that while the total number of lines stays roughly the same, the related properties and fields are better grouped together, and also the repetition in property names is avoided.

proposals/explicit-backing-fields.md
elizarov: Ok. Corrected.

proposals/explicit-backing-fields.md
elizarov: Added clarification on this point.

proposals/explicit-backing-fields.md
elizarov: Added a note to that effect.

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,355 @@
...
+  * [Protected Fields](#protected-fields)
+  * [Mutable Fields for Read-only Properties](#mutable-fields-for-read-only-properties)
+
+<!--- END -->
+
+## Use Cases
+
+This proposal caters to a variety of use-cases that are currently met via a backing property pattern.
+
+### Expose read-only subtype
```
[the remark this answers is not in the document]
elizarov: Thanks. Fixed.

proposals/explicit-backing-fields.md
```diff
@@ -14,417 +14,396 @@ but one is part of a public API and another is an implementation detail.
...
 
-For this use-case to become actually usable, the TBD syntax of [Direct Backing Field Access](#direct-backing-field-access) 
-will need to be added. That's because such applications need to modify the mutable data structure that is stored in the field,
-but the rules of [Smart Type Narrowing](#smart-type-narrowing) will not make it directly available. 
+#### Access backing property inside
 
-### Access field from outside of getter and setter
+It is worth emphasizing that in most cases the primary property is only for public use and is not
+referenced within the class. Obviously, there is no point in doing so, since the backing property usually provides
+all the same functionality, as a primary property and even more.
```
lunakoly, line 143: Please don't use the word "obviously" in such a context. This is not a conclusion within a set of strict constraints, but the rationale behind a free choice we decide to make

proposals/explicit-backing-fields.md
```diff
@@ -14,417 +14,396 @@ but one is part of a public API and another is an implementation detail.
...
-Explicit backing field declaration has an optional visibility, an optional type, and an optional 
-initialization expression. 
+### Visibility
 
-#### Restrictions
+`private` is the default and the only allowed visibility for explicit backing fields
+(supporting other visibilities might be considered, see [Future enhancements](#future-enhancements)).
 
-There are the following additional semantic restrictions on the property declaration grammar:
+Visibility of property can be any, except `private`.
```
lunakoly, line 223: Minor: let's say "must be more permissive than", because this statement would stay correct w.r.t. possible future enhancements

proposals/explicit-backing-fields.md
```diff
@@ -14,417 +14,396 @@ but one is part of a public API and another is an implementation detail.
...
+    field = MutableStateFlow("")
 ```
 
-The special syntax to explicitly access the backing field from outside the code of property accessors is TBD, 
-but the field can be implicitly access when it is visible via the [Smart Type Narrowing](#smart-type-narrowing).
+In this case, the property initializer expression (`field.asStateFlow()` in example above) is evaluated once
+immediately after the backing field is initialized and then returned every time the getter of property is called.
 
-#### Lateinit
+If property with explicit backing field has initializer, it's prohibited to declare accessors.
```
lunakoly, line 286: One more candidate for the future enhancements list? Or a way "to prevent feature abuse"?

proposals/explicit-backing-fields.md
```diff
@@ -14,417 +14,396 @@ but one is part of a public API and another is an implementation detail.
...
-The implementation is not stable yet and will generate pre-release binary. 
-There is no IDE support yet.
+1. Make it possible to access getter instead of explicit backing field when referenced within the scope of visibility of explicit backing field. This is not supported because usually there is no need for calling getter (see [Access backing property inside](#access-backing-property-inside)).
+2. Support other visibilities of explicit backing field (`protected` or `internal`).
+3. Support delegation of explicit backing field or property.
+4. Support `lateinit` explicit backing field.
+5. Support combining mutable explicit backing field and non-mutable property. 
+Despite its popularity (see [`var` backing property](#var-backing-property)), this use case is not supported in this proposal, because it is impossible for one property to behave as mutable + nullable and
+at the same time non-mutable and non-nullable.
+6. Add API in Reflection to retrieve information about explicit backing field.
```
lunakoly, line 422: Since this KEEP also suggests bringing in the `init` field, it can probably raise a similar concern?

proposals/explicit-backing-fields.md
```diff
@@ -14,417 +14,396 @@ but one is part of a public API and another is an implementation detail.
...
-Consider the following snippet from the 
-[implementation of `lazy` in the Kotlin standard library](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/src/kotlin/util/LazyJVM.kt#L57):
+`@JvmField` is prohibited on properties with explicit backing field, because in this case a field type becomes exposed,
+which makes use of the feature completely pointless.
+
+#### Property with initializer compilation
+
+Property with explicit backing field and initializer is represented by additional
+auxiliary synthetic field which is needed for storing the value from property initializer.
+Auxiliary field is given the name `propertyName$init` and is `private` regardless of property visibility.
```
lunakoly, line 381: It may be worth mentioning if allowing everything (accessors + access to both the fields as `field` and `init` within them) is considered a future enhancement or something we'd like to avoid for design reasons

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,424 @@
...
+
+_Statistics_: backing property delegation found in 2693 files, primary property delegation in 1010 files.
+
+## Design
+
+### Explicit Backing Fields
+
+By analogy with the already existing term ["implicit backing fields"](https://kotlinlang.org/docs/properties.html#backing-properties), it is proposed to introduce a syntax
+for declaring "explicit backing fields" of properties.
+Syntax is consisted of a keyword `field`, optional type definition and initialization expression
```
ice-phoenix, line 204:
```suggestion
Syntax consists of a keyword `field`, optional type definition and initialization expression
```

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,424 @@
...
+    vm.city // visible as LiveData, calling getter
+} 
+```
+
+There is no possibility to call a getter inside the class
+(might be reconsidered in [Future enhancements](#future-enhancements)).
+
+### Accessors
+
+Property, which has explicit backing field, can also specify getter or setter.
```
ice-phoenix, line 252:
```suggestion
Property, which has explicit backing field, can also specify getter and/or setter.
```

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,424 @@
...
+
+### Other restrictions
+
+There are the following additional semantic restrictions on the property declaration.
+Property with explicit backing field:
+
+* must be final (otherwise calling a field instead of a getter would be undesirable behavior)
+* can't be `const`
+* can't be `expect` or `external`
+* can't be `inline` (but its accessors can)
```
ice-phoenix, line 321: FTR: an `inline` property means its accessors are `inline`.

Why do we want to have `inline` accessors for EBF properties? This means that the EBF leaks into the public ABI, has to be accessible from the outside and I'm not sure I like this.

Atm, `inline` properties cannot have backing fields, and I do not understand why we want to relax the restriction here.

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,424 @@
...
+* or explicit backing field itself can't be `lateinit` (added to [Future enhancements](#future-enhancements)).
```
ice-phoenix, line 322: Why "or"?

proposals/explicit-backing-fields.md
ice-phoenix: Smth I originally forgot -- do we say anywhere that EBF are prohibited for extension properties? =)

proposals/explicit-backing-fields.md
```diff
@@ -0,0 +1,433 @@
...
+    field = MutableLiveData<String>()
+    get() = field.value ?: "-" // It is an error if getter is not explicitly specified here
+```
+
+### Explicit backing field in combination with property initializer
+
+It's allowed to have both property initializer and explicit backing field specified.
+
+```kotlin
+val city: StateFlow<String> = field.asStateFlow()
```
gildor, line 286: What about cases this case:
```kotlin
val city: StateFlow<String> get() = field.asStateFlow()
    field = MutableStateFlow("")
```

Will it be supported? Now it's not really clear for me from the KEEP text

gildor: Ah, indeed. Thanks

merfemor:
Reopened as https://github.com/Kotlin/KEEP/pull/428

PR #293, opened 2022-03-28
Fix a small typo in Coroutines proposal

philipphofmann:
The changed sentence doesn't make sense. I think removing the `coroutine` fixes it.


philipphofmann:
Thanks for your quick reply. So for me, the missing article `the` was a bit confusing. I'm going to change it to your suggestion. Thanks, @elizarov.

elizarov:
Thanks!

PR #295, opened 2022-03-30
Update definitely-non-nullable-types.md

antonarhipov:
Corrected the example snippet: interface abstract methods cannot have a body.


elizarov:
Thanks!

PR #308, opened 2022-06-29
A proposal for the assign operator overload

asodja:
This is a proposal on providing an assign operator overload that would provide DSL for assigning values to mutable container objects.

Please, use [KEEP-309](https://github.com/Kotlin/KEEP/issues/309) issue for discussion of this proposal and post text and editorial changes into this PR.

### DIFF
--- a/proposals/assign-operator-overload.md
+++ b/proposals/assign-operator-overload.md
@@ -1,0 +1,441 @@
+# Assign operator overload
+
+* **Type**: Design proposal
+* **Authors**: Anže Sodja, Lóránt Pintér, Stefan Wolf, Roman Elizarov
+* **Status**: In review
+* **Discussion**: // TODO add KEEP issue
+* **Prototype**: Implemented
+
+
+## Abstract
+
+This KEEP is a proposal on providing an assign operator overload that would provide DSL for assigning values to mutable container objects.
+
+
+## Background
+
+Kotlin has a rich set of operators that make working with various containers in the standard library easier. For example, you can write the following code with lists:
+
+```kotlin
+val list = mutableListOf<Int>()
+list += 1 // operator plusAssign (1)
+list += listOf(2, 3) // operator plusAssign (2)
+list[0] = 4 // operator set
+```
+
+Notice that the `list` is defined as a read-only (`val`) property, but contains a reference to a mutable container of type `MutableList` which is modified with various operators using variants of assign (`=`) operator.
+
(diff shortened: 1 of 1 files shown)


proposals/assign-operator-overload.md
```diff
@@ -0,0 +1,441 @@
...
+myTask.input2 = calculatedProperty
+myTask.output = File("input-file.txt")
+```
+
+Or as a DSL
+
+```kotlin
+myTask.apply {
+    input1 = File("input-file.txt")
+    Input2 = calculatedProperty
```
alllex, line 126: `Input2` is accidentally capitalized here.
There are also a couple more places in the document where this typo has leaked.

JakeWharton:
Also, in general, opinions on the proposal should go in the associated issue (which hopefully is coming) rather than on the PR where its comments should be reserved for improving only the language / content / syntax of the proposal.

proposals/assign-operator-overload.md
asodja: Thank you, should be fixed now.

proposals/assign-operator-overload.md
```diff
@@ -0,0 +1,440 @@
+# Assign operator overload
+
+* **Type**: Design proposal
+* **Authors**: Anže Sodja, Lóránt Pintér, Stefan Wolf, Roman Elizarov
+* **Status**: In review
+* **Discussion**: // TODO add KEEP issue
```
elizarov, line 6: Please, link to https://github.com/Kotlin/KEEP/issues/309 here.

proposals/assign-operator-overload.md
```diff
@@ -0,0 +1,440 @@
...
+```kotlin
+val list = mutableListOf<Int>()
+list += 1 // operator plusAssign (1)
+list += listOf(2, 3) // operator plusAssign (2)
+list[0] = 4 // operator set
+```
+
+Notice that the `list` is defined as a read-only (`val`) property, but contains a reference to a mutable container of type `MutableList` which is modified with various operators using variants of assign (`=`) operator.
+
+These operators are not hardcoded in Kotlin to operate on lists of arrays, for that matter, but are implemented in the language via _operator conventions_. The language is designed to recognize those conventions and resolve the corresponding operators (in the above example we use operators `plusAssign` and `set`), while the actual support for the specific containers is provided in the standard library. Other libraries are free to define those operators on their own containers to suit the intended usage of those containers.
```
elizarov, line 28: `lists of arrays` -> `lists on arrays`

asodja: Fixed

proposals/assign-operator-overload.md
asodja: Done

elizarov:
Since we've rejected the idea of adding the corresponding convention, as described in the KEEP, to the language directly, but are providing an extension point that allows the similar implementation as a compiler plugin to be used in Gradle, I'm closing this PR and issue.

The documentation for the feature shall change (as only a specific Gradle DSL behavior needs to be documented) and shall be hosted elsewhere. The documentation of the extension points itself will be performed in the future during compiler API stabilization.

PR #312, opened 2022-07-20
Add KEEP for sealed inline classes

ilmirus:


proposals/sealed-inline-classes.md
```diff
@@ -0,0 +1,587 @@
...
+Currently, there is no type-safe way to create an inline class, which can hold several types, similar to union type. 
+The only way to use `Any?` as underlying type, which leads to a lot of `is` checks and cast the underlying value to expected type.
+
+We propose to support sealed inline classes, which are inline classes, which can hold several distinct types. 
+
+Sealed inline classes are sealed classes as well, which allows a programmer to distinguish the types, which the inline class can hold.
+
+## Motivation / use cases
+
+The main use-case for sealed inline classes is `Result`. Currently it is decraled as inline class with underlying type `Any?`
```
fvasco, line 21: typo: decraled

proposals/sealed-inline-classes.md
```diff
@@ -0,0 +1,587 @@
...
+
+If we do not box `O` to `I`, when we map to `Any?` (remember, sealed inline classes are mapped to `Any?`), it will appear, that `foo` returns `O` instead of `IC`.
+
+In case of `Result`, that will mean, that there is no way to represent `Success(Failure)`.
+
+Thus, we box noinline children with sealed inline class type, when the sealed inline class type should be boxed as well. In other words, when inline class with underlying type `Any?` should be boxed.
+
+This is the reason for the following restrictions.
+
+- Noinline children (classes and objects) should have `value` modifier, as shown in the examples. Since the noinline children can be boxed, they cannot have stable identity. In other words, they are identitiless, and we use `value` modifier to mark them.
```
fvasco, line 138: typo: identitiless

proposals/sealed-inline-classes.md
```diff
@@ -0,0 +1,587 @@
...
+```
+
+becomes:
+```kotlin
+fun foo-<hash>(ic: Any?)
+```
+
+`<hash>` is computed using usual [inline classes mangling rules](https://github.com/Kotlin/KEEP/blob/master/proposals/inline-classes.md#mangling-rules).
+
+- When we pass nullable sealed inline class, it is mapped to boxed selaed inline class:
```
fvasco, line 217: typo: selaed

proposals/sealed-inline-classes.md
```diff
@@ -0,0 +1,587 @@
...
+
+value object ICObject: IC()
+
+val ic: IC = ICInt(1)
+```
+
+the compiler generates the following code
+
+```kotlin
+val ic = Integet.valueOf(1)
```
fvasco, line 200: typo: Integet

proposals/sealed-inline-classes.md
```diff
@@ -0,0 +1,587 @@
...
+```
+
+The following rulues for passing to function and returning from the function apply:
```
fvasco, line 203: typo: rulues

proposals/sealed-inline-classes.md
```diff
@@ -0,0 +1,587 @@
...
+}
+
+// In I2
+// Nothing
+
+// In I3
+fun `toString-impl`(value: Any?): String = "I3${`$value`}"
+```
+
+The logic for `toString` applies for all methods, decrared in interfaces: if the method is not overridden in inline child, we do not generate redirect.
```
fvasco, line 364: typo: decrared

proposals/sealed-inline-classes.md
```diff
@@ -0,0 +1,587 @@
...
+
+value object O2: I2()
+```
+
+Here, we need to generate synthetic method in `I1`, since we only box to `I1`
+and we need to get to `I2`.
+
+```kotlin
+// In I1
+synthetic fun `str-impl`(value: Any?): String = I2.`str-impl`(value)
```
fvasco, line 544: `synthetic` has been used inconsistently in the examples.
See the following lines or the definition of the `$value` property.

proposals/sealed-inline-classes.md
```diff
@@ -0,0 +1,587 @@
...
+```kotlin
+@JvmInline
+sealed value class IC
+```
+
+becomes:
+
+```kotlin
+class IC {
+  val $value: Any?
```
fvasco, line 178: Is this a Kotlin property?

We should expect on JVM?

```java
private Object $value
public Object get$value()
```

proposals/sealed-inline-classes.md
```diff
@@ -0,0 +1,593 @@
...
+@JvmInline
+sealed value class IC {
+  @JvmInline
+  value class ICInt(val i: Int): IC()
+  @JvmInline
+  value class ICLong(val l: Long): IC()
+}
+```
+
+here, we can use `Number` as the underlying type. However, adding non-`Number` child will change the underlying type,
```
elizarov, line 110: It is better to replace "we can use Number" with "we could have used Number" to let a reader understand what's the actual design we are going with and what is just a digression on other potential alternatives we have considered in the design.

proposals/sealed-inline-classes.md
```diff
@@ -0,0 +1,593 @@
+# Sealed Inline Classes
+* **Type**: Design proposal
+* **Author**: Ilmir Usmanov
+* **Contributors**: Roman Elizarov, Alexander Udalov
+* **Status**: Experimental
+* **Prototype**: TODO
+
+Discussion of this proposal is held in [this issue](TODO).
```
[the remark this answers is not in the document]
udalov: #318

proposals/sealed-inline-classes.md
```diff
@@ -0,0 +1,593 @@
...
+fun foo(icObject: ICObject)
+``` 
+
+becomes:
+```kotlin
+fun foo(icObject: ICObject)
+```
+
+## Is checks
+Consider the following example
```
udalov, line 266: This example does not compile because `sealed inline class child cannot be distinguished from another one`

proposals/sealed-inline-classes.md
```diff
@@ -0,0 +1,593 @@
...
+  - Final class and class or interface: the final class cannot extend or implement the other type.
+  - Open class and open class: they cannot be subtypes of one another.
+  - Open class and interface: forbidden - one can extend the class and implement the interface at the same time, so there will be no way to distinguish them.
+  - Interface and interface: ditto.
+
+- Sealed inline classes cannot have primary constructors.
+
+- If sealed inline class is a child of another sealed inline class, there can be no other inline children, including other sealed inline class, since sealed inline classes are mapped to `Any?`, which is open class, which other classes override and there is no way to distinguish `Any?` from other type.
+
+- Value objects must be a child of sealed inline class.
```
udalov, line 156: The proposal doesn't explain what a "value object" is and when it's allowed

proposals/sealed-inline-classes.md
```diff
@@ -0,0 +1,593 @@
...
+@JvmInline
+value class I3(val s: String): I2() {
+    override fun toString(): String = "I3"
+}
+
+object O2: I2()
+```
+
+Note, that `O2` does not override `toString`, so we should call `Any.toString()`
+is case of `O2`, since `I2` does not override `toString` as well.
```
grechkovlad, line 350: _i**s** case of_ -  typo?

proposals/sealed-inline-classes.md
```diff
@@ -0,0 +1,593 @@
...
+```kotlin
+@JvmInline
+sealed value class I1 {
+    open fun str(): String = "I1"
+}
+
+@JvmInline
+sealed value class I2 : I1()
+
+sealed object O1: I1() {
```
grechkovlad, line 387: Shouldn't be `value object` instead of `sealed object` here?

MairwunNx:
The proposed design solves the `Result` case (and it's probably _the only_ case) but there are some caveats:
* amongst classes with the same underlying type, only one can be inline,
* primitive wrappers or any control over the underlying value type are absolutely destroyed.

Consider the following example with different design:
```kt
@JvmInline
sealed value class Color protected constructor(val value: Long) {
    @JvmInline
    value class Fixed(color: Int) : Color(color.toLong() and 0xFFFFFFFFL) {
        val color: Int get() = value.toInt()
        override fun toString(): String = "Color.Fixed(#%08X)".format(color)
    }

    @JvmInline
    value class Resource(@ColorRes colorRes: Int) : Color((1L shl 32) or (colorRes.toLong() and 0xFFFFFFFFL)) {
        val resId: Int @ColorRes get() = value.toInt()
        override fun toString(): String = "Color.Resource(0x${resId.toString(16)})"
    }

    operator fun getClass(): KClass<out Color> = when (value ushr 32) {
        0L -> Fixed::class
        1L -> Resource::class
        else -> throw AssertionError()
    }
}
```
Probable desugaring:
```kt
@JvmInline
sealed value class Color {
    static fun constructor-impl(value: Long): Long = value

    @JvmInline
    value class Fixed {
        static fun constructor-impl(color: Int): Long {
            val expressionPassedToSuper = color.toLong() and 0xFFFFFFFFL
            val value = Color.constructor-impl(expressionPassedToSuper)
            Intrinsics.requireConsistentInlineCtor(Fixed::class, Color.getClass(value))
            return value
        }
        // val color ...
        // fun toString() ...
    }

    @JvmInline
    value class Resource { /* contents identical */ }

    // static fun getClass() ...

    // proxies for 'overridden' functions
    static fun toString-impl(value: Long) = when (getClass(value)) {
        Fixed::class -> Fixed.toString-impl(value)
        Resource::class -> Resource.toString-impl(value)
        else -> throw NoWhenBranchMatchedException()
    }
    static fun hashCode-impl(value: Long) { /* either `when (getClass()) {}` or default value.hashCode() */ }
    static fun equals-impl(value: Long, other: Long) { /* either `when (getClass()) {}` or default value.equals(other) */ }
}
```

Probable emulation on top of current Kotlin:
```kt
@JvmInline
value class Color(val value: Long) {
    @JvmInline
    value class Fixed @PublishedApi internal constructor(val value: Long) {
        constructor(color: Int) : this(color.toLong() and 0xFFFFFFFFL)
        val color: Int get() = value.toInt()
        override fun toString(): String = "Color.Fixed(#%08X)".format(color)
        fun upcast() = Color(value)
    }

    @JvmInline
    value class Resource @PublishedApi internal constructor(val value: Long) {
        constructor(@ColorRes colorRes: Int) : this((1L shl 32) or (colorRes.toLong() and 0xFFFFFFFFL))
        val resId: Int @ColorRes get() = value.toInt()
        override fun toString(): String = "Color.Resource(0x${resId.toString(16)})"
        fun upcast() = Color(value)
    }

    inline fun <R> switch(
        fixed: (Fixed) -> R,
        resource: (Resource) -> R,
    ): R = when (value ushr 32) {
        0L -> fixed(Fixed(value))
        1L -> resource(Resource(value))
        else -> throw AssertionError()
    }

    fun asFixed(): Fixed =
        if ((value ushr 32) == 0L) Fixed(value) else throw TypeCastException()

    fun safeAsFixed(): Fixed? =
        if ((value ushr 32) == 0L) Fixed(value) else null

    fun asResource(): Resource =
        if ((value ushr 32) == 1L) Resource(value) else throw TypeCastException()

    fun safeAsResource(): Resource? =
        if ((value ushr 32) == 1L) Resource(value) else null
}
```

ilmirus:
After a lot of discussions and several iterations to improve the design, we decided to the feature altogether, since the `Result` issue can be solved like everything else with `Result` - adding yet another hack to the compiler.

Issue #314, opened 2022-07-22
Open-ended ranges and rangeUntil operator


ilya-g:
This issue is for discussion of [the proposal](https://github.com/Kotlin/KEEP/blob/master/proposals/open-ended-ranges.md) to introduce `rangeUntil` (`..<`) operator in Kotlin and the corresponding types for open-ended ranges.

Proposal PR #315

Whathecode:
Thank you for this excellent overview of planned additions and tradeoffs considered!

I want to comment on the following:

> **Common supertype of ClosedRange and OpenEndRange**
> ...
> Introducing a more dedicated common supertype of the ClosedRange and OpenEndRange could help in this situation, however, it's unclear what useful operations such supertype would provide.

I'm the library author of `kotlinx.interval`, which approaches ranges using a common generic base type (including a potentially separate type for range 'sizes'). To help assess whether or not such a base type may be useful, I will soon continue porting some of the existing C# implementation I did to Kotlin, and add example use cases.

In short, I remember using them in a generic binary sort algorithm, and operations on range _collections_. I.e., given a collection of time intervals, does a specified timestamp fall within any of the ranges? But, likely those could have been simple collections of closed ranges only. I don't recall fully whether mixing half-open and closed ranges was ever a true requirement for the latter use case.

zarechenskiy:
Closing the issue as the feature was graduated to Stable in Kotlin 1.9.0

PR #315, opened 2022-07-22
Open-ended ranges and rangeUntil operator proposal

ilya-g:
Discussion goes in [KEEP-314](https://github.com/Kotlin/KEEP/issues/314)

### DIFF
--- a/proposals/open-ended-ranges.md
+++ b/proposals/open-ended-ranges.md
@@ -65,7 +65,7 @@ println(eqivalent)  // 1..1.9999999999999998
 - Swift: `...` end-inclusive range, `..<` end-exclusive, supports one-sided ranges 
 - Ruby: `..` end-inclusive range, `...` end-exclusive, supports one-sided ranges
 - Groovy: `..` end-inclusive range, `..<` end-exclusive, `<..` start-exclusive, `<..<` both bounds-exclusive range
-
+- Rust: `..=` end-inclusive range, `..` end-exclusive, supports one-sided ranges
 
 ### Libraries for representing ranges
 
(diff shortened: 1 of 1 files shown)


proposals/open-ended-ranges.md
```diff
@@ -0,0 +1,260 @@
...
+
+```kotlin
+class IntRange : IntProgression(...), ClosedRange<Int>, OpenEndRange<Int> {
+    override val start: Int
+    override val endInclusive: Int
+    override val endExclusive: Int
+}
+```
+
+There's a subtlety in implementing `endExlcusive` property in such ranges: usually it returns `endInclusive + 1`, but
```
JakeWharton, line 158:
```suggestion
There's a subtlety in implementing `endExclusive` property in such ranges: usually it returns `endInclusive + 1`, but 
```

ilya-g: Thanks for spotting

proposals/open-ended-ranges.md
```diff
@@ -0,0 +1,260 @@
...
+println(eqivalent)  // 1..1.9999999999999998
+```
+
+## Similar API review
+
+### Languages that distinguish end-inclusive and end-exclusive ranges
+
+- Swift: `...` end-inclusive range, `..<` end-exclusive, supports one-sided ranges 
+- Ruby: `..` end-inclusive range, `...` end-exclusive, supports one-sided ranges
+- Groovy: `..` end-inclusive range, `..<` end-exclusive, `<..` start-exclusive, `<..<` both bounds-exclusive range
```
elizarov, line 67: A Rust is also worth noting here to add a variety to the syntax that different languages use (Rust is an outlier in using `..` for an end-exclusive range).

proposals/open-ended-ranges.md
```diff
@@ -0,0 +1,260 @@
...
+```
+
+### Equality of open-ended ranges
+
+Similar to closed ranges, the `OpenEndRange` interface does not specify contract for `equals`/`hashCode` implementations,
+however its concrete implementations can do that. For example, an open-ended range of double values equal to another such range
+when bounds are respectively equal to each other, or to any empty range of doubles, when it is empty itself.
+
+Also, as a consequence of implementing both `OpenEndRange` and `ClosedRange` in concrete range implementations for standard discrete types,
+an open-ended range is equal to the closed range with the same `start` value and `endExlcusive` equal to `endInclusive + 1`:
```
hfhbd, line 198:
```suggestion
an open-ended range is equal to the closed range with the same `start` value and `endExclusive` equal to `endInclusive + 1`:
```

ilya-g: One was already fixed by Jake, now this is the last one! Thanks

proposals/open-ended-ranges.md
```diff
@@ -0,0 +1,260 @@
...
+adjacent ranges neither have a point where they overlap, nor a point between them that is not contained in these ranges.  
+
+Even sometimes when the value is already discrete, for example, when it is expressed as a `Double` number, 
+and it is possible to represent a half-open range with a closed one by adjusting one of its bounds, 
+in practice it is not convenient to work with such ranges:
+
+```kotlin
+val equivalent = 1.0..2.0.nextDown() // contains the same values as 1.0..<2.0 range
+println(eqivalent)  // 1..1.9999999999999998
+```
```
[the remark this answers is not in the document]
ilya-g: Yes, probably "discrete" is an unfortunate generalization term for the types like `Int`, `Long`, and `Char`. I've rewritten parts where it was used this way. 
`Double` is still a discrete number type, IMO. An example of non-discrete number would be a rational number represented by `BigInteger` numerator and denominator: no matter which two such numbers are given, we can always find another one between them.

PR #316, opened 2022-08-04
Add a proposal for "data objects"

strangepleasures:


fvasco:
> The @hash part is meaningless even for non-ADT objects, because object is a singleton

On JVM, a data object is a singleton in each `ClassLoader`, so multiple objects with the same FQN (and maybe the same code) can coexist in the same process.
The `hash` part is useful to detect different *instances* of the "same singleton".
This kind of issue occurs when multiple `ClassLoader`s are used, ie OSGi.
I note that only for completeness, but I don't think that the Kotlin language should address this issue.

> copy makes no sense because it’s a singleton. Or should it be generated and always return this?

the copy method should never return `this`

proposals/data-objects.md
```diff
@@ -0,0 +1,76 @@
+# Data objects
+
+* **Type**: Design proposal
+* **Authors**: Alexander Udalov, Roman Elizarov, Pavel Mikhailovskii
+* **Status**: Experimental since 1.7.20, stable since 1.8.0
+
```
elizarov, line 6: Please, also add a link to the issue for discussion of the proposal: https://github.com/Kotlin/KEEP/issues/317

proposals/data-objects.md
```diff
@@ -0,0 +1,76 @@
...
+* It leads to extra bytecode which is often unnecessary
+* Simple name instead of FQ name might be confusing, especially for cases like companion object
+
+### Data object
+
+Another possible solution is to allow the data modifier on objects, which will lead to toString being generated as for data classes.
+
+Other data class methods should not be generated:
+
+* equals and hashCode work fine already since it’s a singleton
```
elizarov, line 52: This part conflicts with the subsequent proposal without giving any rationale.

elizarov (changes requested):

proposals/data-objects.md
strangepleasures: > This part conflicts with the subsequent proposal without giving any rationale.

Thanks for reviewing. Please take another look.

proposals/data-objects.md
```diff
@@ -0,0 +1,84 @@
...
+* It leads to extra bytecode which is often unnecessary
+* Simple name instead of FQ name might be confusing, especially for cases like companion object
+
+### Data object
+
+Another possible solution is to allow the data modifier on objects, which will lead to `toString` being generated as for data classes.
+
+Even though data objects are singletons, `equals` and `hashCode` probably should also be generated:
+* it may be useful to have a stable `hashCode`
+* as it's in principle possible to create multiple instances of a data objec` class via JVM reflection, and some third-party frameworks may do that,
```
vanniktech, line 53: Should it be like this?

```suggestion
* as it's in principle possible to create multiple instances of a `data object` class via JVM reflection, and some third-party frameworks may do that,
```

proposals/data-objects.md
```diff
@@ -0,0 +1,84 @@
...
+A workaround for the user is either to declare `toString `manually, which leads to boilerplate, or use data class with a placeholder default argument, which seems bizarre.
+
+## Other aspects
+
+* Default toString implementation differs on JVM, JS, Native
+    * In JS, toString returns [object Object]
+    * Native behaves similarly to JVM, but inner classes are delimited with ., i.e. test.List*.*Nil@1d251891 in the example above
+* The @hash part is meaningless even for non-ADT objects, because object is a singleton
+* Can’t declare an empty @JvmRecord class in Kotlin (KT-48155 (https://youtrack.jetbrains.com/issue/KT-48155/JvmRecord-does-not-allow-one-to-define-no-parameters))
+    * Probably not a problem
```
nikitabobko, line 33: Let's add that Kotlin reflect `isData` method will now return `true` for `data object`? 

It might be useful for people to understand what consequences does annotating `object` with `data` bring

proposals/data-objects.md
```diff
@@ -0,0 +1,82 @@
...
+## The Proposed Solution
+
+* Support `data object` syntax in all frontends & backends
+* Generate equals, hashCode, toString for data objects
+  * `equals` should only check whether `this` and the other object are of the same type
+  * `hashCode` returns the value of the hash code of the qualified class name
+  * `toString` returns the simple name of the object
+* On JVM: if data object is serializable, generate `readResolve` which returns the value of the INSTANCE field
+* Prohibit to declare or inherit custom `equals`/`hashCode` (but not `toString`!) in data objects
+  * It’s OK if there’s a non-final implementation of `equals`/`hashCode` in a superclass of a data object, since it’s overridden anyway
```
nikitabobko, line 75: Why do we want to prohibit custom `equals`/`hashCode`? I don't see the motivation for this decision in the document. 

You may say that if later we decide that custom `equals`/`hashCode` might be useful then we can safely allow them. But that's not true because we'd break existing code:
```kotlin
data object Foo : Base(0)
data object Bar : Base(0)

abstract class Base(val x: Int) {
    override fun equals(other: Any?): Boolean = (other as? Base)?.x?.equals(x) == true
    override fun hashCode(): Int = x.hashCode()
}

fun main() {
    println(Foo.equals(Bar)) // not expected false
}
```

Right now, I only see downsides of prohibiting `equals`/`hashCode` because of the puzzlers like above. In `data class` we don't prohibit custom `equals`/`hashCode`, so the prohibition is somehow inconsistent too. What are the upsides?

nikitabobko: Oh, I'm wrong, `data object` doesn't bring a new puzzler in the language, `data class` allows to write exactly the same puzzler in existing Kotlin:
```kotlin
data class Foo(val z: Int) : Base()

abstract class Base {
    override fun equals(other: Any?): Boolean = false
    override fun hashCode(): Int = 0
}

fun main() {
    println(Foo(1) == Foo(1)) // true
}
```

And after thinking a bit, it's not a puzzler for me anymore, `data` is a synonymous for overriding `equals`. Now the ban to override `equals`/`hashCode` in `data object` seems reasonable to me, sorry for the noise

strangepleasures:
> The `hash` part is useful to detect different _instances_ of the "same singleton".

Wouldn't it be more natural to use the `===` operator for that?

Zhuinden:
```
    kotlinOptions {
        jvmTarget = '1.8'
        languageVersion = '1.9' // data objects
    }

kotlin.sourceSets.all {
    it.languageSettings.enableLanguageFeature("DataObjects")
}
```

PR #319, opened 2022-08-15
KEEP for SubclassOptInRequired

qwwdfsad:


proposals/subclass-opt-in-required.md
```diff
@@ -0,0 +1,131 @@
...
+@Target(CLASS)
+@Retention(BINARY)
+@SinceKotlin("1.8")
+@ExperimentalSubclassOptIn
+public annotation class SubclassOptInRequired(
+    val markerClass: KClass<out Annotation>
+)
+```
+
+The `SubclassOptInRequired` accepts only opt-in markers (annotations that are explicitly marked with `RequiresOptIn`) and can be
```
ilya-g, line 63: The name makes the impression that only deriving a class (subclass) from this type requires an opt-in, whereas it affects derived interfaces as well.

qwwdfsad: This name is explained in follow-up `Naming alternatives` section.
We failed to come up with a much better name

qwwdfsad: We are going to branch-out 1.8.0 soon and feature is available as experimental (guarded by opt-in), so if you have any objections or suggestions, it's time to raise the question to our DM

hrach: Did you consider subtype?

qwwdfsad: Subtyping is considered to be a use-site beast while subclassing is the declaration-site, and this feature is purely declaration-site based. Also, subtyping is a more overloaded term as it implies generics and variance binary relation

PR #327, opened 2022-10-31
Add a proposal for supporting non-local `break` and `continue` inside inline lambdas

strangepleasures:


strangepleasures:
Co-authored with @nikitabobko

udalov (approved):

proposals/break-continue-in-inline-lambdas.md
```diff
@@ -0,0 +1,175 @@
...
+        run {
+            if (i == 3) return@breaking
+        }
+    }
+}
+```
+
+## Potential issues with data-flow analysis
+
+We should check that the new feature doesn't introduce new data-flow analysis issues:
```
nikitabobko, line 171: I just realized that I don't like "action items" in our KEEP documents. IMHO, KEEP document should be a document that tries to convince that the feature is worth to be added. 

Action items in such kind of document only raise a question: "so what did someone do the action item"

So :) Did we check that non-local `break` and `continue` don't introduce data-flow analysis issues?

proposals/break-continue-in-inline-lambdas.md
```diff
@@ -0,0 +1,175 @@
...
+
+* **Q**: Why not make labels mandatory for non-local `break`/`continue`? That would eliminate `forEach` puzzler.
+* **A**: Because Kotlin doesn't require labels for non-local `return`. For sake of consistency, We should either require labels
+  for both `break`/`continue` and `return` cases or don't require the labels again for the both cases.
+
+* **Q**: Ok, why not make labels mandatory for both non-local `return` and non-local `break`/`continue` then?
+* **A**: We can't make non-local `return` to be obligatory labeled because it would break inline functions first-class
+  citizenship in language constructs. E.g. we value "return if not null" idiomatic Kotlin construct `nullable?.let { return it }`.
+  If non-local `return` were obligatory labeled then the code would be much more verbose
+  `nullable?.let { return@functionName it }`.
```
nikitabobko, line 104: I have a third question for which I don't have an answer:

* **Q**: Why not make labels mandatory for both non-local `return` and non-local `break`/`continue` but only for lambdas that don't have `callsInPlace(block, InvocationKind.EXACTLY_ONCE)` contract?

proposals/break-continue-in-inline-lambdas.md
```diff
@@ -0,0 +1,175 @@
...
+        list.forEach {
+            if (it == 0) break // Does it exit from while or from forEach?
+            println(it)
+        }
+    }
+}
+```
+
+To reduce the risk of confusion, it was proposed to introduce an IDE inspection and a quickfix recommending to use labeled `break`
+and `continue` in lambdas passed to parameters without an `EXACTLY_ONCE` or `AT_MOST_ONCE` contract (potentially "loop-like").
```
nikitabobko, line 127: > recommending to use labeled `break` and `continue` in lambdas passed to parameters without an `EXACTLY_ONCE` or `AT_MOST_ONCE` contract

Concern: it means that we will warn users about non-local break and continue for user-defined inline functions because usually, users don't define contracts for own functions. Contracts are still experimental and we can't afford requiring users to use OptIn annotations

nikitabobko: Right now I'm thinking that the inspection scope should be restricted to checking only Stdlib functions

proposals/break-continue-in-inline-lambdas.md
```diff
@@ -0,0 +1,175 @@
...
+* **Authors**: Alexander Udalov, Nikita Bobko, Pavel Mikhailovskii, Roman Elizarov, Simon Ogorodnik
+* **Status**: Experimental since 1.8.20, stable since 1.9.0
+* **Discussion and feedback**: [KEEP-326](https://github.com/Kotlin/KEEP/issues/326)
+* **YouTrack issues:** [KT-1436](https://youtrack.jetbrains.com/issue/KT-1436),
+  [KT-19748](https://youtrack.jetbrains.com/issue/KT-19748)
+
+## Introduction
+
+In Kotlin, it is possible to write [non-local returns](https://kotlinlang.org/docs/inline-functions.html#non-local-returns).
+Non-local `return` "jumps" over an inlined lambda. But the very same feature but for `continue` and `break` statements is
```
cortinico, line 13:
```suggestion
Non-local `return` "jumps" over an inlined lambda. But the very same feature for `continue` and `break` statements is
```

proposals/break-continue-in-inline-lambdas.md
strangepleasures: @nikitabobko We cannot make labels mandatory for non-local returns as that would break existing code. Instead, we have a simple rule: an unlabeled return within a lambda is always non-local, a labeled one, well, returns to the label.

proposals/break-continue-in-inline-lambdas.md
strangepleasures: @nikitabobko That's a valid concern. Not sure if limiting the inspection scope to Stdlib is a good idea. If we cannot make it work consistently, maybe we shouldn't implement it at all.

proposals/break-continue-in-inline-lambdas.md
```diff
@@ -0,0 +1,141 @@
...
+* **Authors**: Alexander Udalov, Nikita Bobko, Pavel Mikhailovskii, Roman Elizarov, Simon Ogorodnik
+* **Status**: Experimental since 1.8.20, stable since 1.9.0
+* **Discussion and feedback**: [KEEP-326](https://github.com/Kotlin/KEEP/issues/326)
+* **YouTrack issues:** [KT-1436](https://youtrack.jetbrains.com/issue/KT-1436),
+  [KT-19748](https://youtrack.jetbrains.com/issue/KT-19748)
+
+## Introduction
+
+One of the key features of Kotlin inline functions is that lambdas passed as arguments to them can contain
+[non-local return statements] (https://kotlinlang.org/docs/inline-functions.html#non-local-returns) which jump over
```
elizarov, line 13: Bad MD reference syntax (space in between `]` and `(`)

proposals/break-continue-in-inline-lambdas.md
```diff
@@ -0,0 +1,141 @@
+# Non-local break and continue
+
+* **Type**: Design proposal
+* **Authors**: Alexander Udalov, Nikita Bobko, Pavel Mikhailovskii, Roman Elizarov, Simon Ogorodnik
```
elizarov, line 4: The author is the person who had authored the text of the KEEP or otherwise wrote significant portions of it. Others are usually listed as `Contributors`.

proposals/break-continue-in-inline-lambdas.md
```diff
@@ -0,0 +1,141 @@
...
+Inline functions with [trailing lambdas](https://kotlinlang.org/docs/lambdas.html#passing-trailing-lambdas) are designed in a way
+that allows to use them as if they were first-class language constructs (e.g. `synchronized` or `run`, `with`, `let` scope
+functions from stdlib).
+
+Since users can invoke `return`, `break` or `continue` inside of bodies of native language constructs (such as body of `if` or
+body of `try-catch`), it would be only logical to allow the same for lambdas passed as parameters to inline functions. At the
+moment only non-local `return`s are supported. That's why we are proposing to add support for non-local `break`/`continue` as
+well.
+
+So the Motivation is the following:
```
elizarov, line 61: `Motivation` -> `motivation`

proposals/break-continue-in-inline-lambdas.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+However, it seems that both features can be implemented without introducing ambiguities.
+
+First of all, we should never allow unlabeled `break`/`continue` for "loop-like" functions because users
+wouldn't be able to understand whether it is applied to a certain function without checking its implementation.
+```kotlin
+foo { bar { break } } // which function is loop-like?
+```
+That leave us only with the labeled syntax option for "loop-like" functions:
```
elizarov, line 90: `That leave us` -> `It leaves us`

proposals/break-continue-in-inline-lambdas.md
```diff
@@ -0,0 +1,141 @@
...
+```kotlin
+for (i in 0 until 100) {
+    (0 until 200).forEach {
+        if (it == 42) break@forEach // Obviously, it breaks `forEach`
+        if (i == 10) continue // No label, so it applies to the 'for' loop
+    }
+}
+```
+
+A similar rule applies to `return` statements: an unlabeled return within a lambda is always non-local, a labeled one returns to the label.
```
elizarov, line 100: It is important to add here that an unlabeled `return` goes to an innermost enclosing block that is clearly marked with `fun` _hard_ keyword. In a similar way, an unlabeled `break`/`continue` go to an innermost enclosing block that is clearly marked with `for`, `do`, or `while` hard keywords.

proposals/break-continue-in-inline-lambdas.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Tooling for puzzling code
+
+Even we don't implement support for `break`/`continue` for "loop-like" function,
```
elizarov, line 104: `Even we` -> `Even as we`
`function,s` -> `functions,`

proposals/break-continue-in-inline-lambdas.md
elizarov: Tying the inspection to the contract is a good option. 

It will typically be used only with stdlib, and we must be sure it is not triggered on the usual scope functions. It is not bad if it is triggered on a user-defined function without a contract. It is better to be safe (by suggesting to use labelled break/continue) than end up with a more ambiguous code on a user-defined loop-like function.

proposals/break-continue-in-inline-lambdas.md
```diff
@@ -0,0 +1,141 @@
...
+        val list = producer()
+        list.forEach {
+            if (it == 0) break@myLoop
+            println(it)
+        }
+    }
+}
+```
+
+The inspection should not triggerred for `let`, `run`, etc as they all have the corresponding contract.
```
elizarov, line 133: `should not triggered` -> `should not be triggerred`

elizarov (changes requested):
General comments: 

1. Please, run the whole text though Grazie. There are tons of obvious grammatical errors that should have been automatically detected.
2. Mass-replace "loop-like" (in quotes) with loop-like without quotes for style issues.

proposals/break-continue-in-inline-lambdas.md
```diff
@@ -0,0 +1,175 @@
...
+The inspection should not triggerred for `let`, `run`, etc as they have the corresponding contract. There seems to be no need for
+a similar inspection for non-local returns. We assume that they are less confusing.
+
+## Implementation note
+
+Similarly to non-local returns, non-local break/continue will only work for lambdas passed as parameters not marked
+with `noinline` and `crossinline`.
+
+Non-local `break` and `continue` could be emulated using non-local return statements.
+For example, the following code snippets should be equivalent with respect to DFA:
```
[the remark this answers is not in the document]
nikitabobko: @strangepleasures sorry, I don't understand what you mean. I don't understand why this section relies on the semantic of `run`. `run` is just used because it's the most canonical way to invoke a lambda.

My point is different: I want that we really don't forget to check that the implementation is identical. And once we check it, we should mention it in the KEEP document. So the phrase "the following code snippets should be equivalent with respect to DFA" will turn into "in the current implementation the following code snippets is equivalent with respect to DFA"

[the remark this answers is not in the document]
elizarov: No need to mention "the current implementation", because people will be reading this document years later. Also, not every reader knows what DFA is, so the correct phrase would be "Note that the following code snippets are equivalent in terms of their data flow".

proposals/break-continue-in-inline-lambdas.md
elizarov: I think we've discussed this during design. We cannot make it an error, because we don't have a concept of loop-like functions in the language. We simply use a heuristic that is based on AT_MOST_ONCE/EXACTLY_ONCE contract. We know that functions with such a contract are _definitely not loop-like_, but the converse is not true. In reality, we don't know much about lambdas not marked and such, so a hard-requirement (as in "produces compiler ERROR") to use labelled return there would be an overkill. It much better belongs to a category of inspections, a kind of style issue.

proposals/break-continue-in-inline-lambdas.md
```diff
@@ -0,0 +1,148 @@
...
+    (0 until 200).forEach {
+        if (it == 42) break@forEach // Obviously, it breaks `forEach`
+        if (i == 10) continue // No label, so it applies to the 'for' loop
+    }
+}
+```
+
+A similar rule applies to `return` statements: an unlabeled return within a lambda is always non-local, a labeled one returns to the label.
+In other words, an unlabeled return goes to an innermost enclosing block that is clearly marked with fun hard keyword.
+In a similar way, an unlabeled break/continue go to an innermost enclosing block that is clearly marked with for, do, or while hard keywords.
```
elizarov, line 103: Please, use back-ticks around keywords. "`fun` hard keyword" reads much better and less ambiguously than "fun hard keyword". The same with `for`, `do`, and `while`.

proposals/break-continue-in-inline-lambdas.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Tooling for puzzling code
+
+Even if we don't implement support for `break`/`continue` for loop-like functions,
+s some users may find constructions containing `break` or `continue` within lambda arguments of loop-like inline functions puzzling:
```
elizarov, line 108: "loop-like functions, s some users ..." -> "loop-like functions, some users ..."

elizarov (changes requested):

proposals/break-continue-in-inline-lambdas.md
```diff
@@ -60,57 +64,45 @@ So the Motivation is the following:
...
-
-* **Q**: Ok, why not make labels mandatory for both non-local `return` and non-local `break`/`continue` then?
-* **A**: We can't make non-local `return` to be obligatory labeled because it would break inline functions first-class
-  citizenship in language constructs. E.g. we value "return if not null" idiomatic Kotlin construct `nullable?.let { return it }`.
-  If non-local `return` were obligatory labeled then the code would be much more verbose
-  `nullable?.let { return@functionName it }`.
-
-It worth noticing that we don't abandon the idea to provide some "break"/"continue" mechanism for "loop-like" functions
-[KT-19748](https://youtrack.jetbrains.com/issue/KT-19748). It just that we agree that we will never support unlabeled non-local
-`break`/`continue` for "loop-like" functions.
```
nikitabobko: @strangepleasures I think that you dropped an important part from the KEEP.

1. `break`/`continue` <-> `for`/`while` invariant notice
2. The reasoning that answers the question: "why are we ok with introducing a new language feature that brings `forEach` vs `for`/`while` puzzler" (Well, I think that that we can move this reasoning into a "Tooling for puzzling code" but I don't think that we should drop this reasoning)

proposals/break-continue-in-inline-lambdas.md
nikitabobko: Ok, since it's an inspection for an unreleased language feature we can try to be more aggressive because we don't color a lot of existing code in yellow. Users will file a ticket if it turns out that the inspection is too aggressive

elizarov (approved):

proposals/break-continue-in-inline-lambdas.md
nikitabobko: @elizarov Thanks! Add a section describing this decision https://github.com/Kotlin/KEEP/pull/327/commits/cb55a237387718a45d38edd5169e43c787d7f3cc

proposals/break-continue-in-inline-lambdas.md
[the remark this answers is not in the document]
strangepleasures: @nikitabobko but WHY do we need to check that in the first place? To me, this equivalence is just a fun fact, not a requirement. We could have also provided a semantically equivalent program written in different language, but what for?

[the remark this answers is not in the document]
nikitabobko: > but WHY do we need to check that in the first place?

We need to check it to try to find potential bugs we have in the current implementation. It's just one more check. What is the point of "it should work like that" phrase in a KEEP document if, in reality, it doesn't work "like that"? Let's please really make sure that it "works like that"

"it works like that" phrase (without "should"), gives one more invariant about the language.

nikitabobko (approved):

PR #331, opened 2022-11-07
`@JvmExposeBoxed` annotation for opening API of inline classes to Java

CommanderTvis:


elizarov (changes requested):
This document requires a lot more work to turn it into a finished design. We'll work offline on the list of required changes.

proposals/annotation-to-mark-accessible-api-for-java.md
```diff
@@ -0,0 +1,193 @@
...
+* Status: TODO
+* Discussion and feedback: TODO
+
+This document describes an annotation for the transformation of API written in Kotlin to be convenient for use from Java.
+
+## Motivation and use-cases
+
+Currently, `@JvmName` is used in three use cases:
+
+1. To bypass the “Platform declaration crash” error in many cases when Kotlin overloads are more potent than JVM ones. In this case, the user is not interested in setting a specific name.
```
IRus, line 14: Typo: "Platform declaration clash"

proposals/annotation-to-mark-accessible-api-for-java.md
```diff
@@ -0,0 +1,174 @@
...
+```java
+public static Example f-impl(java.lang.String x)
+public static Example f(Example x) {  }
+```
+
+The certain problem is that one cannot instantiate an inline class with its constructor because it is generated with `ACC_SYNTHETIC`. A solution for that could be annotating a constructor `@JvmExpose`  to have a constructor exposed by the compiler (it also will create an internal synthetic overload of it taking something like `Nothing?`). Since this requirement is unobvious, an IDE inspection must report that instances of an inline class taken as an argument of the exposed function cannot be instantiated.
+
+```kotlin
+@JvmInline
+value class Example @JvmExpose constructor(val s: String)
```
ilmirus, line 148: We cannot expose default constructor, as it is used in boxing. Thus, if inline class has `init` block, it will not be executed. The only possible solution I see is to generate factory function, which is accessible only from Java.

ilmirus (changes requested):

proposals/annotation-to-mark-accessible-api-for-java.md
CommanderTvis: Like Example.of("42")? Adding it implicitly is probably a bad solution.

CommanderTvis: Can't we do boxing by using a synthetic constructor with some flag?

k163377:
I did not see an issue, so let me comment directly.

How about allowing this annotation to be given to a single constructor or function as well?

I think there are only a limited number of situations where `value class` would be a problem.
Therefore, I feel that assigning `value class` on a per-class basis unnecessarily undermines the value of `value class`.

In addition, `value class` may not be defined or modified by the user.
Examples are `unsigned integers` and `kotlin.time.Duration`.

serras:
Closed in favor of #393

PR #332, opened 2022-11-09
Mark Markdown files as `linguist-detectable`

fwcd:
This makes them show up in the language bar following the precedent of similar repos like [`swift-evolution`](https://github.com/apple/swift-evolution):

image


elizarov:
Thanks a lot! Great stuff. 🚀

PR #333, opened 2022-11-20
Design Notes on Kotlin Type Expressions

thumannw:
Dear Kotlin designers, could this be a basis for a valuable design note?
@elizarov

### DIFF
--- a/notes/type-expressions.md
+++ b/notes/type-expressions.md
@@ -63,7 +63,7 @@ The following table lists the mentioned type expressions and their equivalents i
 | function type     | `fun interface`                        |
 | product type      | `data class` or `value class`          |
 | union type        | `sealed interface/class`, `enum class` |
-| tagged type       | -                                      |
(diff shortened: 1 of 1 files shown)


elizarov:
It would be better to follow the contribution process as explained int the [README](https://github.com/Kotlin/KEEP/blob/master/README.md). First of all, you shall start with the problem you are facing with your code and collect similar use-cases. I  would recommend you to start with the following issues that seems to be centered around  similar problems:

* [KT-45587](https://youtrack.jetbrains.com/issue/KT-45587) Tuples (structural literals and structural types)
* [KT-13108](https://youtrack.jetbrains.com/issue/KT-13108) Denotable union and intersection types

Also, when it comes to the type system, it is important to have some kind of a prototype or a good understanding of how the extensions can be made to work in the framework of the existing type system of the Kotlin language and the algorithms that are used for type inference.

PR #335, opened 2022-12-08
Introduce secondary constructors with body

zhelenskiy:


ilmirus (approved):

proposals/inline-classes.md
```diff
@@ -636,6 +636,21 @@ value class Negative(val x: Int) {
...
+```kotlin
+@JvmInline
+value class A(val x: Int) {
+    constructor(x: Double) : this(x.toInt()) {
+        println(x)
+    }
+}
+```
+
+As it is possible to use init blocks for inline classes since 1.4.30, this restriction can also be lifted in 1.8.20.
```
nikitabobko, line 652: > this restriction can also be lifted in 1.8.20.

The feature should have an experimental status at least for one Kotlin version. It's not possible to release "secondary constructor with body" feature in 1.8.20. Only in 1.9 (BTW, you use 1.9 Target version in YouTrack issue https://youtrack.jetbrains.com/issue/KT-55333)

If you meant that we will make it experimental in 1.8.20 then please be specific: "this restriction can also be lifted in 1.9 (release), 1.8.20 (experimental)."

zhelenskiy: Fixed

ilmirus (approved):

PR #338, opened 2022-12-19
KEEP for custom equals in inline classes

grechkovlad:


proposals/custom-equals-in-value-classes.md
```diff
@@ -0,0 +1,262 @@
...
+concept of **typed equals** was proposed:
+
+```Kotlin
+@JvmInline
+value class Degrees(val value: Double) {
+    fun equals(other: Degrees) = (value - other.value) % 360.0 == 0.0
+}
+```
+
+Type of `other` will be erased during compilation and passing argument will not require boxing.
```
ilmirus, line 96: Not erased, but unboxed.

proposals/custom-equals-in-value-classes.md
```diff
@@ -0,0 +1,262 @@
...
+
+More precise, we define typed equals as a function such that:
+
+* Has name `"equals"`
```
ilmirus, line 100: Better to make examples conform the restrictions. For example, the one above does not have the annotation.

proposals/custom-equals-in-value-classes.md
```diff
@@ -0,0 +1,262 @@
...
+    setOf(IC(1), IC(2))
+}
+```
+
+## Other questions
+
+### Inline classes over inline classes
+
+Suppose we have an inline class `B` over inline class `A` and `A` declares typed equals. In this case,
+default-generated equals in `B` will be comparing it underlying values using custom equals of `A`.
```
ilmirus, line 189: Will `B` require opt-in?

proposals/custom-equals-in-value-classes.md
```diff
@@ -0,0 +1,262 @@
...
+```Java
+value class ValueClass {
+    int x;
+
+    public ValueClass(int x) {
+        this.x = x;
+    }
+
+    @Override
+    public boolean equals(Object other) {
```
ilmirus, line 206: First, this `equals` does not follow its contract. Second, how typed equals will be compiled into Valhalla class method?

proposals/custom-equals-in-value-classes.md
```diff
@@ -0,0 +1,262 @@
...
+        System.out.println(zeroObj.hashCode());            // 42
+    }
+}
+```
+
+## Possible future improvements
+
+### Resolving references to typed equals in compiler frontend
+
+Currently `'=='` operator is being resolved to typed equals in compiler backend, but resolving it in frontend could let
```
ilmirus, line 234: Current decision is that `==` is never resolved on front-end, but IDE makes everything it can to resolve it. So, it is not about FE support, but rather IDE one.

ilmirus (changes requested):

proposals/custom-equals-in-value-classes.md
grechkovlad: Sure. Any usage of `A` will require opt-in and `B` will use `A` for declaring its primary constructor.

ilmirus (approved):

nikitabobko:
Closing due to inactivity

PR #351, opened 2023-04-19
Merge Data objects to master

nikitabobko:


nikitabobko:
@elizarov @zarechenskiy are you ok with "Rebase and merge"?

nikitabobko:
But let's firstly merge "data objects use-cases section" into `data-objects` branch https://github.com/Kotlin/KEEP/pull/352/files

qwwdfsad:
[Just passing by, don't mind me]

IMO it should be squashed -- there is no value in re-iterative commits, "invalid markdown syntax" etc. in the history; these are purely the technicality of the development process

PR #356, opened 2023-05-08
KEEP for Self types

maxim092001:


maxim092001:
> mgroth0

Hi!

There is [KT-6494](https://youtrack.jetbrains.com/issue/KT-6494) and [PR](https://github.com/JetBrains/kotlin/pull/4796)

proposals/self-types.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+## Motivation
+
+Self types can be implemented by programmer with boilerplate code using recursive generics and some additional unchecked casts. 
+They can be used in multiple useful and popular patterns, so there is a reason and community need to make it a language feature.
+
+## Usage examples
+
+One of the most common example of Self types application is [abstract builder pattern](https://medium.com/@hazraarka072/fluent-builder-and-powering-it-up-with-recursive-generics-in-java-483005a85fcd). However, in Kotlin builders are usually implemented via extension recievers.Although, if we want to have transformation chain of immutable object/data, Self types are really usefull.
```
cleveralbatraoz, line 32: > extension reciever**s.A**lthough

Missing space character

maxim092001: Thanks!

He-Pin:
Cool to have this, reduce a lots of boilerplates

proposals/self-types.md
```diff
@@ -0,0 +1,460 @@
...
+
+```kotlin
+open class A {
+    fun foo(): Self {
+        return this;
+    }
+}
+
+class B : A {
+
```
He-Pin, line 19:
```kotlin
class B: A {
  fun bar():Self {
     return this;
  }
}
```

proposals/self-types.md
```diff
@@ -0,0 +1,460 @@
...
+    fun foo(): Self {
+        return this;
+    }
+}
+
+class B : A {
+    
+}
+
+val x: B = B().foo(); // foo return type is B
```
He-Pin, line 22: val x:B = B().foo().bar()

proposals/self-types.md
maxim092001: We do not need `bar` to return `this`.

`foo` from class `A` returns `Self` type.
When we call `foo` for type `B` it will return type `B`.

This is basically the point of having self-types. Take a look at the transformation chains example.

proposals/self-types.md
maxim092001: `;` will be removed, for `bar` function take a look at the previous comment.

Thanks!

proposals/self-types.md
He-Pin: i mean we can call bar with self type, even the bar is defind on b, I think that's would be a better example.

maxim092001: I see, thats right!

I will update example

maxim092001: Done, thanks!

LaylaMeower:
I would replace the annotation with a modifier. Gate-locking a language feature behind an annotation isn't a common Kotlin thing to do - annotations are usually used for meta data (e.g., `@PublishedApi`).
```kotlin
class A<this Self>

class B<this Self, T>

class C<T, this Self>
```

That way, you reduce the confusion from the explicit-self-generic syntax (the `C<Int, C<Int, *>>` instead of `C<Int>` part), remove the need for an annotation (therefore increasing readability and beginner-approachability), and allows using whatever name (and order) you want for the generic instead of forcing it by language.

X1nto:
IMO a more applicable name would be `This`, since we also have the `this` keyword.

zarechenskiy:
Thanks for your PR! However, I'm closing it as right now we don't have plans of introducing self types in Kotlin. Please follow https://youtrack.jetbrains.com/issue/KT-6494 for the updates

PR #361, opened 2023-09-09
HexFormat proposal

qurbonzoda:


proposals/stdlib/hex-format.md
```diff
@@ -0,0 +1,320 @@
...
+* **Target issue**: [KT-57762](https://youtrack.jetbrains.com/issue/KT-57762/)
+* **Discussion**: TBD
+
+## Summary
+
+Convenient API for formatting binary data into hexadecimal string form and parsing back.
+
+## Motivation
+
+Our research has shown that hexadecimal representation is more widely used than other numeric bases,
```
fvasco, line 16: Citation needed

proposals/stdlib/hex-format.md
```diff
@@ -0,0 +1,320 @@
...
+// Formatting a ByteArray
+macAddress.toHexString(HexFormat { bytes.byteSeparator = ":" }) // "00:1b:63:84:45:e6"
+
+// Defining a format and assigning it to a variable
+val threeGroupFormat = HexFormat { upperCase = true; bytes.bytesPerGroup = 2; bytes.groupSeparator = "." }
+// Formatting a ByteArray using a previously defined format
+macAddress.toHexString(threeGroupFormat) // "001B.6384.45E6"
+```
+
+## Alternatives
```
fvasco, line 201: Another option is a third-party implementation. Logging, for example, is only available this way.

proposals/stdlib/hex-format.md
```diff
@@ -0,0 +1,320 @@
...
+            var prefix: String = ""  
+            var suffix: String = ""  
+            var removeLeadingZeros: Boolean = false  
+        }
+    }
+}
+```
+
+`BytesHexFormat` and `NumberHexFormat` classes hold format options for `ByteArray` and numeric values, correspondingly.
+`uppercase` option, which common to both `ByteArray` and numeric values, is stored in `HexFormat`.
```
qurbonzoda, line 132:
```suggestion
`uppercase` option, which is common to both `ByteArray` and numeric values, is stored in `HexFormat`.
```

mahozad: Hi,
I think the `uppercase` option is vague.
Does `uppercase = false` means force all letters to lowercase or does it mean keep the letters as is (not touch them)?
Can it be renamed to something like `forceUppercase` or similar?

proposals/stdlib/hex-format.md
qurbonzoda: We don't have any public document with the results. But search in [grep.app](https://grep.app/search?q=toString%5C%28%28radix%20%3D%20%29%3F16%5C%29&regexp=true&filter[lang][0]=Kotlin) finds much more usages of `toString(16)` than any other base.

proposals/stdlib/hex-format.md
qurbonzoda: The option affects only the hexadecimal digits to be used for formatting. It doesn't change the case of `prefix` or other options. Please refer to documentation of [the property](https://github.com/JetBrains/kotlin/blob/5d88a1f1c382f96fb88fa1d1a8741c6019795e1d/libraries/stdlib/src/kotlin/text/HexFormat.kt#L24) for details.
Anyway, any name suggestions are welcome. But `forceUppercase` doesn't seem to be more informative.

proposals/stdlib/hex-format.md
qurbonzoda: Could you also please provide a link to the library you are referring to?

fvasco: https://github.com/oshai/kotlin-logging

proposals/stdlib/hex-format.md
```diff
@@ -306,7 +306,7 @@ Only a subset of Kotlin Standard Library available on all supported platforms is
 * Adding the ability to limit the number of hex digits when formatting numeric values
   * `NumberHexFormat.maxLength` could be introduced
     * When formatting an `Int`, combination of `maxLength = 6` and `removeLeadingZeros = false` results to exactly 6 least significant hex digits
-    * Combination of `maxLength = 6` and `removeLeadingZeros = true` returns maximum of 6 hex digits without leading zeros
+    * Combination of `maxLength = 6` and `removeLeadingZeros = true` returns a maximum of 6 hex digits without leading zeros
```
ilya-g, line 309: Can be read as that one maximal digit among 6 digits is returned. Perhaps better "at most 6 hex (least-significant) digits"

proposals/stdlib/hex-format.md
fvasco: Probably `.toString( /* 10 */ )` is more common, should we also consider introducing `DecFormat`?
I'm not against this proposal; it seems strange to tackle hexadecimal number formatting before decimal ones.

proposals/stdlib/hex-format.md
qurbonzoda: fixed

proposals/stdlib/hex-format.md
qurbonzoda: The decimal representation of numeric values is indeed more common, and this proposal explicitly agrees with the proposition. However, this proposal is about introducing an API to facilitate formatting (and parsing) the hex representation of binary data. It provides use cases and proposes an API to make it easy to handle those use cases.
Of course, there are other potentially more impactful features. They deserve separate proposals.
`HexFormat` is not the most important or useful feature out there, but it addresses the pain points outlined in this document.

proposals/stdlib/hex-format.md
```diff
@@ -0,0 +1,320 @@
...
+            var prefix: String = ""  
+            var suffix: String = ""  
+            var removeLeadingZeros: Boolean = false  
+        }
+    }
+}
+```
+
+`BytesHexFormat` and `NumberHexFormat` classes hold format options for `ByteArray` and numeric values, correspondingly.
+`uppercase` option, which is common to both `ByteArray` and numeric values, is stored in `HexFormat`.
```
ilya-g, line 132: We should likely discuss the name of this option once again. While it is named `upperCase` in code, in other places of KEEP, it is referred to as `uppercase`, which may indicate that it could be a more natural choice of the name, also more consistent with the name of the `String/Char.uppercase()` function.

Another approach could use some prefix, for example, `isUpperCase` or `useUpperCase`.

ilya-g (approved):

proposals/stdlib/hex-format.md
qurbonzoda: Good point. I added a comment regarding the option name in the proposal discussion issue: https://github.com/Kotlin/KEEP/issues/362#issuecomment-1758610418
Hopefully, some naming variants will be proposed by the community.

proposals/stdlib/hex-format.md
qurbonzoda: Hi,
Discussion of the proposal has been moved here: https://github.com/Kotlin/KEEP/issues/362
Please don't hesitate to express your concerns there.

PR #363, opened 2023-11-01
Infinite loop

vmishenev:


proposals/infinite-loop.md
```diff
@@ -0,0 +1,250 @@
...
+
+Infinite loops is an idiom that is widely used in Kotlin programming. 
+In Kotlin the percentage of `while(true)` among all written `while` loops is 19% (in 1.2M repositories from the BigCode tool).
+
+- Infinite loops are widely used to monitor user input or device activity. 
+The idiomatic approach to reading all the lines from the input stream until it is over (until `readLine` function returns null):
+
+	```kotlin
+	while(true) {
+	    val line = input.readLine() ?: break
```
hoc081098, line 30: Hi, `readLine` is deprecated, ` readlnOrNull` can be used instead 🙏

vmishenev: Do you mean Java's  [`DataInputStream.readLine`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/io/DataInputStream.html#readLine())? Kotlin's [`readLine`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/read-line.html) is not deprecated. You can also have a look at [the implementation of `readlnOrNull` for JVM](https://github.com/JetBrains/kotlin/blob/05896a42da04620a2e8ca35e952ff31c3c48aa9d/libraries/stdlib/jvm/src/kotlin/io/Console.kt#L163)

Colerar: As in [What's new in Kotlin 1.6.0﻿ #New readline functions﻿](https://kotlinlang.org/docs/whatsnew16.html#new-readline-functions):

> We're planning to gradually deprecate the `readLine()` function in future releases.

> Do you mean Java's [DataInputStream.readLine](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/io/DataInputStream.html#readLine())? Kotlin's [readLine](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/read-line.html) is not deprecated. You can also have a look at [the implementation of readlnOrNull for JVM](https://github.com/JetBrains/kotlin/blob/05896a42da04620a2e8ca35e952ff31c3c48aa9d/libraries/stdlib/jvm/src/kotlin/io/Console.kt#L163)

Though the implementation in current version is the same, the semantics of `readlnOrNull` is more clear.

vmishenev: Indeed. It should be [soft deprecated](https://youtrack.jetbrains.com/issue/KT-54106/Provide-API-for-perpetual-soft-deprecation-and-endorsing-uses-of-more-appropriate-API).
Thank you for noticing it. I will fix it to avoid deprecated examples.

rebokdev:
Using "while { ... }" seems uncomfortable as there is no condition, maybe replace it with "repeat { ... }"

nikitabobko:

PR #366, opened 2023-12-21
Context parameters

serras:
Only comment here about the text itself. The discussion about the feature is done in [this issue](https://github.com/Kotlin/KEEP/issues/367).

This is an updated proposal for [KEEP-259](https://github.com/Kotlin/KEEP/issues/259), formerly known as _context receivers_. The new design addresses the issues raised by the users of the prototype and across the community.

### DIFF
--- a/proposals/context-parameters.md
+++ b/proposals/context-parameters.md
@@ -51,6 +51,7 @@ This document is not (yet) formally a KEEP, since it lacks some of the technical
 **§1.1** *(declaration)*: Function and property declarations get support for **context parameters**. Context parameters are declared with the `context` keyword followed by a list of parameters, each of the form `name: Type`.
 
 * Constructors may **not** declare context parameters.
+* Context parameters are allowed in operators, except for those related to [property delegation](https://kotlinlang.org/spec/declarations.html#delegated-property-declaration).
 * Within the body of the declaration, the value of the context parameter is accessible using its name, similar to value parameters.
 * It is allowed to use `_` as a name; in that case, the value is not accessible through any name (but still participates in context resolution).
 
@@ -172,6 +173,7 @@ fun <A, B, C, R> context(a: A, b: B, c: C, block: context(A, B, C) () -> R): R =
 **§2.2** *(`implicit` function)*: We also provide a generic way to obtain a value by type from the context. It allows access to context parameters even when declared using `_`, or within the body of a lambda.
 
 * Implementations are encouraged, but not required, to mark this function as `inline`.
+* If possible, type inference for type variable `A` should be restricted (for example, using `@NoInfer`). Developers are expected to use the function with explicit type arguments. 
 
 ```kotlin
 context(ctx: A) fun <A> implicit(): A = ctx
--- a/proposals/context-parameters.md
+++ b/proposals/context-parameters.md
@@ -447,15 +447,15 @@ class B { fun foo(x: Object) }
 // here 'foo' resolves to the one in 'B'
 // because 'B' is introduced in an inner scope
 // the fact that the one from 'A' is more specific doesn't play a role
-context(A) B.example1() = foo(5)
+context(A) fun B.example1() = foo(5)
 
 // 'this' always resolves to 'B'
-context(A) B.example2() = this.foo(5)
+context(A) fun B.example2() = this.foo(5)
 
 // to access the members from 'A' use a name, 'summon'
-context(A) B.example3a() = summon<A>().foo(5)
-context(a: A) B.example3b() = a.foo(5)
-context(a: A) B.example3b() = with(a) { foo(5) }
+context(A) fun B.example3a() = summon<A>().foo(5)
+context(a: A) fun B.example3b() = a.foo(5)
+context(a: A) fun B.example3b() = with(a) { foo(5) }
 ```
 
 **§29** *(resolution, most specific candidate)*: When choosing the **most specific candidate** we follow the Kotlin specification, with one addition:
@@ -514,7 +514,7 @@ class B { fun foo(x: Object) }
 // here 'foo' resolves to the one in 'A'
 // because both 'A' and 'B' are in the same scope
 // but the one from 'A' is more specific
-context(A) B.example1() = foo(5)
+context(A) fun B.example1() = foo(5)
 ```
 
 ## Q&A about design decisions
--- a/proposals/context-parameters.md
+++ b/proposals/context-parameters.md
@@ -142,6 +142,7 @@ withConsoleLogger {
 **§2.1** *(`context` function)*: The `context` function adds a new value to the context, in an anonymous manner.
 
 * The implementation may be built into the compiler, instead of having a plethora of functions defined in the standard library.
+* Implementations are encouraged, but not required, to mark these functions as `inline`.
 
 ```kotlin
 fun <A, R> context(context: A, block: context(A) () -> R): R = block(context)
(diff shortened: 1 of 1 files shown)


proposals/context-parameters.md
```diff
@@ -0,0 +1,560 @@
...
+
+```kotlin
+// Uses extension receivers, but File is in fact the "subject"
+fun ResourceScope.openFile(file: File): InputStream
+
+// Doing this is OK, but provides no additonal benefit
+context fun ResourceScope.openFile(file: File): InputStream
+
+// The API is nicer using context receivers
+@JvmName("contextualFileOpen")  // needed to avoid platform clash
```
lukellmann, line 329:
```suggestion
```
why would there be a platform clash? the "old" function is named `openFile`, the "new" function is called `open`

proposals/context-parameters.md
```diff
@@ -0,0 +1,560 @@
...
+class ConsoleLogger: Logger { ... }
+class FileLogger: Logger { ... }
+
+context(Logger) fun logWithTime(message: String) = ...
+
+context(ConsoleLogger, FileLogger) fun example1a() =
+  logWithTime("hello")  // ambiguity error
+  
+context(console: ConsoleLogger, file: FileLogger) fun example1b() =
+  logWithTime("hello")  // ambiguity error
```
lukellmann, line 488: Wouldn't `logWithTime` be unresolved here instead of an ambiguity error? From my understanding, the type of a named context parameter isn't part of the scope inside the function. Or am I wrong here?

serras: No, in every case `logWithTime` is resolved to the top-level function. The difference is whether there's ambiguity or not in the "context resolution" phase.

proposals/context-parameters.md
```diff
@@ -0,0 +1,559 @@
+# Context parameters
+
+* **Type**: Design proposal
+* **Author**: Alejandro Serrano
+* **Contributors**: Marat Ahkin, Nikita Bobko, Ilya Gorbunov, Mikhail Zarechenskii, Denis Zharkov
```
lounres, line 5: It looks like a typo.

```suggestion
* **Contributors**: Marat Akhin, Nikita Bobko, Ilya Gorbunov, Mikhail Zarechenskii, Denis Zharkov
```

proposals/context-parameters.md
```diff
@@ -0,0 +1,508 @@
...
+
+functionType: [ functionContext ] [ receiverType '.' ] ...
+functionContext: 'context' '(' receiverType { ',' receiverType } [ ',' ] ')'
+```
+
+**Recommended style:** annotations, context parameters, other modifiers as per the [usual style guide](https://kotlinlang.org/docs/coding-conventions.html#modifiers-order).
+
+### Extended resolution algorithm
+
+**§23** *(declaration with context parameters)*: The context parameters declared for a callable are available in the same way as "regular" value parameters in the body of the function. Both value and context parameters are introduced in the same scope, there is no shadowing between them.
```
dzharkov, line 410: Maybe it's worth mentioning that we should forbid redeclarations there.

proposals/context-parameters.md
```diff
@@ -0,0 +1,508 @@
...
+  logWithTime("hello")  // no ambiguity, uses the new ConsoleLogger
+}
+
+context(console: ConsoleLogger, file: FileLogger) fun example3() =
+  with(console) { logWithTime("hello") }  // no ambiguity, uses 'console'
+```
+
+**§26** *(applicability, `DslMarker`)*: During context resolution, if at a certain scope there is a potential contextual value in scope (either coming from a context parameter or from an implicit receiver) marked with an annotation `@X` which is itself annotated with [`@DslMarker`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-dsl-marker/) then:
+
+- It is an _error_ for two such values to be available in the same scope.
```
dzharkov, line 446: Even if only one of them is applicable?

proposals/context-parameters.md
```diff
@@ -0,0 +1,508 @@
...
+- If context resolution chooses a contextual value with the same annotation, but in an outer scope, it is a compilation _error_.
+- If a call binds to a receiver with the same annotation, it is a compilation _error_.
```
dzharkov, line 448: TBH, I don't understand what this part means. May you please give some example?

dzharkov (changes requested):

proposals/context-parameters.md
```diff
@@ -0,0 +1,559 @@
...
+
+### Dependency injection
+
+**§E.1** *(Dependency injection use case)*: You can view context parameters as values that must be "injected" from the context for the code to work. Since context arguments are completely resolved at compile-time, this provides something like dependency injection in the language.
+
+**§E.1** *(Dependency injection use case, companion object)*: In some cases, you may want to define that instantiating a certain class requires some value — this creates the typical hierarchy of dependency we see in DI frameworks. We can accomplish this by faking a constructor with those contexts:
+
+```kotlin
+interface Logger { ... }
+interface UserService { ... }
```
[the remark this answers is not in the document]
serras: There was a missing extends clause, thanks for catching it!

proposals/context-parameters.md
```diff
@@ -0,0 +1,571 @@
...
+  fun body(block: HtmlScope.() -> Unit)
+}
+```
+
+Context parameters lift two of the main restrictions of this mode of use:
+
+* It's possible to add new members with an extension receiver without modifying the Scope class itself.
+* It's possible to add members which are only available when the DSL Scope has certain type arguments.
+
+**§4.3.1** *(`DslMarker`)*: we strive to make `@DslMarker` annotations work uniformly across receivers and context parameters, as described in §7.5.
```
osipxd, line 290: Shouldn't it be 4.3.2?
```suggestion
**§4.3.2** *(`DslMarker`)*: we strive to make `@DslMarker` annotations work uniformly across receivers and context parameters, as described in §7.5.
```

proposals/context-parameters.md
```diff
@@ -0,0 +1,571 @@
...
+* [Technical design](#technical-design)
+  * [Syntax](#syntax)
+  * [Extended resolution algorithm](#extended-resolution-algorithm)
+  * [Extended type inference algorithm](#extended-type-inference-algorithm)
+  * [JVM ABI and Java compatibility](#jvm-abi-and-java-compatibility)
+* [Q\&A about design decisions](#qa-about-design-decisions)
+* [Acknowledgments](#acknowledgments)
+
+
+## Members with context parameters
```
ilya-g, line 49: "Members" makes an impression that context parameters are allowed only in member functions and properties, i.e. ones inside a class or an object.

proposals/context-parameters.md
```diff
@@ -0,0 +1,571 @@
...
+  // you can use 'implicit' to access the context parameter
+  implicit<Logger>().log("hello")
+}
+```
+
+## Standard library support
+
+**§2.1** *(`context` function)*: The `context` function adds a new value to the context, in an anonymous manner.
+
+* The implementation may be built into the compiler, instead of having a plethora of functions defined in the standard library.
```
ilya-g, line 144: This likely would complicate providing documentation for them.

proposals/context-parameters.md
```diff
@@ -0,0 +1,571 @@
...
+
+context(serializer: ToJson<T>) fun <T> T.toJson(): Json =
+  serializer.toJson(this)
+
+context(_: ToJson<T>) fun <T> ApplicationCall.json(thing: T) =
+  this.respondJson(thing.toJson())
+```
+
+We strongly advise _against_ creating "copies" of an API but with an added context parameter.
+Overload resolution is tricky, and the developer may end up calling the unintended version.
```
ilya-g, line 309: The example doesn't illustrate what is dangerous in this case.

proposals/context-parameters.md
serras: We might need to be a bit more creative here. Maybe the regular Dokka-style docs are not good in this case; I would imagine this being part of the language documentation, given how central they are going to be to the context parameters feature itself.

PR #370, opened 2024-02-07
Guards

serras:
Only comment here about the text itself. The discussion about the feature is done in ???.

This KEEP proposes an extension of branches in `when` expressions with subject, which unlock the ability to perform multiple checks in the condition of the branch.

### DIFF
--- a/proposals/guards.md
+++ b/proposals/guards.md
@@ -1,0 +1,245 @@
+# Guards
+
+* **Type**: Design proposal
+* **Author**: Alejandro Serrano
+* **Contributors**: Mikhail Zarechenskii
+* **Discussion**: ??
+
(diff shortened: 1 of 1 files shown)


proposals/guards.md
```diff
@@ -0,0 +1,245 @@
+# Guards
+
+* **Type**: Design proposal
+* **Author**: Alejandro Serrano
+* **Contributors**: Mikhail Zarechenskii
+* **Discussion**: [KEEP-371](https://github.com/Kotlin/KEEP/issues/371)
+
```
udalov, line 7: Maybe there should be a link to [KT-13626](https://youtrack.jetbrains.com/issue/KT-13626) somewhere.

PR #372, opened 2024-02-20
Base64 proposal

qurbonzoda:
Only comment here about the text itself. The discussion about the feature is done in https://github.com/Kotlin/KEEP/issues/373.

This KEEP proposes to introduce a multiplatform API for Base64 encoding and decoding.

### DIFF
--- a/proposals/stdlib/base64.md
+++ b/proposals/stdlib/base64.md
@@ -6,7 +6,7 @@
 * **Status**: Implemented in Kotlin 1.8.20
 * **Prototype**: Implemented
 * **Target issue**: [KT-57762](https://youtrack.jetbrains.com/issue/KT-55978/)
-* **Discussion**: TBD
+* **Discussion**: [KEEP-373](https://github.com/Kotlin/KEEP/issues/373)
 
 ## Summary
 
@@ -92,6 +92,8 @@ operations is required, the following operations are proposed:
 
 ### Proposed API
 
+#### Encode and decode functions
+
 It is proposed to introduce a `Base64` class that declare the operations above as member functions:
 ```kotlin
 public open class Base64 {
@@ -138,8 +140,10 @@ public open class Base64 {
 }
 ```
 
-However, the class won't have public constructors. It is not supposed to be instantiated or inherited.
-Instead, a predefined instance of this class will be provided for each supported Base64 variant:
+#### Construction
+
+The class won't have public constructors, and it is not supposed to be inherited or instantiated by calling its
+constructor. Instead, a predefined instance of this class will be provided for each supported Base64 variant:
 ```kotlin
 public open class Base64 {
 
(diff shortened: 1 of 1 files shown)


proposals/stdlib/base64.md
```diff
@@ -0,0 +1,313 @@
...
+## Reference implementation
+
+* `Base64` class: https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/src/kotlin/io/encoding/Base64.kt
+* `InputStream` and `OutputStream` extensions: https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/src/kotlin/io/encoding/Base64IOStream.kt
+* Test cases for `Base64` member functions: https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/test/io.encoding/Base64Test.kt
+* Test cases for `InputStream` and `OutputStream` extensions: https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/test/io/Base64IOStreamTest.kt
+
+## Future advancements
+
+* Adding ability to allow non-zero pad bits
```
ilya-g, line 312:
```suggestion
* Adding an option to allow non-zero pad bits on decoding
```

proposals/stdlib/base64.md
```diff
@@ -0,0 +1,313 @@
...
+```
+
+### Proposed naming
+
+When researching usages of Java `Base64`, we observed two patterns that influenced our design and naming.
+First, the vast majority of users obtain an encoder or decoder and use it immediately,
+e.g., `Base64.getDecoder().decode(...)`. This led us to avoid separating the encoder and decoder into different entities.
+We also didn't find evidence that the benefits of such a separation outweigh the usage inconvenience. 
+Second, encoding to `String` and decoding from `String` are considerably more popular than encoding to `ByteArray` and
+decoding from `ByteArray`. Thus, we decided to make the more frequently used functions shorter.
```
ilya-g, line 283: Another reason was that a compound operation like "encode bytearray to base64 string + encode string to utf8 byte array" should have a compound name.

qurbonzoda: The reverse operation (utf8 byte array -> string -> byte array) doesn't have a compound name.

proposals/stdlib/base64.md
```diff
@@ -0,0 +1,313 @@
...
+```
+
+#### Construction
+
+The class won't have public constructors, and it is not supposed to be inherited or instantiated by calling its
+constructor. Instead, a predefined instance of this class will be provided for each supported Base64 variant:
+```kotlin
+public open class Base64 {
+
+    public companion object Default : Base64() {
```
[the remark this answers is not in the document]
qurbonzoda: Done. Thanks for pointing out!

proposals/stdlib/base64.md
qurbonzoda: I've applied the proposed changes.

proposals/stdlib/base64.md
ilya-g: Yes, in that case we didn't complicate the name thanks to overloading on parameter types.

ilya-g (approved):

proposals/stdlib/base64.md
qurbonzoda: Done!

PR #374, opened 2024-04-12
Multi-dollar interpolation

serras:
Only comment here about the text itself. This discussion about the feature itself takes place in [this issue](https://github.com/Kotlin/KEEP/issues/375).

### DIFF
--- a/proposals/dollar-escape.md
+++ b/proposals/dollar-escape.md
@@ -1,0 +1,154 @@
+# Multi-dollar interpolation
+
+* **Type**: Design proposal
+* **Authors**: Alejandro Serrano Mena
+* **Discussion**: [KEEP-375](https://github.com/Kotlin/KEEP/issues/375)
+* **Status**: Experimental expected for 2.1
+* **Prototype**: Implemented in [this branch](https://github.com/JetBrains/kotlin/compare/rr/serras/dollar-escape-3)
+* **Related YouTrack issue**: [KT-2425](https://youtrack.jetbrains.com/issue/KT-2425/Provide-a-way-for-escaping-the-dollar-sign-symbol-in-multiline-strings-and-string-templates)
+
+## Abstract
+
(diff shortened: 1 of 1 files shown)


proposals/dollar-escape.md
```diff
@@ -0,0 +1,144 @@
...
+* [Proposed solution](#proposed-solution)
+  * [Single-line string literals](#single-line-string-literals)
+* [Alternatives](#alternatives)
+
+## Motivating examples
+
+Strings are one of the fundamental types in Kotlin, developers routinely create (parts of) them by using string literals. However, the current design has a few inconveniences, as witnessed by this [YouTrack issue](https://youtrack.jetbrains.com/issue/KT-2446/String-literals). This KEEP pertains to how to improve the situation around `$` and `"`, especially in multiline strings. It is a non-goal to change the behavior regarding indentation (or stripping thereof).
+
+[Kotlin's multiline strings](https://kotlinlang.org/docs/strings.html#multiline-strings) are raw, that is, every character from the start to the end markers is taken as it appears. In particular, there are no escaping sequences (`\n`, `\t`, ...) as found in single-line strings. Still, `$` is used to mark interpolation, and `""""` is used to mark the end of the string. If you need those characters in the string, the most often used workaround is to interpolate the character, leading to an awkward sequence of characters, like `${'$'}`.
```
lukellmann, line 27:
```suggestion
[Kotlin's multiline strings](https://kotlinlang.org/docs/strings.html#multiline-strings) are raw, that is, every character from the start to the end markers is taken as it appears. In particular, there are no escaping sequences (`\n`, `\t`, ...) as found in single-line strings. Still, `$` is used to mark interpolation, and `"""` is used to mark the end of the string. If you need those characters in the string, the most often used workaround is to interpolate the character, leading to an awkward sequence of characters, like `${'$'}`.
```
only three `"` end the string

Issue #385, opened 2024-08-12
Resolution of links to extensions in KDoc


whyoleg:
This is an issue to discuss **Resolution of links to extensions** in KDoc. The full text of the proposal is [here](https://github.com/Kotlin/KEEP/blob/kdoc/extension-links/proposals/kdoc/links-to-extensions.md).

During the migration of Dokka analysis to K2, several questions arose around KDoc links resolution.
In particular, it was unclear how a link to an extension should be resolved in the presence of type parameters.
Link resolution in KDoc is not fully specified, and for this reason some cases are currently implemented
differently in K1 and K2 analysis.

The goal of the proposal is to try to describe consistent rules on how KDoc links to extensions should be resolved.

CLOVIS-AI:
I'm not sure this is on-topic for this KEEP, but I'd very much like to see a way to disambiguate links of different things that have the same name. For example, when a class and a top-level function exist, it is not currently possible to create a link that refers to one of them specifically. Unlike regular function overloads, which are all listed on the same page, so the disambiguation isn't particularly needed, homonyms on different pages hinder the documentation quality at the moment.

whyoleg:
Hey @CLOVIS-AI, this is known and tracked in https://youtrack.jetbrains.com/issue/KT-19765/KDoc-Provide-ability-disambiguate-links-to-classes-vs-links-to-functions
Feel free to add a use-case there.

marcopennekamp:
> If `Type` has type parameters, e.g `interface Container<T, C: Iterable<T>>`, then the link should be resolved if it's possibly to substitute them with some concrete types in callable reference

We should describe the substitution more precisely. The definition here is quite mathematical (as opposed to algorithmic), which has the problem that the search space might be very large. When the compiler compiles a callable reference in code, the reference already has its type arguments set. Hence, it's quite easy to check them against the expected type parameters. On the other hand, a type reference in KDoc does not have its type arguments set. So there needs to be another mechanism to match the type reference against the extension function's extension receiver.

To make sure that there can be a proper and well performing implementation, I think the KEEP should describe the algorithm which matches the type references against the declared extension receiver.

(Maybe I got a bit carried away, but what follows is my attempt at such an algorithm.)

### Naive instantiation

Let's say we have an extension function KDoc reference `Type.extension` with `Type<A_0, A_1, ..., A_N>`.

[...]

PR #386, opened 2024-08-12
Resolution of links to extensions in KDoc

whyoleg:
Only comment here about the text itself. The discussion about the feature is done in https://github.com/Kotlin/KEEP/issues/385

The goal of the proposal is to try to describe consistent rules on how KDoc links to extensions should be resolved.


nikitabobko:

PR #387, opened 2024-08-22
Describe moving kotlinx.datetime.Instant to stdlib

dkhalanskyjb:


proposals/stdlib/instant.md
```diff
@@ -0,0 +1,1365 @@
...
+theoretical.
+
+#### The deprecated `now`
+
+A common way to obtain the current `Instant` on the JVM is to call
+`Instant.now()`:
+
+* <https://grep.app/search?q=Instant.now%28%29> is 13000 hits as of writing.
+* <https://grep.app/search?q=Clock.systemUTC%28%29.instant%28%29> is less than
+  100 hits.
```
JakeWharton, line 945: Not sure this comparison is useful here since these expressions are equivalent. The point of the `Clock` abstraction is instead to facilitate inversion of control as noted in its section below. You'd have to compare it to those calling `Clock::instant` which is much harder to search for. If we assume people name their injected `Clock` as "clock", you get 1200 hits: https://grep.app/search?q=clock.instant%28%29. That's not exhaustive, but it's probably a majority.

proposals/stdlib/instant.md
```diff
@@ -0,0 +1,1365 @@
...
+ * The conversion is lossy: JS uses millisecond precision to represent dates,
+ * and [Instant] allows for nanosecond resolution.
+ */
+public fun Instant.toJSDate(): Date
+
+public fun Date.toKotlinInstant(): Instant
+```
+
+```kotlin
+// Darwin
```
JakeWharton, line 1080: According to the UUID KEEP, the stdlib is not specialized for various native targets today. These extensions would no longer be possible.

proposals/stdlib/instant.md
```diff
@@ -0,0 +1,1365 @@
...
+
+```kotlin
+val maxTimestamp = Instant.fromEpochSeconds(MAX_POSTGRES_INTERVAL)
+
+val nextEvent =
+    timestamps.map(Instant::fromEpochSeconds)
+        filter { it > currentTime }.minOrNull() ?: maxTimestamp
+
+if (nextEvent != maxTimestamp) {
+    println("The next event is $earliestEvent")
```
lukellmann, line 818:
```suggestion
val nextEvent =
    timestamps.map(Instant::fromEpochSeconds)
        .filter { it > currentTime }.minOrNull() ?: maxTimestamp

if (nextEvent != maxTimestamp) {
    println("The next event is $nextEvent")
```

proposals/stdlib/instant.md
```diff
@@ -0,0 +1,1365 @@
...
+A common way to obtain the current `Instant` on the JVM is to call
+`Instant.now()`:
+
+* <https://grep.app/search?q=Instant.now%28%29> is 13000 hits as of writing.
+* <https://grep.app/search?q=Clock.systemUTC%28%29.instant%28%29> is less than
+  100 hits.
+
+It is not obvious to people who want to obtain the current `Instant` that a
+class like `Clock` even exists, so we direct them to it using a deprecation with
+a proposed replacement.
```
lukellmann, line 949: Isn't documentation for the `Instant` class enough?

proposals/stdlib/instant.md
```diff
@@ -0,0 +1,1365 @@
...
+  - Deprecates `kotlinx.datetime.Instant` and `kotlinx.datetime.Clock`
+    with a warning and a replacement with
+    `kotlin.time.Instant` and `kotlin.time.Clock`.
+  - Deprecates every function that accepts an `Instant`/`Clock`,
+    adds an overload that accepts `kotlin.time.(Instant|Clock)`
+    instead. The proposed suggestions involve the new converters.
+    Exception: `Clock.asTimeSource` does not get the new overload.
+  - For every function that returns but doesn't accept an `Instant`,
+    hides it and adds a new function, one that returns `kotlin.time.Instant`,
+    but with a different `JvmName`.
```
lukellmann, line 1235: I think a different `JvmName` isn't needed. Iirc two methods with the same name but different return types can exist in the JVM bytecode.

proposals/stdlib/instant.md
```diff
@@ -0,0 +1,1365 @@
...
+* "2024-08-05, 14:34:31" is not an `Instant`,
+  because people in Tokyo and Marocco would not observe this on their clocks
+  simultaneously.
+* "2024-08-05 UTC" is not an `Instant`, as it's too coarse.
+  We define `Instant` values to have a precision of several seconds.
+
+### `Clock`
+
+A `Clock` implementation is a way to obtain the current `Instant`.
+The `Clock.SYSTEM` default implementation, available on all platforms, queries
```
fvasco, line 48: Clock.SYSTEM -> Clock.System

proposals/stdlib/instant.md
```diff
@@ -0,0 +1,1365 @@
...
+This information is stored as `Instant` values.
+
+#### Determining time intervals between events
+
+**Example**: "This user was last online three minutes ago".
+
+**Example**: listing all files that were modified in the last 24 hours.
+
+These calculations require knowing the current moment in time (accessed via
+`Clock.SYSTEM`) and the moment in time when some event happened.
```
fvasco, line 104: Clock.SYSTEM -> Clock.System

proposals/stdlib/instant.md
dkhalanskyjb: Yes, you are right. For other platforms, though, it's still needed. Edited.

proposals/stdlib/instant.md
```diff
@@ -0,0 +1,1420 @@
...
+the `Instant` class and the `Clock` interface.
+The goal of this proposal is to move this widely applicable functionality into
+the Standard Library.
+
+Overview
+--------
+
+### `Instant`
+
+`Instant` denotes a moment in time, regardless of the observer.
```
benjishults, line 29: Maybe best to remove the phrase "regardless of observer" since that evokes relativistic effects.  According to the theory of relativity, simultaneity is relative.  If events occur in different places, there may be frames of reference in which they are simultaneous, other frames in which one occurs first and other frames in which the other occurred first.

There is no "absolute" frame of reference from which we can say that two events occurring at different places occurred at the "same instant".

So, I'm thinking it's probably best not to mention "observers" since none of our usual language and intuition about instants or clocks works when we try to account for observers in different reference frames.

dkhalanskyjb: Doesn't "regardless of the observer" strongly imply that relativistic effects are not accounted for, given that this definition is absurd in the context of relativity?

benjishults: "Regardless of observer" means "for any observer."  Or "in any frame of reference."

But everything else we're talking about here is only true if everyone is in the same frame of reference.

In other words, only for specific observers.

So using the phrase "regardless of observer" is the opposite of what we want.

Probably best just not to mention observers.

proposals/stdlib/instant.md
dkhalanskyjb: If anyone does read it, it should be enough. `Instant.now()` is for establishing a smooth enough migration using the compiler errors that it's straightforward to arrive at correct code without reading anything, just with autocomplete.

proposals/stdlib/instant.md
```diff
@@ -0,0 +1,1476 @@
...
+public val Instant.isDistantFuture: Boolean
+    get() = this >= Instant.DISTANT_FUTURE
+```
+
+
+All parts of this API have existed in `kotlinx-datetime` for a while now and
+have stable, widely used and thoroughly tested implementations.
+
+This API is based on the eponymous API entry in JSR 310 (available in
+the 310bp project and the standard library of Java9 and later):
```
lukellmann, line 490: Isn't it available since Java 8?

fvasco:
> The Java Time API defined by JSR 310 is provided as part of JDK-8

https://jcp.org/aboutJava/communityprocess/pfd/jsr310/JSR-310-guide.html

proposals/stdlib/instant.md
```diff
@@ -0,0 +1,1476 @@
...
+### `Instant` API
+
+```kotlin
+/**
+ * A moment in time.
+ *
+ * A point in time must be uniquely identified in a way that is independent of a time zone.
+ * For example, `1970-01-01, 00:00:00` does not represent a moment in time since this would happen at different times
+ * in different time zones: someone in Tokyo would think it is already `1970-01-01` several hours earlier than someone in
+ * Berlin would. To represent such entities, use the `kotlin-time time.
```
qurbonzoda, line 213:
```suggestion
 * Berlin would. To represent such entities, use `LocalDateTime` from the `kotlinx-datetime` library.
```

proposals/stdlib/instant.md
```diff
@@ -0,0 +1,1476 @@
...
+This makes the interface a clock in the sense of "system clock",
+without the ability to serve as a wall-clock.
+
+`now()` is not a method on `Instant`, even though that would be more convenient,
+in order to encourage writing testable code using dependency injection.
+If some function in the depths of the system calls `Instant.now()`, it becomes
+nondeterministic and difficult to reliably test, whereas passing a `Clock`
+to it makes calling `now()` a functionally pure operation.
+
+### Migration process
```
ilya-g, line 1225: I'd leave a description of that process we've chosen to implement.

proposals/stdlib/instant.md
```diff
@@ -0,0 +1,1476 @@
+Instant and Clock.
+=================
+
+* **Type**: Standard Library API proposal
+* **Author**: Dmitry Khalanskiy
+* **Contributors**: Ilya Gorbunov, Roman Elizarov, Vsevolod Tolstopyatov
+* **Status**: pending
+* **Prototype**: available in `kotlinx-datetime`
```
ilya-g, line 8:
```suggestion
* **Prototype**: available in `kotlinx-datetime`
* **Discussion**: [KEEP-395](https://github.com/Kotlin/KEEP/issues/395)
```

proposals/stdlib/instant.md
fzhinkin: We also need to update KDoc in the corresponding implementation. @dkhalanskyjb

proposals/stdlib/instant.md
```diff
@@ -0,0 +1,1420 @@
...
+
+However, practically, the "perfectly synchronized clocks" and "same physical
+moment" requirements make this discrepancy impossible to observe unless
+someone is actively trying to: this one-second difference between clock readings
+occurs much more often due to
+[clock drift](https://en.wikipedia.org/wiki/Clock_drift) than it does due to
+leap seconds.
+
+<https://github.com/ThreeTen/threeten-extra/tree/4e016340b97cab604114d10e02a672c1e94c6be5/src/main/java/org/threeten/extra/scale>
+is an implementation that converts between "UTC `Instant`", a new class, and
```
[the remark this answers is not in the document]
fzhinkin: @dkhalanskyjb should we use the actual class name here?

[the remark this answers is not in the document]
dkhalanskyjb: I don't think the name of the class is important here. If it were called `LeapSecondAwareInstant` or something similar, I'd still prefer calling it a "UTC `Instant`", that is, an `Instant` using the UTC time-scale.

proposals/stdlib/instant.md
ilya-g: I meant leave only that description because having others at this point doesn't bring much use.

proposals/stdlib/instant.md
fzhinkin: > We also need to update KDoc in the corresponding implementation.

Never mind, it looks OK as of now.

ilya-g (changes requested):
It's better to leave only the actual chosen migration process description.

ilya-g (approved):

PR #388, opened 2024-08-26
Streamline KDoc ambiguity links

vmishenev:
Only comment here about the text itself. The discussion about the feature is done in #389.

This document introduces clarifications and improvements to the existing implementation of [KDoc](https://kotlinlang.org/docs/kotlin-doc.html) regarding resolving ambiguous KDoc links.


nikitabobko:

PR #390, opened 2024-08-30
Update proposals/subclass-opt-in-required.md

BlondeHex:


proposals/subclass-opt-in-required.md
```diff
@@ -184,8 +190,76 @@ may be used within its body or signatures (`UnstableApi` types or overridden met
...
+    @RequiresOptIn(scope = Scope.All)
+    annotation class PoisonAll(val message: String)
+    
+    @RequiresOptIn(scope = Scope.Inheritance)
+    annotation class PoisonOnlySubclasses(val message: String)
+    ```
+   This design was rejected because it limits the ability to use the same experimental annotation marker for different scopes:
+   either marking the entire API as unstable or marking only inheritance as unstable.   
+    
+
```
ice-phoenix, line 260: Nit: Maybe talk about the `@Repeatable` design also and reject it, because of the symmetry with `@OptIn`?

ice-phoenix: And the fact we do not really want to introduce repeatable annotations

ice-phoenix (changes requested):

ice-phoenix (approved):
👍

PR #393, opened 2024-09-12
Expose boxed inline value classes in JVM

serras:
Only comment here about the text itself. The discussion about the feature is done in https://github.com/Kotlin/KEEP/issues/394.

We propose modifications to how value classes are exposed in JVM, with the goal of easier consumption from Java. This includes exposing the constructor publicly and giving the ability to expose boxed variants of operations.

### DIFF
--- a/proposals/jvm-expose-boxed.md
+++ b/proposals/jvm-expose-boxed.md
@@ -19,7 +19,6 @@ We propose modifications to how value classes are exposed in JVM, with the goal 
 * [Motivation](#motivation)
   * [Design goals](#design-goals)
 * [The `JvmExposeBoxed` annotation](#the-jvmexposeboxed-annotation)
-  * [Explicit API mode](#explicit-api-mode)
 * [Expose boxed constructors](#expose-boxed-constructors)
   * [Serialization](#serialization)
   * [No argument constructors](#no-argument-constructors)
@@ -27,6 +26,8 @@ We propose modifications to how value classes are exposed in JVM, with the goal 
   * [Other design choices](#other-design-choices)
 * [Expose operations and members](#expose-operations-and-members)
   * [`Result` is excluded](#result-is-excluded)
+  * [Abstract members](#abstract-members)
+  * [When no exposure may happen](#when-no-exposure-may-happen)
   * [Annotations](#annotations)
   * [Other design choices](#other-design-choices-1)
   * [Further problems with reflection](#further-problems-with-reflection)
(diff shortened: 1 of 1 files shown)


proposals/jvm-expose-boxed.md
```diff
@@ -0,0 +1,268 @@
...
+  public static box-impl(number: Int): PositiveInt = <init>(number, null)
+  // and others
+}
+```
+
+Note that this change is binary compatible with previous users of this code:
+
+- The previous compilation scheme never calls the constructor directly; it always uses `box-impl` if boxing is required in the Kotlin side. But the signature of `box-impl` is not changed, only its implementation.
+- Java users were not able to call the constructor, since it was marked as synthetic (and private). So exposing the constructor is compatible.
+- In the scenario in which the constructor was called directly (for example, by reflection), the code is still correct. There might be some performance hit, since now the `init` block is executed, but this is not a big concern.
```
udalov, line 123: Maybe I misunderstand this, but I don't see how the code calling the constructor directly by reflection is still working, because now that constructor has an extra parameter.

proposals/jvm-expose-boxed.md
```diff
@@ -0,0 +1,268 @@
...
+> [!NOTE]
+> In the following code examples we use `static` to highlight those places where the compilation scheme produced a static method in JVM bytecode.
+> This does not imply any particular stance on future inclusion of `static` members in the Kotlin language.
+
+### Design goals
+
+"Exposing" a `value` class is not a single concept. The following are some scenarios that we would like to support:
+
+1. Creating boxed values: for example, creating an actual instance of `PositiveInt`.
+2. Calling operations that take or return value classes using their boxed representations.
```
udalov, line 72: Besides calling, there's also overriding, which the current proposal doesn't handle, but perhaps it should be mentioned explicitly.

proposals/jvm-expose-boxed.md
```diff
@@ -0,0 +1,268 @@
...
+
+**Exposing a factory method**: instead of exposing the constructor, use `PositiveInt.of(3)`. This option is nowadays quite idiomatic in the Java world (for example, in `List.of(1, 2)`), but not as much as constructors. Using a factory method also has the drawback of possible clashes for the name of the factory method, which we avoid when using the constructor.
+
+## Expose operations and members
+
+The current compilation scheme transforms every operation where a value class is involved into a static function that takes the unboxed variants, and whose name is mangled to prevent clashes. This means those operations are not available for Java consumers. We propose introducing a new `@JvmExposeBoxed` annotation that exposes a variant of the function taking and returning the boxed versions instead (if more than one argument or return type is a value class, the aforementioned variant uses the boxed versions for _all_ of them). We call it the _boxed variant_ of the function.
+
+```kotlin
+annotation class JvmExposedBoxed(val jvmName: String = "")
```
udalov, line 165: It would be nice to specify retention and targets explicitly here.

proposals/jvm-expose-boxed.md
```diff
@@ -0,0 +1,268 @@
...
+```
+
+The `@JvmExposeBoxed` annotation may be applied to a declaration, or a declaration container (classes, files). In the latter case, it should be taken as applied to every single declaration within it.
```
udalov, line 168: Is it allowed on open members? If yes, it kind of makes [KT-31420](https://youtrack.jetbrains.com/issue/KT-31420) partially supported which is a bit suspicious. In particular, we should probably examine all problems keeping us from implementing KT-31420 and see how they affect `@JvmExposeBoxed`.

proposals/jvm-expose-boxed.md
```diff
@@ -0,0 +1,268 @@
...
+
+If the `@JvmExposeBoxed` annotation is applied to a member of a value class (or the value class itself, or the containing file), the boxed variant is compiled to a member function of the corresponding boxed class. The unboxed variant is still compiled as a static member of the class.
+
+The name of the boxed variant coincides with the name given in the Kotlin code unless the `jvmName` argument of the annotation is present. In that case, the name defined in the annotation should be used. Note that `@JvmName` applies to the unboxed variant.
+
+> [!NOTE]
+> There is a corner case in which `@JvmExposeBoxed` with a name is always needed: when the function has no argument which is a value class, but returns a value class.
+> In that case the compilation scheme performs no mangling. As a result, without the annotation the Java compiler would produce an ambiguity error while resolving the name.
```
udalov, line 176: This is not fully correct, mangling is still performed for class members.

proposals/jvm-expose-boxed.md
```diff
@@ -0,0 +1,325 @@
...
+## The `JvmExposeBoxed` annotation
+
+We propose introducing a new `@JvmExposeBoxed` annotation, defined as follows.
+
+```kotlin
+@Target(
+  CONSTRUCTOR, PROPERTY, FUNCTION, // callables
+  CLASS, FILE,                     // containers
+)
+annotation class JvmExposedBoxed(val jvmName: String = "", val expose: Boolean = true)
```
ilmirus, line 103: Typo - `JvmExposedBoxed` instead of `JvmExposeBoxed`.

ilmirus (approved):

serras:
The experimental implementation of this feature shall appear in Kotlin 2.2.0.

PR #396, opened 2024-10-30
Clarify interaction between context parameters and smart casting

serras:


proposals/context-parameters.md
```diff
@@ -528,6 +528,16 @@ context(console: ConsoleLogger, file: FileLogger) fun example4() =
   with(console) { logWithTime("hello") }  // no ambiguity, uses 'console'
 ```
 
+**§7.4bis** *(context resolution, smart casting)*: During context resolution any smart cast over available contexts should be considered. For example, the following code should be accepted:
```
cypressious, line 531: What does `bis` mean?

serras: (Actually, it seems this is only done in [Spanish-speaking countries](https://www.proz.com/kudoz/english-to-spanish/law-general/2653430-bis-tris-quarter.html), other places use `7.4b` instead).

cypressious (approved):

PR #401, opened 2024-11-20
Improvements to annotation use-site targets on properties

serras:
Only comment here about the text itself. The discussion of the feature is done in #402.


hossen71 (commented):
76048091a5d098f3517e9a35607d9adfcc83d36f

PR #405, opened 2024-12-20
Nested (non-capturing) type aliases

serras:
Only comment here about the text itself. The discussion about the feature is done in #406.

Right now type aliases can only be used at the top level. The goal of this document is to propose a design to allow them within other classifiers, in case they do not capture any type parameters of the enclosing declaration.

### DIFF
--- a/proposals/nested-typealias.md
+++ b/proposals/nested-typealias.md
@@ -161,7 +161,7 @@ class Example<T> {
     // = capture(Inner<Int>, capture(Example<S>, { T }))
     // = capture(Inner<Int>, { S })
     // = capture(Int) + { S } = { S } ⊆ { S } => OK
-    typelias Boo<S> = Example<S>.Inner
+    typealias Boo<S> = Example<S>.Inner
 }
 ```
 
--- a/proposals/nested-typealias.md
+++ b/proposals/nested-typealias.md
@@ -1,4 +1,4 @@
-# Nested (and inner) type aliases
+# Nested (non-capturing) type aliases
 
 * **Type**: Design proposal
 * **Author**: Alejandro Serrano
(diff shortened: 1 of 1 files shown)


proposals/nested-typealias.md
```diff
@@ -0,0 +1,241 @@
...
+    // capture(Inner<Int>, { T })
+    // = capture(Int) + { T }
+    // = { T } ⊈ { T } => not allowed
+    typealias Moo = Inner<Int>
+
+    // capture(Example<S>.Inner<Int>, { T })
+    // = capture(Inner<Int>, capture(Example<S>, { T }))
+    // = capture(Inner<Int>, { S })
+    // = capture(Int) + { S } = { S } ⊆ { S } => OK
+    typelias Boo<S> = Example<S>.Inner
```
lukellmann, line 164:
```suggestion
    // capture(Example<S>.Inner<Int>, { T })
    // = capture(Inner<Int>, capture(Example<S>, { T }))
    // = capture(Inner<Int>, { S })
    // = capture(Int, { S }) + { S } = { S } ⊆ { S } => OK
    typelias Boo<S> = Example<S>.Inner<Int>
```
`capture` has two "parameters", `Inner` has a type parameter

proposals/nested-typealias.md
```diff
@@ -0,0 +1,241 @@
...
+
+// incorrectUsage.kt
+val i = A().I()    // ⚠️ `I` lives in the static scope of `A`
+val d = A().C.D()  // ⚠️ cannot use `C.D()` to refer to a function
+
+// correctUsage.kt
+import A.*  // imports `I`
+import C.*  // imports `D`
+
+val d = A().D()
```
lukellmann, line 196: it seems like this is missing the usage of `I`:
```kt
val i = A().I()
```

proposals/nested-typealias.md
lukellmann: this one is still missing `<Int>` for `Inner`

Issue #382, opened 2024-07-23
Uuid


qurbonzoda:
This issue is for discussion of the proposed `Uuid` API.

The API is already available in Kotlin 2.0.20-Beta2 and later, and is marked with `@ExperimentalStdlibApi`.

axelfontaine:
As Uuid really is a base building block of data models, the package dependencies really matter. For the JVM, instead of having uuid depend on nio, it should really be the other way around!

lukellmann:
> As Uuid really is a base building block of data models, the package dependencies really matter. For the JVM, instead of having uuid depend on nio, it should really be the other way around!

If I understood this correctly, only the convenience extension functions for reading from / writing to `ByteBuffer`s depend on NIO, not the `Uuid` class itself.

joffrey-bion:
@axelfontaine also, what exactly do you mean by "package dependencies"? It's not like there is a maven dependency, and Java 9 modules are at the level of the entire module (the entire stdlib?) not the individual `kotlin.uuid` package, so I'm not sure what exact problem you're worried about here.

adam-enko:
I would prefer the name to be `UUID` over `Uuid`.

> It is consistent with existing names `Base64.UrlSafe` and `Base64.Mime` names.

Mime and UUID are both abbreviations, but the former is an acronym, and the later is an initialism.

- 'Mime' is an acronym, and is pronounced as a single word.  JSON is another acronym.
- 'UUID' is an initialism, because each letter is pronounced individually. So is URL.
   (Although maybe I'm wrong and people do say 'oooo-eye-dee' and 'uhrrrll'(!)). 

I can see how `UrlSafe` could be a counterpoint (should it be `URLSafe`?), but it follows [the official Kotlin naming conventions](https://kotlinlang.org/docs/coding-conventions.html#choose-good-names).

> When using an acronym as part of a declaration name, capitalize it if it consists of two letters (`IOStream`); capitalize only the first letter if it is longer (`XmlFormatter`, `HttpInputStream`).

The visual difference between UUID/Uuid is confusing when looking at the Kotlin/Java converter helper:

```
public fun Uuid.toJavaUuid(): java.util.UUID
```

Shouldn't the extension `fun Uuid.toJavaUuid()` instead be named `fun Uuid.toJavaUUID()`, to match the Java naming scheme?

> * It helps disambiguate whether the Java or Kotlin UUID type is being used without looking at imports.

[...]

PR #408, opened 2025-02-02
UUID version 5 and version 7 support

Dunemaster:
In this PR I propose adding support for UUID version 5 and version 7.

Please, consider the KEEP.

If the proposal is green-lighted I wll go ahead with prototype implementation.

There is already an issue for it: https://youtrack.jetbrains.com/issue/KT-74411

### DIFF
--- a/proposals/stdlib/uuid-additional-versions.md
+++ b/proposals/stdlib/uuid-additional-versions.md
@@ -221,6 +221,15 @@ The factory method for Uuid generators will look something like this:
    fun uuidV7Generator(): () -> Uuid
 ```
 
+Example APIs using explicit names:
+* [Writer](https://github.com/JetBrains/amper/blob/fffc216664588c7a369064c157cbc33f68826ed9/sources/concurrency/src/org/jetbrains/amper/concurrency/fileUtils.kt#L23)
+* [Comparator ](https://github.com/JetBrains/kotlin/blob/f0bc24bcbb6d4c3dc4611547c93f48bbaca04094/libraries/stdlib/wasm/src/kotlin/Comparator.kt#L11)
+* [History Listener](https://github.com/JetBrains/EzArgs/blob/74b56a033141d2388ce86a72b56776d6fc8c6fd4/src/main/kotlin/com/jetbrains/rider/ezargs/services/EzArgsService.kt#L13)
+
+Example APIs not using explicit names:
+* generateSequence https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.sequences/generate-sequence.html
+
+
 The current approach was selected mainly because of:
 
 * `UuidGenerator` interface is declared a functional interface, so consumers can also use `() -> Uuid` in their parameters
--- a/proposals/stdlib/uuid-additional-versions.md
+++ b/proposals/stdlib/uuid-additional-versions.md
@@ -89,9 +89,9 @@ an object represents the stateful nature of UUID generation better. It is also m
 * Provide a pooling implementation (see UUID pooling).
 
 ```kotlin
-    public interface UuidGenerator {
+    fun interface UuidGenerator {
     
-        public fun generate() : Uuid
+        fun generate() : Uuid
         
    }
 ```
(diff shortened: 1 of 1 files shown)


proposals/stdlib/uuid-additional-versions.md
```diff
@@ -0,0 +1,233 @@
...
+
+It is proposed to support two additional version of UUID as specified in RFC 9562(https://www.rfc-editor.org/rfc/rfc9562.txt).
+
+The earlier proposed v6 version is not included, as it was never widely adopted and fully superseded by v7.
+
+### Proposed API
+
+#### UUID generator interface
+
+A new interface, representing a UUID generator is added. Although a pure function-based API is often simpler to use,
```
bcmedeiros, line 84: this is true, I end up creating generators in my code anyway, but I wonder if this should be addressed by the standard library or not. People might think of different ways to create generators, so to introduce these interfaces in the stdlib I'd expect the language team to require a more thorough analysis of what kinds of generators we have out there, what parameters will be part of the generator state and which ones will be parameters of the generate call (e.g should the Random instance be provided at the generator level or maybe be a optional parameter in the generate call? What about the time seed?)

At the end of the day I feel like we will go the path of having elemental methods in the companion objects, and let people create generators in their cod (my opinion).

Dunemaster: > Why add this at all? `() -> Uuid` exists. Just declare functions, and people can pass them as lambdas on their own.

That will be quite a lot of functions, to accomodate the plethora of possible UUID creation methods. And customisations by user (e.g. custom random number generators) will require even more complex signatures, accepting other functions and stateful objects.

Dunemaster: @bcmedeiros , I like generators because they provide a lot of flexibilty. I started with rather simple generators in the proposal, but they can be easily extended to accomodate additional parameters like thouse you have mentioned  (I think, those parameters should me passed in generator constructors).

Dunemaster: @Laxystem , I made `NameGenerator `a functional interface, so consumers can also use` () -> Uuid`

Laxystem: > @Laxystem , I made `NameGenerator` a functional interface, so consumers can also use `() -> Uuid`

But again, why do you even *need* a generator interface?

Dunemaster: > > @Laxystem , I made `NameGenerator` a functional interface, so consumers can also use `() -> Uuid`
> 
> But again, why do you even _need_ a generator interface?

This is more an style or convention thing.  

As the generator is a possibly long-lived, stateful object I would prefer to have a separate name for it. Using just a function signature suggests (at least for my intuition) a stateless function without side effects.

And, again, this thing obviously has a name already - a generator! So why do not use it in code if it is used  when discussing the concept?

Laxystem: @Dunemaster because adding to the stdlib has a lot of implications. If `Uuid` has a generator, that means every other class needs a generator as well!

What use case do you have for including this in the stdlib? What problem does it solve?

Besides, generating a UUID *is* stateless; all you need is a `Random` and an `Instant` as input for most variants.

Also note, that `buildSequence` exists. It's stateful, and yet, it does not use an interface.

proposals/stdlib/uuid-additional-versions.md
```diff
@@ -0,0 +1,263 @@
...
+    val uuidV7_1 = uuidV7Generator.generate()
+    val uuidV7_2 = uuidV7Generator.generate()
+}
+
+```
+
+## Dependencies
+
+There should be no new dependencies for JVM platform.
+TODO for other runtimes.
```
fzhinkin, line 179: Would it be possible to generate v5 UUIDs on platforms other than JVM?
For example, on Wasm/WASI?

Dunemaster: @fzhinkin the main challenge would be finding a suitable SHA-1 implementation.
A quick search shows that there are widely used implementations with compatible licenses.

For example, https://github.com/RustCrypto/hashes can be used for WebAssembly.

P.S. This opens the very interesting topic of having crypto primitives in Kotlin stblib. IMHO, there are very much needed

fzhinkin:
@Dunemaster, I appreciate you published the proposal. However, the proposed changes are not a completely new API, and are rather an extension of the existing `Uuid` API. The corresponding KEEP (#383) was not merged yet, so the best way to contribute into its design is discussing it in #382.

Regarding the UUIDv7, we (Kotlin Libraries team) discussed possible API shape internally (as part of https://youtrack.jetbrains.com/issue/KT-74411) and the original KEEP (#383) will be updated soon with the proposed design (which is, of course, a subject to further discussion).

proposals/stdlib/uuid-additional-versions.md
Dunemaster: > At the moment, Libraries team won't be able to drive that process. So, unfortunately, chances of including the SHA1 implementation are diminishingly small.

I would not suggest a home-grown implementation. Rather, I would use a well-established library similar to how Firefox uses NSS, OpenSSH uses LibreSSL and plethore of other projects use BouncyCastle

Issue #409, opened 2025-02-04
Subtyping Reconstruction


e2e4b6b7:
This issue is for discussion of the proposal to introduce Subtyping Reconstruction into the Kotlin language.
The full proposal text is available here: [proposals/subtyping-reconstruction.md](https://github.com/e2e4b6b7/KEEP/blob/subtyping-reconstruction/proposals/subtyping-reconstruction.md)

The KEEP presents the Subtyping Reconstruction technique that introduces smart casts for generics.

Please, use this issue for the discussion on the substance of the proposal. For minor corrections to the text, please open comment directly in the PR: #410.

kyay10:
Just clarifying, this would also be fine, right?
```
fun <T> evalWhen(e: Expr<T>): T = when (e) {
    is IntLit -> 42
}
```

kyay10:
From "Can we fix these problems":

> Fundamentally, this means we may need to add ways to encode recursive types (aka µ types) or some other ways to propagate such complex (e.g., recursive) constraints in the type system.
> 
> We leave the solutions to these problems for possible future work.

Would introducing Shape interfaces do the trick here? Just food for thought

e2e4b6b7:
@kyay10,

> Just clarifying, this would also be fine, right?
> 
> ```
> fun <T> evalWhen(e: Expr<T>): T = when (e) {
>     is IntLit -> 42
> }
> ```

Yes, it works in the same way as with any other Int-typed expression on the rhs. More precisely, in this case it is: `T :> Int :> ILT` and everything is fine.

> Would introducing Shape interfaces do the trick here?

Yes, inferring only material interfaces will simplify everything, but there are some other issues. Therefore, we decided to address them separately to keep things manageable.

RossTate:
[...]

PR #410, opened 2025-02-04
Subtyping reconstruction

e2e4b6b7:
Only comment here about the text itself. The discussion about the feature is in #409.

### DIFF
--- a/proposals/subtyping-reconstruction.md
+++ b/proposals/subtyping-reconstruction.md
@@ -1,0 +1,1254 @@
+# Subtyping reconstruction
+
+* **Type**: Design proposal
+* **Authors**: Roman Venediktov, Daniil Berezun
+* **Contributors**: Marat Akhin, Mikhail Zarechenskiy, Alejandro Serrano Mena
+* **Discussion:** [KEEP-409](https://github.com/Kotlin/KEEP/issues/409)
+* **Status**: In discussion
+* **Related YouTrack issue**: [KT-18510](https://youtrack.jetbrains.com/issue/KT-18510/GADT-style-smart-casts)
+
+## Table of Contents
+
+- [Introduction](#introduction)
+  - [Motivation](#motivation)
+  - [Real-world examples](#real-world-examples)
+  - [Broad overview](#broad-overview)
+    - [Constrained subclasses](#constrained-subclasses)
+    - [Scope of the constraints](#scope-of-the-constraints)
+    - [Outline](#outline)
+    - [Limitations](#limitations)
+  - [More advanced use-cases](#more-advanced-use-cases)
+    - [Run-time subtyping evidence](#run-time-subtyping-evidence)
+    - [Type-safe extensions](#type-safe-extensions)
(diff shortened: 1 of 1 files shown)


LPTK:
Hi Roman, glad to see that Kotlin is moving in this direction!

I was pointed to your proposal by a student and was curious whether you were aware of our recent paper on the topic: [A case for DOT: Theoretical Foundations for Objects With Pattern Matching and GADT-style Reasoning](https://2022.splashcon.org/details/splash-2022-oopsla/79/A-case-for-DOT-Theoretical-Foundations-for-Objects-with-Pattern-Matching-and-GADT-St).

While our paper was focused on the soundness of subtyping reconstruction rather than on the algorithm, you might find it interesting because it basically shows how we ended up _thinking_ about the feature in the context of Scala 3, while implementing it. The paper also mentions Kotlin and elucidates important differences between Kotlin and Scala 3 with respect to subtyping reconstruction.

This is also the paper where we came up with the term "Subtyping Reconstruction", which you seem to have adopted! So I was a bit surprised not to find it in your references section.

e2e4b6b7:
Hi Lionel (@LPTK),

Thank you for your comment. We are indeed aware of that work and have drawn significant inspiration from it. You'll find it referenced as “[Description of the Scala 3 implementation](https://dl.acm.org/doi/pdf/10.1145/3563342) (Section 6.2)” in the second position of our references list.

We chose to emphasize the algorithm section, as we believe it may be interesting to more readers. Please note that we did not format it extensively, as it is not typical for KEEP documents to have references list and since it isn’t indexed elsewhere. However, if you’d prefer a specific format, we’re happy to adjust it.

Also, if you have any concerns about the soundness of our approach, we’d be glad to hear your thoughts!

LPTK:
Thanks for your quick response, Roman!

> You'll find it referenced as “Description of the Scala 3 implementation (Section 6.2)” in the second position of our references list.

Oh, I had not realized that link (which I had not clicked) was pointing to our paper, my bad! I had assumed it pointed to some internal discussions of the Dotty/Scala 3 compiler project.

> if you’d prefer a specific format, we’re happy to adjust it

Not at all; I just wanted to check you were aware of the paper 😉 

> Also, if you have any concerns about the soundness of our approach, we’d be glad to hear your thoughts!

For this I'll need to take some time to sit down and look into the details for a while, which won't happen in the near future unfortunately (too many deadlines these days). But it would help if you could come up with a formalization or paper draft at some point.

proposals/subtyping-reconstruction.md
```diff
@@ -0,0 +1,1254 @@
+# Subtyping reconstruction
+
```
hossen71, line 2: UQA5NWl5eH6KvxwnMBCj7vDbb-Xebok-pWGfFuHhq1zMxmv5

hossen71: d6e588135615591e79ebc0e086717ba522a8fc78

nikitabobko:

Issue #412, opened 2025-03-05
Unused return value checker


sandwwraith:
This is an issue to discuss **Unused return value checker** proposal. The text of the proposal is located [here](https://github.com/Kotlin/KEEP/blob/underscore-for-unused-local/proposals/unused-return-value-checker.md).

This proposal depends on the other sub-proposal, **Underscore for unused local variables**, located [here](https://github.com/Kotlin/KEEP/blob/underscore-for-unused-local/proposals/underscores-for-local-variables.md).
Since both proposals are connected, we'll discuss them together.

The goal of the proposals is to enhance the functionality of the existing 'unused expression' diagnostic in the Kotlin compiler.
New diagnostic should be able to report complex expressions whose return values are meaningful (e.g., non-Unit) but are not used.

kyay10:
You mention that "class or top-level properties" cannot be unnamed. This makes sense for class properties, since one can use an `init` block. However, there's no `init` option for the top-level. I would dislike having unnamed top-level properties, but I also dislike that such code:
```
val unused = run {
  initialiseSomethingStatic()
}
```
may be needed sometimes. What are your thoughts on this?

kyay10:
You also mention that "Unnamed variables cannot be delegated". I have a very clear use case for this though: triggering `provideDelegate` without using the resulting delegate object. An almost-example of this is Gradle sourceSets:
```
val myCustomSourceSet by creating {
  ...
}
```
When one won't refer to `myCustomSourceSet` afterwards, it might be helpful to have:
```
val _ by creating {
  ...
}
```
Except that, in the case of Gradle, the name of a property is used. Maybe such properties should have an empty name to signal that they're unused?

sandwwraith:
@kyay10 I would say that using top-level code for static initialization may be bad practice and lead to hard-to-debug errors because such code is actually executed only when the corresponding file facade class is loaded, which may happen at different points in time. But if I had to do this, my solution would be to wrap `init {}` into `object StaticInits { ... }`.

sandwwraith:
@kyay10 Can you explain what is the use-case for creating a source set, but not referring to it afterwards?

hrach:
Could you please evaluate the argumentation behind this?

> ```kotlin
>  foo() ?: return // foo() is considered unused because its value is lost after comparison with null.`
> ```

The value IS USED to check if it is null or not. 

Isn't the rewrite a useless noise?

```Kotlin
val _ = foo() ?: return
```

-----

[...]

PR #413, opened 2025-03-05
KEEPs for Unused return value checker and Underscore for unused local variables

sandwwraith:
Discussion thread: #412


nikitabobko:

PR #414, opened 2025-03-06
Fix broken internal link in coroutines.md (coroutine builders section)

easternkite:
In `coroutines.md`, the internal link to the "coroutine builders" section was incorrectly written as `[coroutine builders](coroutine-builders)`,  
which does not work because it lacks the `#` symbol required for internal navigation.

### Fix:
- Changed `[coroutine builders](coroutine-builders)` → `[coroutine builders](#coroutine-builders)`.
- This ensures proper navigation within the same document.

Tested and confirmed that the link now correctly redirects to the intended section.

Let me know if any further modifications are needed. Thanks!


fzhinkin (approved):
@easternkite thank you for fixing it!

PR #415, opened 2025-03-08
Proposal: Add URI Type to Kotlin Standard Library

iseki0:


gildor:
It would be great to have it in the standard library. But I feel that it must be both URI and URL, it's very useful to know that it's an actual URL and it mostly should be used everywhere instead of more generic URI which rarely used correctly

proposals/stdlib/uri.md
```diff
@@ -0,0 +1,64 @@
...
+
+**Author**: iseki zero  
+**Status**: Pending  
+**Submitted**: 2025-03-09  
+
+---
+
+## Summary
+
+This proposal advocates for introducing a first-class `URI` type to the Kotlin Standard Library.
```
gildor, line 11: It's against Kotlin naming convention, it must be called Uri (see for example recent UUID KEEP)

iseki0: Sounds great. I will fix it soon.

iseki0:
OK, maybe we can add both `Uri` and `Url` into the stdlib

iseki0:
@gildor What do you think about add IP address types? Since the URI/URL might use IP address in the authority segment.

proposals/stdlib/uri.md
```diff
@@ -0,0 +1,64 @@
...
+      path {
+         append("kotlin")
+         append("multiplatform")
+      }
+      // other representation of path-adder...
+   }
+}
+```
+
+## Sample implementation
```
gildor, line 55: I think it's important to have this section, at least cover differences between Java and Android Uri class.

Also see how other languages do this

gildor:
@iseki0 I personally never had use case for this, but I'm curious about support in other languages. Distinguish between ipv4/v6 could be useful, but I feel that it's not a job for URL/uri to give IP information, but maybe I'm wrong

proposals/stdlib/uri.md
iseki0: Yes, I'm in writing a library as reference implementation.

iseki0:
> Distinguish between ipv4/v6 could be useful, but I feel that it's not a job for URL/uri to give IP information

Maybe you're right, it should be in network library

iseki0:
@gildor Thank you for your reviewing, in the past serval days, I added more content into the proposal. Please review it again.
Thank you

iseki0:
YouTrack: https://youtrack.jetbrains.com/issue/KT-76022/Proposal-Add-URI-Type-to-Kotlin-Standard-Library

Issue #421, opened 2025-05-19
Build Tools API


ALikhachev:
This is an issue to discuss the Build Tools API proposal. The text of the proposal is located [here](https://github.com/Kotlin/KEEP/blob/build-tools-api/proposals/extensions/build-tools-api.md).

CLOVIS-AI:
This looks very good!

> A special note about Kotlin/Native: currently, Kotlin/Native has a completely separated distribution that is to be supplied by a build system. It can be bundled into BTA, or the provisioning may be encapsulated in it to simplify the integration. The exact way of simplification is an implementation detail.

Could it be possible to clarify in the KEEP that **anytime there is a setup step**, there should be a way for the build tool to do it separately or not do it at all. For example, the downloading of the Kotlin Native toolchain, the downloading of Node when using JS, etc. It's fine if the BTA implementation does it transparently by default, but there should be a way for build tools to explicitly take responsibility of these steps themselves, without opting out of everything else the BTA offers. For example, because a build tool has better caching capabilities, because internet connection requires specific configuration that the Kotlin daemon won't have, etc.

Amejonah1200:
Will this Build Tools API be able to allow integration into an LSP Server? Such as, it is trivially possible to implement an LSP Server using this API?

ALikhachev:
> Could it be possible to clarify in the KEEP that **anytime there is a setup step**, there should be a way for the build tool to do it separately or not do it at all. For example, the downloading of the Kotlin Native toolchain, the downloading of Node when using JS, etc. It's fine if the BTA implementation does it transparently by default, but there should be a way for build tools to explicitly take responsibility of these steps themselves, without opting out of everything else the BTA offers. For example, because a build tool has better caching capabilities, because internet connection requires specific configuration that the Kotlin daemon won't have, etc.

Thanks for feedback! Sure, we are keeping that in mind. We aim to provide ways to delegate some work to build systems. Another example would be detecting touched files for incremental compilation: BTA is capable of doing it, but it can also rely on information provided by the build system instead, if the build system does this anyway.

hfhbd:
Do you plan to add [Build Server Protocol](https://build-server-protocol.github.io) support or would this be an abstraction on top of  BTA?

Tapchicoma:
[...]

PR #422, opened 2025-05-19
Introduce Build Tools API

ALikhachev:
Discussion thread: #421


nikitabobko:

PR #425, opened 2025-06-06
Static members and type extensions

serras:
Only comment here about the text itself. The discussion about the feature is done in #427.

We propose to bring the idea of _statics_ in Kotlin with _static members_ and _type extensions_. 
The underlying idea is surfacing the notion of _static scope_ more clearly in the language.


nikitabobko:

PR #436, opened 2025-07-02
Name-based + new position-based destructuring

serras:
Only comment here about the text itself. The discussion about the feature is done in #437.

We propose a new form of _destructuring_ based on _names_ instead of positions. In addition, we propose new syntax for _position_-based destructuring.

```kotlin
(val age, val name = fullName) = person
```


nikitabobko:

PR #443, opened 2025-07-14
Adjust the KEEP according to Dokka / Analysis API design discussions

BlaBlaHuman:


BlaBlaHuman:
Hi, please pay close attention to all the examples and details in the tag-sections chapter. I tried to gather all the information on them I could from our discussions and from the original KEEP, however, I might have misunderstood something.

And I'm still not quite sure about the details of this multi-resolution. In the KEEP I mentioned that we should return all the symbols the compiler could resolve this reference to (i.e. all symbols from some scope), not just symbols of the same kind. It seems like a valid and requested feature, and there won't be too many symbols with the same name in the same scope anyways. It's probably more important than just handling overloads by returning multiple symbols of the same kind.

```kotlin
/**
 * @see Foo - now it's able to point to the function
 */
interface Foo 

fun Foo(): Foo = object : Foo {}
```

nikitabobko:
> [BlaBlaHuman](https://github.com/BlaBlaHuman) wants to merge 1 commit into [kdoc/Streamline-KDoc-ambiguity-references](https://github.com/Kotlin/KEEP/tree/kdoc/Streamline-KDoc-ambiguity-references) from [kdoc/omak/Streamline-KDoc-ambiguity-references](https://github.com/Kotlin/KEEP/tree/kdoc/omak/Streamline-KDoc-ambiguity-references)

The proposal is already merged to the `main` branch https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0389-kdoc-streamline-KDoc-ambiguity-references.md, so please update the proposal in the `main` branch, avoid merging non-main branches into each other

proposals/KEEP-0389-kdoc-streamline-KDoc-ambiguity-references.md
```diff
@@ -1,24 +1,29 @@
 ﻿# Streamline KDoc ambiguity links
 
-* **Type**: KDoc proposal
+* **Type**: Proposal
 * **Author**: Vadim Mishenev
 * **Status**: Submitted
```
nikitabobko, line 5: Can you also please update the status according to https://github.com/Kotlin/KEEP?tab=readme-ov-file#keep-document-lifecycle

marcopennekamp (commented):
Thanks for the comprehensive update! 🙌 

I left a lot of comments above, but I think my main points of contention are:

1. We need clear definitions for the key concepts, such as the "KDoc context", the exact steps of the resolution algorithm, and generally just more precise formulations for a lot of things.
2. I think the resolution algorithm could be described more clearly by mentioning the exact layers (self-links, scopes, global references, packages), essentially rearranging some sections a bit to arrive at a single, well-described algorithm.
3. We need to discuss property resolution and whether (1) properties should be resolved as first segments during multi-segment resolution and (2) legally accessing a property (e.g. `[name.length]`) in KDoc references is desired in the future.

And this definitely needs another round of copyediting, but let's get the specifics right first. 👍

proposals/KEEP-0389-kdoc-streamline-KDoc-ambiguity-references.md
```diff
@@ -280,257 +467,509 @@ class A(p: Int, val abc: String) {
...
+ */
+interface Foo 
+
+fun Foo(): Foo = object : Foo {}
+```
+
+It's important to preserve the priority of symbols in the resulting collection for purposes
+specified in [Handling resolved symbols on the use-site](#handling-resolved-symbols-on-the-use-site).
+
+## Handling resolved symbols on the use-site
```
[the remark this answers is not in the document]
nikitabobko: Sorry, I'm too much out-of-context for this proposal.

In general, it's ok to mention IDE and tooling integration in the language proposal. Sometimes the way the feature is toolable affects the design

proposals/KEEP-0389-kdoc-streamline-ambiguous-KDoc-links.md
```diff
@@ -0,0 +1,1752 @@
...
+```
+
+We can see that the constructor parameter `param` is not accessible in this context.
+It's not called "unresolved" as this link is a context declaration, so it's handled separately.
+
+Declarations don't have to be resolvable in the code as-is.
+It's contained in the KDoc context when there is a case this declaration is correctly referenced in the code.
+It might be a direct call, might be a reference, 
+but there must be a case when this declaration could be resolved in the code.
+```kotlin
```
vmishenev: TBH, this example is awkward:
It could be rewritten:
```kt
class A {
    fun myFun() {}

    /**
     * [myFunFromC] is visible and resolved
     */
    class B {
        fun myFakeFun(x: C) {
            C.myFunFromC() // Valid
            C::myFunFromC // Valid
        }
    }
}

class C {
    fun myFunFromC() {}
}

```

proposals/KEEP-0389-kdoc-streamline-ambiguous-KDoc-links.md
```diff
@@ -0,0 +1,1752 @@
...
+In this case, the relative name is `A.B.C` and it's relative to the `Global` class.
+
+The resolver has already calculated the list of visible scopes 
+when searching for short names in [Short names treatment](#short-names-treatment).
+If a relative link is resolvable, then it must be relative to any of these visible scopes.
+
+To reduce this multi-segment search to a single-segment one, 
+all these scopes have to be slightly modified.
+
+`A.B.C` implies that somewhere there is a chain of two nested classifiers `A` and `B` with `B`
```
vmishenev: What is a classifier? Can it be a typealias or a type parameter? 
```kt
/**
* [T.toString] - is resolved?
*/
class<T: Any> A {
}
```

proposals/KEEP-0389-kdoc-streamline-ambiguous-KDoc-links.md
```diff
@@ -0,0 +1,1752 @@
...
+containing some member `C`.
+The resolver should try to find a classifier `A` in each of these scopes and, if found,
+replace the original scope with the member scope of this class.
+If no suitable classes are found in some scope, this scope should be discarded.
+Then the resolver should take all the modified scopes and repeat the search for nested classifiers (`B` in this case)
+until just the short name `C` is left.
+Then it's just a single-segment name search in the constructed list of scopes.
+
+Note that currently it's prohibited to have a variable as a link segment.
```
vmishenev: Not only a variable, but also a function? Typealias?
`non-last/final link segment` instead of `a link segment`?

proposals/KEEP-0389-kdoc-streamline-ambiguous-KDoc-links.md
```diff
@@ -1024,6 +1152,343 @@ It makes the result consistent and reliable.
...
+    fun usage() {
+        A.B // Unresolved, A is resolved to a more local property
+    }
+}
+```
+
+#### Why this idea was rejected
+
+The main problem is that KDoc resolution has a larger scope than the language does.
+That's because KDoc links ignore all visibility / accessibility modifiers during the search.
```
marcopennekamp: Super nitpick: I agree that it's visibility and accessibility, but accessibility is more than just modifiers, imo. Maybe we can call this simply "visibility and accessibility" without the "modifiers"?

PR #454, opened 2025-09-11
[Design Notes] Refinement Types

InversionSpaces:
Research summary on Refinement Types for the Kotlin language.


ice-phoenix:
Could I ask you to resize the charts in the Survey part, as they are now, they are too big =) It seems we can do this by changing them to ``, for example.

PR #492, opened 2026-07-01
evgeniy.zhelenskiy/will-become-value

zhelenskiy:


proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,155 @@
...
+* **Discussion and feedback**: TODO
+
+# Summary
+
+This KEEP proposes the `@WillBecomeValue` annotation — a Kotlin equivalent of Java's
+`@jdk.internal.ValueBased` from [JEP 390](https://openjdk.org/jeps/390). The original Java annotation marks existing
+reference-based classes whose identity semantics should not be relied upon, because they are
+intended to become value classes in after the [Valhalla](https://openjdk.org/projects/valhalla/) project release.
+
+The annotation is necessary for the gradual migration process. However, the original Java
```
ice-phoenix: I feel that the "However" part is a bit sudden and disconnected for the uninitiated aka a reader not following the work on MFVC / Valhalla closely does not understand why this "However" is important.

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,155 @@
...
+
+# Motivation
+
+## Background: value classes in Kotlin
+
+Kotlin supports `value class` — a class modifier that strips object identity, enabling more safety and effeciency. Value classes guarantee that two
+instances with the same state are indistinguishable: there is no reliable way to distinguish
+them by reference equality (`===`), `System.identityHashCode`, or synchronization. Thus, such usages are forbidden for them.
+
+However, the process of migrating existing classes to `value class` is a breaking change for two reasons:
```
ice-phoenix: 2 reasons and list with 4 items. Something is off.

ice-phoenix: I think you are conflating here two kinds of migrations -- existing classes to full value classes and existing inline value classes to full value classes, and we actually care about only one of them w.r.t. `@ValueBased`

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,155 @@
...
+1. The mentioned restrictions
+2. ABI change on JVM (in case of the existing `@JvmInline` value classes)
+3. Currently existing `@JvmInline` value classes support only a single field.
+4. Currently existing `@JvmInline` value classes cannot be abstract/sealed.
+
+The second problem is not a problem for full value classes,
```
ice-phoenix: I feel we are missing a short intro of inline and full value classes somewhere in the KEEP.

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,155 @@
...
+  should be marked `value`, but it cannot be done because of the compatibility constraint, usage restrictions,
+  field number restriction, and others.
+- The Kotlin compiler and tooling can then warn at use sites where identity-sensitive operations
+  are applied to annotated instances, giving developers time to fix their code.
+- When the class is eventually migrated to `value class`, the change will be much less likely
+  to cause silent regressions.
+
+## Naming
+
+The annotation is named **`@WillBecomeValue`**. The name was chosen over three other candidates:
```
ice-phoenix, line 157: We should discuss the name at some point once again :hehe:

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+## Semantics
+
+A class annotated with `@WillBecomeValue`:
+
+1. **Must not be relied upon for identity.** Reference equality (`===`), `identityHashCode`,
+   and synchronization on instances are considered undefined behavior and may be flagged by
+   the compiler or runtime.
+2. **Must be immutable (or effectively immutable).** Mutating state through shared references
```
ice-phoenix: Shallow immutable?

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,155 @@
...
+   undermines the value semantics the annotation promises.
+3. **Is a candidate for future migration to `value class`.** Library authors are not obligated
```
ice-phoenix: I'd say that's another reason why `@WillBecomeValue` is not such a great name, if the migration is never completed, the name misdirects. `@ValueBased` is better in this regard.

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,155 @@
...
+The annotation can be applied to:
+- Regular (reference) final classes.
+- Abstract classes and sealed classes, intended as base types.
+
+It **cannot** be applied to:
+- Classes already declared as `value class`.
+- Interfaces (identity is not a concept that applies directly).
+- `object` declarations (singletons intentionally have identity).
+- Enums (enumerations cannot become value types).
+- Open classes (value classes are not going to be ever supported to be open).
```
ice-phoenix: I would not say "ever", let's tone it down a bit?

ice-phoenix (changes requested):
I will do a second pass tomorrow, here's the first immediate feedback.

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,155 @@
...
+* **Author**: Evgeniy Zhelenskiy
+* **Status**: Proposed
+* **Discussion and feedback**: TODO
+
+# Summary
+
+This KEEP proposes the `@WillBecomeValue` annotation — a Kotlin equivalent of Java's
+`@jdk.internal.ValueBased` from [JEP 390](https://openjdk.org/jeps/390). The original Java annotation marks existing
+reference-based classes whose identity semantics should not be relied upon, because they are
+intended to become value classes in after the [Valhalla](https://openjdk.org/projects/valhalla/) project release.
```
serras:
```suggestion
intended to become value classes once the [Valhalla](https://openjdk.org/projects/valhalla/) project releases.
```

proposals/KEEP-0465-will-become-value.md
serras: I would go as far as removing the part about the annotation being internal. We could later say that we cannot `typealias` is in the JVM because it's internal, but regardless of that fact, we still need a new one.

proposals/KEEP-0465-will-become-value.md
serras: If possible, I would avoid saying that we're doing something because Java is doing it.

```suggestion
This KEEP proposes the `@WillBecomeValue` annotation to mark existing reference-based classes
whose identity semantics should not be relied upon. That prepares existing code for the transition
to a [multi-field value class](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0454-better-immutability-value-classes-MFVC.md). In particular, developer can start receiving warnings about
wrong usages of those classes _now_, instead of waiting for them to become full value classes.
```

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,155 @@
+# `@WillBecomeValue` annotation
+
+* **Type**: Design proposal
+* **Author**: Evgeniy Zhelenskiy
+* **Status**: Proposed
+* **Discussion and feedback**: TODO
+
+# Summary
```
ice-phoenix: I would suggest for the summary to be a bit more descriptive about why and what `@WillBecomeValue` does. Ideally, after reading the summary I should understand 1) what it is about 2) how it does things 3) whether I should care about it.

Also, a ToC after would be appreciated.

proposals/KEEP-0467-will-become-value.md
```diff
@@ -0,0 +1,155 @@
...
+The third and the fourth problems are not problem for full value classes as well,
+since the full `value class`es allow multiple fields and can be abstract/sealed.
+
+The standard library already contains classes that behave like values — they are immutable,
+have no meaningful identity, and are candidates for eventually becoming `value class`. Examples
+include `Pair`, `Triple`, and various result/wrapper types.
+
+## The migration problem
+
+Turning an existing reference class into a `value class` is a **breaking change**:
```
ice-phoenix: Considering you are covering the inheritance-based limitations also, maybe it makes sense to mention those as breaking changes as well?

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,155 @@
...
+toward value types under Project Valhalla, while giving users time to fix problematic usage.
+
+Kotlin already supports the annotation: as tracked in [KT-70722](https://youtrack.jetbrains.com/issue/KT-70722),
+the compiler recognizes `@jdk.internal.ValueBased` on JDK classes and issues warnings for
+identity-sensitive operations on their instances. `@WillBecomeValue` extends this mechanism
+to Kotlin-defined classes.
+
+## Kotlin's need for an equivalent
+
+Kotlin needs the same mechanism for its own migration path:
```
ice-phoenix: KMP looks like yet another major reason to get `@WillBecomeValue` aka other platforms

proposals/KEEP-0465-will-become-value.md
ice-phoenix: Also, will we do warnings when one tries to apply `@WillBecomeValue` to smth with a backing field `var`?

ice-phoenix: I guess there's a more general question of: do we want to ensure that, if one marks their class with `@WillBecomeValue` and there are no warnings, once it is changed to `value` there are also no warnings?

proposals/KEEP-0465-will-become-value.md
zhelenskiy: Why would they need to do it? IMHO, it's actually a great example of the annotation's misuse.

proposals/KEEP-0465-will-become-value.md
zhelenskiy: Thanks, I'll write about it.

proposals/KEEP-0467-will-become-value.md
zhelenskiy: They are inline-class-specific. They are mentioned below.

proposals/KEEP-0465-will-become-value.md
zhelenskiy: My answer is yes and I explained it now

proposals/KEEP-0468-will-become-value.md
zhelenskiy: /here we go again/

proposals/KEEP-0465-will-become-value.md
zhelenskiy: Fixed it

proposals/KEEP-0465-will-become-value.md
zhelenskiy: Agree

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+
+# Abstract
+
+This KEEP proposes the `@WillBecomeValue` annotation to mark existing reference-based classes
+whose identity semantics should not be relied upon. That prepares them for migration
+to [full value classes](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0454-better-immutability-value-classes-MFVC.md).
+
+In particular, by using the annotation a developer initiates reporting warnings about usages of identity for those classes in the downstream code,
+allowing not to break with errors the downstream code immediately by marking the classes as `value class`es.
+The annotation is necessary for the gradual migration process.
```
ice-phoenix:
```suggestion
In particular, by using this annotation a developer helps their downstream users to not break their code when the class is changed to a value class, as those users will get migration warnings about usages which will stop being valid.
Without this mechanism, the adoption of full value classes would be a slower and harder process.
```

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+  * [Compiler warnings](#compiler-warnings)
+  * [Applicability](#applicability)
+  * [Interaction with value classes](#interaction-with-value-classes)
+* [References](#references)
+
+# Motivation
+
+## Background: value classes in Kotlin
+
+Kotlin supports `value class` — a class modifier that strips object identity, enabling more safety and effeciency. Value classes guarantee that two
```
ice-phoenix:
```suggestion
Kotlin supports `value class` — a class modifier that strips object identity, enabling more safety and efficiency. Value classes guarantee that two
```

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+  * [Interaction with value classes](#interaction-with-value-classes)
+* [References](#references)
+
+# Motivation
+
+## Background: value classes in Kotlin
+
+Kotlin supports `value class` — a class modifier that strips object identity, enabling more safety and effeciency. Value classes guarantee that two
+instances with the same state are indistinguishable: there is no reliable way to distinguish
+them by reference equality (`===`), `System.identityHashCode`, or synchronization. Thus, such usages are forbidden for them.
```
ice-phoenix, line 42: I'm not sure that distinguishing by synchronization is even possible.

I would try to explain the two connected, but different things: that we take identity away and what that means (value) semantically, and what one cannot do w/o identity.

ice-phoenix: Between this and the next paragraph, I feel we are missing a bridge which explains why people should even care aka why many data should be represented as value types.

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+instances with the same state are indistinguishable: there is no reliable way to distinguish
+them by reference equality (`===`), `System.identityHashCode`, or synchronization. Thus, such usages are forbidden for them.
+
+Kotlin currently supports inline `value class`es which are a subset of `value class`es that have a single underlying field.
```
ice-phoenix:
```suggestion
Kotlin currently supports a very limited subset of `value class`es: inline `value class`es with a single underlying field.
```

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+
+Kotlin currently supports inline `value class`es which are a subset of `value class`es that have a single underlying field.
+Their main purpose is to create type-safe wrappers around existing types being transparent in the runtime.
+Many libraries and frameworks (`kotlinx.serialization`, `Spring`) adopted the usage and embed the underlying field, keeping safe wrapper only on the source code level.
```
ice-phoenix, line 62: I'm not sure this is even needed.

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+Such a behavior is not acceptable for general purpose value classes.
```
ice-phoenix, line 63: I'd rephrase this more towards "current value classes in Kotlin are very limited, but people need them for many more cases => many classes which should have been `value` are not marked as such".

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+Their main purpose is to create type-safe wrappers around existing types being transparent in the runtime.
+Many libraries and frameworks (`kotlinx.serialization`, `Spring`) adopted the usage and embed the underlying field, keeping safe wrapper only on the source code level.
+Such a behavior is not acceptable for general purpose value classes.
+Due to this and other restrictions, far not all existing classes can be marked as `value class` while only inline `value class`es are supported.
+
+There are plans to support [full `value class`es](https://github.com/Kotlin/KEEP/blob/69d675e8a15f66ff6b3dace70b2d45bc3d6ad26a/proposals/KEEP-0454-better-immutability-value-classes-MFVC.md) in the future.
+These value classes behave similarly to `data class`es, but have no identity, component functions, copy function.
+They support multiple fields, abstract/sealed classes. Read more about them in the dedicated KEEP.
```
ice-phoenix: I'd rephrase to compare full value classes with the currently supported inline ones: that they lift most of the restrictions important for the data use-cases, and people would want to change their current classes to them.

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+They are the main target to migrate the existing immutable classes to.
+
+However, the process of migrating existing classes to `value class` is a breaking change for several reasons:
+1. The mentioned restrictions regarding identity operations might break existing code.
+2. Currently existing `@JvmInline` value classes support only a single field.
```
ice-phoenix: Will we cover reasons 2 and 3? Do we want to? I'd say drop them and focus more on the actual differences and restrictions between ref-based and value-based types.

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+
+When migration to the existing inline `value class` is theoretically possible (the mentinioned problems do not apply),
+the following problems are still present:
+1. Change in the behavior of `kotlinx.serialization`, `Spring` and other frameworks which have special handling for inline `value class`es. 
+2. ABI change on JVM.
+3. Worse interoperability with Java.
+
+It stops users from migrating to inline `value class`es.
+
+Since only problem 3 can be mitigated (with `@JvmExposeBoxed`), the scope of the rest of the proposal is the migration to the full `value class`es.
```
ice-phoenix: I'm not sure we even wanna mention this, as this is completely parallel to what the current KEEP is about.

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+
+The only problem that still remains for full `value class`es is identity-related operation restrictions because:
+* ABI is kept the same (enabling support of [Project Valhalla](https://openjdk.org/projects/valhalla/) on JVM is going to change only the attributes, not descriptors or signatures).
+* The full `value class`es allow multiple fields and can be `abstract`/`sealed`.
+
+The standard library already contains classes that behave like values — they are immutable,
```
ice-phoenix: I think this should be a part of what I mentioned before -- a section talking about when people would like value-based types and not reference-based ones.

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+the following problems are still present:
+1. Change in the behavior of `kotlinx.serialization`, `Spring` and other frameworks which have special handling for inline `value class`es. 
+2. ABI change on JVM.
+3. Worse interoperability with Java.
+
+It stops users from migrating to inline `value class`es.
+
+Since only problem 3 can be mitigated (with `@JvmExposeBoxed`), the scope of the rest of the proposal is the migration to the full `value class`es.
+
+The only problem that still remains for full `value class`es is identity-related operation restrictions because:
```
ice-phoenix: Not really, we are restricting `open` classes also, right? And maybe we need smth else? E.g., `var` with a backing field?

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+the compiler recognizes `@jdk.internal.ValueBased` on JDK classes and issues warnings for
+identity-sensitive operations on their instances. `@WillBecomeValue` extends this mechanism
+to Kotlin-defined classes.
+
+## Kotlin's need for an equivalent
+
+Kotlin needs the same mechanism for its own migration path:
+
+- Kotlin needs the annotation to be available on all target platforms.
+- Library authors (including the Kotlin standard library team) want to signal that a class is
```
ice-phoenix: I think within this list you are mixing several different things: how we want the annotation to work aka what it means, how it is different from the JDK's one and what additional constraints Kotlin has, how the annotation is applied.

But the section is about: why we need an equivalent, I'd focus on that then.

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+- Kotlin needs the annotation to be available on all target platforms.
+- Library authors (including the Kotlin standard library team) want to signal that a class is
+  should be marked `value`, but it cannot be done because of the compatibility constraint, usage restrictions,
+  field number restriction, and others.
+- The Kotlin compiler and tooling can then warn at use sites where identity-sensitive operations
+  are applied to annotated instances, giving developers time to fix their code.
+- When the class is eventually migrated to `value class`, the change will be much less likely
+  to cause silent regressions.
+
+## Naming
```
ice-phoenix, line 155: Shouldn't this be the (first) part of the Proposal instead of being here?

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+
+The annotation is named **`@WillBecomeValue`**. The name was chosen over three other candidates:
+
+| Name            | Why rejected                                                                                                                         |
+|-----------------|--------------------------------------------------------------------------------------------------------------------------------------|
+| `@ValueBased`   | Ambiguous alongside the existing `value class` keyword — "value-based" vs. "value" is confusing.                                     |
```
ice-phoenix, line 161: I'd say the "why rejected" is pretty weak here, considering it precisely means "this thing is almost a value type, we just could not make it one now, that's why it's value-based and not a `value class`"

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+The annotation is named **`@WillBecomeValue`**. The name was chosen over three other candidates:
+
+| Name            | Why rejected                                                                                                                         |
+|-----------------|--------------------------------------------------------------------------------------------------------------------------------------|
+| `@ValueBased`   | Ambiguous alongside the existing `value class` keyword — "value-based" vs. "value" is confusing.                                     |
+| `@IdentityFree` | A synonym for `value`, effectively introducing a second name for the same concept.                                                   |
+| `@TreatAsValue` | Does not explain *why* the class is not already a `value class`; also does not hint that usages produce warnings rather than errors. |
+
+`@WillBecomeValue` was selected because it:
+- Makes the migration intent explicit: the class *will become* a `value class`, but cannot be one yet due to compatibility or language constraints.
```
ice-phoenix: But if this never happens, this annotation is a lie :hehe:

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+## Semantics
+
+A class annotated with `@WillBecomeValue`:
+
+1. **Must not be relied upon for identity.** Reference equality (`===`), `identityHashCode`,
+   and synchronization on instances are considered undefined behavior and may be flagged by
+   the compiler or runtime.
+2. **Must be shallow-immutable.** Mutating state through shared references
```
ice-phoenix, line 186: Will we be checking that or not?

proposals/KEEP-0465-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+- Abstract classes and sealed classes, intended as base types.
+
+It **cannot** be applied to:
+- Classes already declared as `value class`.
+- Interfaces (interfaces are common for identity and value classes).
+- `object` declarations (singletons intentionally have identity).
+- Enums (enumerations cannot become value types).
+- Open classes (value classes are not going to be supported to be `open`).
+
+## Interaction with value classes
```
ice-phoenix: That's already covered by the first item in the list above, I think

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+- `object` declarations (singletons intentionally have identity).
+- Enums (enumerations cannot become value types).
+- Open classes (value classes are not going to be supported to be `open`).
+
+## Interaction with value classes
+
+Once a class annotated with `@WillBecomeValue` is migrated to `value class`, the annotation
+should be removed. The compiler will report an error if `@WillBecomeValue` is present on a `value class`.
+
+# References
```
ice-phoenix, line 285: Maybe also add a list of things from the stdlib which we think would make sense to make value classes in the future? That would also help us to do impact estimation on KUPs.

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,195 @@
...
+   the compiler or runtime.
+2. **Must be shallow-immutable.** Mutating state through shared references
+   undermines the value semantics the annotation promises.
+3. **Is a candidate for future migration to `value class`.**
+   Library authors are expected to complete the migration to prevent libraries from keeping the annotation forever.
+
+All the checks eligible for full value class declaration itself ***(not usages)*** are also run on classes annotated with `@WillBecomeValue`.
+The only observable difference is that the compiler will issue a warning instead of an error for wrong class **usages**.
+
+## Compiler warnings
```
ice-phoenix, line 219: Maybe we also need smth for the data to value class migration story?

ice-phoenix (changes requested):

proposals/KEEP-0468-will-become-value.md
zhelenskiy: That's the practical impact for the users: how they realize they are affected.

proposals/KEEP-0468-will-become-value.md
zhelenskiy: yes

proposals/KEEP-0468-will-become-value.md
zhelenskiy: I'd name it `@AlmostValue` then. But that seems weird to me. `Based` means the used instrument for me, e.g. `Spring is a reflection-based server-side framework`. That does not mean that Spring is almost reflection, which would be a categorial mistake. You might say that the class is based on some value (which is also confusing as it just provides value semantics, not being based on anything). But then any value class should be marked with the annotation. There is no distinguishable difference between `@ValueBased` and `value` in their name.

proposals/KEEP-0468-will-become-value.md
zhelenskiy: > I'm not sure that distinguishing by synchronization is even possible.

```kotlin
val myNotYetValueClass = MyNotYetValueClass(...)
repeat(100) {
    launch {
        synchronized(myNotYetValueClass) {
            // perform some data race here
        }
    }
}
```

ice-phoenix: I believe on the Valhalla JVM this will just blow up immediately =)

proposals/KEEP-0465-will-become-value.md
zhelenskiy: > Will we cover reasons 2 and 3?

Yes. Full value classes cover them. And that's written below.

ice-phoenix: Yes, but will we cover these reasons with `@WillBecomeValue`? Are they important for the migration trajectory which this annotation targets?

To my mind, it's about ref-based to value-based types, and talking about the `inline value class` => `full value class` migration only confuses the reader.

proposals/KEEP-0468-will-become-value.md
zhelenskiy: why?

proposals/KEEP-0468-will-become-value.md
zhelenskiy: What is the current state regarding name-based destructuring and copy vars?

ice-phoenix: NBD is going through its planned migration stages (only-syntax by default in 2.5, name-mismatch by default in 2.6). Copy vars we don't know yet, maybe experimental in 2.6, why do you ask?

zhelenskiy: I think the original state of data class affects the migration to value class because they must have only name based destructing.

proposals/KEEP-0468-will-become-value.md
zhelenskiy: And we want to make it easier beforehand.

proposals/KEEP-0468-will-become-value.md
ice-phoenix: About the NBD, yes, I'm not immediately sure what's the connection with the copy vars?

zhelenskiy: When migrating data classes, we need to migrate both `dc.copy(a = b)` and `val (a, b) = dc`. Therefore, we depend on the status of the corresponding features.

ice-phoenix: The `copy` migration will need to be covered with a separate mechanism, and not with copy vars, if you want, you could just link to the corresponding part of the MFVC KEEP, which hints at several options

proposals/KEEP-0468-will-become-value.md
ice-phoenix: The name is kinda part of the design, and not of the motivation imho

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,227 @@
...
+Representing such data as `value class`es rather than ordinary reference classes brings several benefits:
+
+- It stops incidental identity operations: accidental `===`, a lock taken on a shared instance, or a cache keyed on identity can no longer silently behave differently from what the value semantics suggest.
+- It enables potential optimizations: value classes can be flattened into their fields and passed without boxing, instead of being heap-allocated and accessed through a pointer.
+- It allows smart-casts to cross module boundaries, since the absence of identity means the value cannot be mutated concurrently between the check and the use.
+- It ensures better concurrency guarantees: without shared mutable identity, instances are safe to publish and share across threads.
+- It provides support for the upcoming features [name-based destructuring](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0438-name-based-destructuring.md) and [copy vars](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0453-better-immutability-value-classes-motivation.md#mutable-value-semantics-and-copy-vars).
```
ice-phoenix: Nit: I'm not sure it's 100% correct to say these are *benefits* of value classes

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,227 @@
...
+
+The standard library already contains classes that behave like values — they are immutable,
+have no meaningful identity, and are candidates for eventually becoming `value class`. Examples
+include `Pair`, `Triple`, and various result/wrapper types.
+
+Kotlin currently supports a very limited subset of `value class`es: inline `value class`es with a single underlying field.
+Their main purpose is to create type-safe wrappers around existing types being transparent in the runtime.
+Many libraries and frameworks (`kotlinx.serialization`, `Spring`) adopted the usage and embed the underlying field, keeping safe wrapper only on the source code level.
+Such a behavior is not acceptable for general purpose value classes.
+As a result, many classes that should have been `value` cannot be marked as such and remain ordinary reference classes.
```
ice-phoenix, line 65: This part is confusing, I don't understand what idea we wanna get across :sad:

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,227 @@
...
+There are plans to support [full `value class`es](https://github.com/Kotlin/KEEP/blob/69d675e8a15f66ff6b3dace70b2d45bc3d6ad26a/proposals/KEEP-0454-better-immutability-value-classes-MFVC.md) in the future.
+Unlike the currently supported inline `value class`es, they lift exactly the restrictions that blocked the data use-cases:
+* May have multiple fields.
+* Are not embedded into the underlying single field by libraries and frameworks, e.g. `kotlinx.serialization`, `Spring`.
+* Do not change the ABI and do not spoil Java interoperability.
+
+In other respects they behave like `data class`es without identity, component functions, or a `copy` function,
+but with name-based destructuring and copy vars.
+
+They also may be `abstract`/`sealed`.
```
ice-phoenix: Nit: technically, that's true, but this fact does not immediately connect to the `@ValueBased` story or move the motivation forward, hence is it worth it to mention this?

proposals/KEEP-0467-will-become-value.md
```diff
@@ -0,0 +1,227 @@
...
+* May have multiple fields.
+* Are not embedded into the underlying single field by libraries and frameworks, e.g. `kotlinx.serialization`, `Spring`.
+* Do not change the ABI and do not spoil Java interoperability.
+
+In other respects they behave like `data class`es without identity, component functions, or a `copy` function,
+but with name-based destructuring and copy vars.
+
+They also may be `abstract`/`sealed`.
+
+This makes full `value class`es the natural target for the value-like classes that could not be expressed before, and people would want to migrate their existing classes to them. Read more about them in the dedicated KEEP.
```
ice-phoenix: Nit: link to the KEEP?

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,227 @@
...
+However, the process of migrating existing classes to `value class` is a breaking change for several reasons:
+1. The mentioned restrictions regarding identity operations might break existing code.
+2. The `open` classes cannot be migrated because of the unclear semantics of the generated `equals`/`hashCode`/`toString` methods similar to `data class`es.
+3. The `var`-based data classes cannot be migrated because of the mutability restrictions.
+
+
+The problems 2 and 3 are the valid reasons for the class not to become a `value class`.
+On the other hand, the first one can and should be mitigated.
+
+## The migration problem
```
ice-phoenix, line 82: We already stated the migration problem in the end of previous section, and say it again here; something must go. Either wrap this section into previous text, or move that to this section. Otherwise, we are saying the same thing twice twice again again.

proposals/KEEP-0467-will-become-value.md
```diff
@@ -0,0 +1,227 @@
...
+
+However, the process of migrating existing classes to `value class` is a breaking change for several reasons:
+1. The mentioned restrictions regarding identity operations might break existing code.
+2. The `open` classes cannot be migrated because of the unclear semantics of the generated `equals`/`hashCode`/`toString` methods similar to `data class`es.
+3. The `var`-based data classes cannot be migrated because of the mutability restrictions.
+
+
+The problems 2 and 3 are the valid reasons for the class not to become a `value class`.
+On the other hand, the first one can and should be mitigated.
```
ice-phoenix:
```suggestion
To avoid having an immediate breaking change on both declaration- (points 2 and 3) and use-sites (point 1) for such a migration, we would like to introduce a mitigation strategy via a dedicated migration process.
```

proposals/KEEP-0467-will-become-value.md
```diff
@@ -0,0 +1,227 @@
...
+the compiler recognizes `@jdk.internal.ValueBased` on JDK classes and issues warnings for
+identity-sensitive operations on their instances. `@WillBecomeValue` extends this mechanism
+to Kotlin-defined classes.
+
+## Kotlin's need for an equivalent
+
+Kotlin needs the same mechanism for its own migration path, and the JDK's annotation cannot serve it:
+
+- `@jdk.internal.ValueBased` is JVM-only, while Kotlin is not.
+- `@jdk.internal.ValueBased` is internal to the JDK.
```
ice-phoenix:
```suggestion
- `@jdk.internal.ValueBased` is internal to the JDK, while Kotlin needs to support a proper ecosystem-level mechanism.
```

proposals/KEEP-0467-will-become-value.md
```diff
@@ -0,0 +1,227 @@
...
+- Kotlin standard library has its own value-like classes awaiting migration.
+- Kotlin libraries authors need to start the migration beforehand.
```
ice-phoenix: Nit: aren't the first two items enough?

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,227 @@
...
+
+Kotlin needs the same mechanism for its own migration path, and the JDK's annotation cannot serve it:
+
+- `@jdk.internal.ValueBased` is JVM-only, while Kotlin is not.
+- `@jdk.internal.ValueBased` is internal to the JDK.
+- Kotlin standard library has its own value-like classes awaiting migration.
+- Kotlin libraries authors need to start the migration beforehand.
+
+## Candidate classes in the standard library
+
```
ice-phoenix, line 123:
```suggestion
To show this "reference-to-value-based" class migration is actually relevant, let's take a look at Kotlin standard library.
```

proposals/KEEP-0467-will-become-value.md
```diff
@@ -0,0 +1,227 @@
...
+
+For contrast, some standard-library types are **already** inline `value class`es and need no migration: `UByte`, `UShort`, `UInt`, `ULong`, [`kotlin.time.Duration`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.time/-duration/), and `Result`. Each of them wraps a single field, which is exactly the case the current inline `value class`es already cover.
+
+Not every candidate can carry `@WillBecomeValue` right away.
+Everything in the first two groups below can be annotated **today**:
+each is `final`, shallow-immutable, and has no `open` supertype, so it already passes the [applicability](#applicability) checks.
+The types in the last group are value-like too,
+but **cannot be annotated yet** because a `value class` may be neither `open` nor a subtype of an `open` class —
+they must first shed `open` from their own declaration or from the class they inherit.
```
ice-phoenix: In general, if you have some items you are talking about, it's best to talk about them when you talk about them. At the moment, this part 1) needs me to understand smth about 3 items I have not yet read 2) duplicates things you mention when you talk about specific items 3) if not, it's easy to move the point to the item.

Such preliminary exposition would be needed if it adds smth unique by comparing A, B and C, but that's not the case here.

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,227 @@
...
+- **Ranges and progressions.** The progression base classes `IntProgression`, `LongProgression`, `CharProgression` (and the unsigned `UIntProgression`, `ULongProgression`) are `open`, and the ranges inherit from them (`IntRange : IntProgression`, and so on). So *neither* the progressions *nor* the ranges [`IntRange`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.ranges/-int-range/), `LongRange`, `CharRange`, `UIntRange`, `ULongRange` can be marked — they would first have to be redesigned as a `sealed`/`abstract` value-class hierarchy, precisely the abstract/sealed support that full value classes add.
+- [`kotlin.io.encoding.Base64`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.io.encoding/-base64/) — an immutable configuration (`isUrlSafe`, `isMimeScheme`, `mimeLineLength`, `paddingOption`), but declared `open` (its `Default` companion object extends it), so it is blocked until it is made non-`open`.
```
ice-phoenix, line 149: > This class is not supposed to be inherited or instantiated by calling its constructor.

FUUUUUU...

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,227 @@
...
+1. **Must not be relied upon for identity.** Reference equality (`===`), `identityHashCode`,
+   and synchronization on instances are considered undefined behavior and may be flagged by
+   the compiler or runtime.
+2. **Must be shallow-immutable.** Mutating state through shared references
+   undermines the value semantics the annotation promises.
+3. **Is a candidate for future migration to `value class`.**
+   Library authors are expected to complete the migration to prevent libraries from keeping the annotation forever.
+
+All the checks eligible for full value class declaration itself ***(not usages)*** are also run on classes annotated with `@WillBecomeValue`.
+The only observable difference is that the compiler will issue a warning instead of an error for wrong class **usages**.
```
ice-phoenix, line 192: Do we plan to make diagnostics *errors* for this annotation? I would say they should (as with JEP 390) be warnings.

proposals/KEEP-0467-will-become-value.md
```diff
@@ -0,0 +1,227 @@
...
+## Applicability
+
+The annotation can be applied to:
+- Regular (reference) final classes.
+- Abstract classes and sealed classes, intended as base types.
+
+It **cannot** be applied to:
+- Classes already declared as `value class`.
+- Interfaces (interfaces are common for identity and value classes).
+- `object` declarations (singletons intentionally have identity).
```
ice-phoenix: Nit: I disagree a bit here, objects intentionally have *an instance*, whether the instance has an identity is separate from being a singleton, actually.

proposals/KEEP-0468-will-become-value.md
ice-phoenix: If we are checking that, I'm missing the part where we describe the corresponding warning or applicability rule.

proposals/KEEP-0468-will-become-value.md
```diff
@@ -0,0 +1,227 @@
...
+- Abstract classes and sealed classes, intended as base types.
+
+It **cannot** be applied to:
+- Classes already declared as `value class`.
+- Interfaces (interfaces are common for identity and value classes).
+- `object` declarations (singletons intentionally have identity).
+- Enums (enumerations cannot become value types).
+- Open classes (value classes are not going to be supported to be `open`).
+
+# References
```
ice-phoenix, line 285: Nit: maybe a section with some examples (e.g., from the stdlib) with annotated declarations, use-sites, and diagnostics?

ice-phoenix (changes requested):
Also, if you want to stick to the `@WillBecomeValue` name, you'll need to convince the stdlib team. IIUC, they very much prefer the `@ValueBased` name.

proposals/KEEP-0468-will-become-value.md
zhelenskiy: why?

proposals/KEEP-0468-will-become-value.md
zhelenskiy: Also, there is `String` that we somehow need to handle.

proposals/KEEP-0468-will-become-value.md
zhelenskiy: I mean the checks must be the same as if the class was marked `value`

zhelenskiy: I'll write more about it.

proposals/KEEP-0468-will-become-value.md
zhelenskiy: I added an example

proposals/KEEP-0468-will-become-value.md
zhelenskiy: I mean saying that they are like data classes without saying about ability to be abstract/sealed would be misleading. Do you have any ideas how to correct?

proposals/KEEP-0468-will-become-value.md
zhelenskiy: @fzhinkin

proposals/KEEP-0468-will-become-value.md
ice-phoenix:
```suggestion
Similarly to data classes, value classes can extend interfaces; additionally, they also may be `abstract` / `sealed`.
```

But I'm still not sure we need this fact here.

proposals/KEEP-0468-will-become-value.md
ice-phoenix: Because NBD is available for everybody, and `copy var`s are just a feature of value classes, it's not a benefit (there's nothing to compare against).

proposals/KEEP-0468-will-become-value.md
ice-phoenix: JDK won't make `String` a value class, what do you propose to do here?

ice-phoenix (approved):

proposals/KEEP-0468-will-become-value.md
zhelenskiy: Maybe add permanent warnings on identity operations alongside the main feature implementation.

proposals/KEEP-0468-will-become-value.md
fzhinkin: `@ValueBased` was always my favorite candidate - it reuses existing name (so no need to reinvent the name), and identifies a class as something that has value semantics, but can't be a value class yet (as Marat mentioned).

zhelenskiy: What about my counterarguments?

proposals/KEEP-0468-will-become-value.md
zhelenskiy: Wrote about it in the further improvements document

fzhinkin (approved):

serras (approved):

PR #503, opened 2026-09-03
@SignatureDiscriminator review

InversionSpaces:


proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,359 @@
...
+* **Contributors**: Marat Akhin, TBD
+* **Discussion**: TBD
+* **Status**: Internal discussion
+* **Related proposal**: [Signature discriminators in platform import and export](KEEP-xxxx-signature-discriminator-interop.md)
+* **Related YouTrack issues**: [KT-52009](https://youtrack.jetbrains.com/issue/KT-52009)
+
+## Abstract
+
+This proposal introduces `@SignatureDiscriminator`, a multiplatform mechanism for assigning an additional identity component to a Kotlin callable.
+It is intended for API evolution when a library must retain an old declaration alongside a new declaration that would otherwise be a conflicting overload or produce a declaration clash.
```
InversionSpaces:
```suggestion
It is intended for API evolution when a library must retain an old declaration alongside a new declaration that would otherwise be a conflicting overload or produce a platform declaration clash.
```

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,359 @@
...
+
+A signature discriminator participates in declaration-conflict checking, override matching, expect/actual matching, and binary linkage.
```
InversionSpaces, line 16:
```suggestion
A signature discriminator participates in overload-conflict checking, override matching, expect/actual matching, and binary linkage.
```

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,359 @@
...
+interface API {
+    @Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
+    fun value(): CharSequence
+
+    fun value(): String
+}
+```
+
+This does not compile because Kotlin does not allow overloads that differ only in return type.
+Similar problems arise when changing suspendness or other features of a Kotlin callable.
```
InversionSpaces: nit: for completeness of our and readers understanding, I believe it would be beneficial to list all the cases we are aware of.

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,359 @@
...
+
+`@SignatureDiscriminator` makes the new declaration a distinct callable for override matching and binary linkage while keeping its Kotlin source name unchanged:
```
InversionSpaces: A bit confusing that the whole section showcases how discriminator solves overload conflict, but this phrase is about other identities.

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,359 @@
...
+```kotlin
+interface API {
+    @Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
+    fun value(): CharSequence
+
+    fun value(): String
+}
+```
+
+This does not compile because Kotlin does not allow overloads that differ only in return type.
```
InversionSpaces, line 83: nit: If I abstract away and read this as if I don't know anything about such library migrations, it becomes confusing. Previous paragraph says that this is a common migration recipe, but this one says that it does not compile. So is it working or not? I think it might be better to frame what we solve here: there are already migration patterns that work, and we are filling the hole where migration involves conflicting overloads.

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,359 @@
...
+Existing options do not solve the core problem: distinguishing two declarations that are identical according to Kotlin signature rules.
+They are only partial remedies for individual symptoms.
+
+Using `@JvmName` can solve some JVM-only clashes for top-level and final declarations, but it is not generally applicable to open or abstract members, and does not work for other platforms.
+
+Suppressing `CONFLICTING_OVERLOADS` or a platform-declaration-clash diagnostic does not guarantee distinct binary identities and can produce broken artifacts.
+
+## Non-goals
+
+This proposal specifically defines `@SignatureDiscriminator` as a solution to the problem of making two declarations distinct.
```
InversionSpaces: nit: It can be understood from the whole proposal, but in isolation the phrase "making two declarations distinct" is very ambiguous. The very next sentence says they are not distinct in call-resolution sense.

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,359 @@
...
+The following rules apply:
+
+* `value` must be a non-empty string literal.
+  Constant expressions, concatenations, and references to `const val` declarations are not accepted.
+  The compiler must be able to read the discriminator before ordinary expression resolution.
+* The annotation cannot be applied through a typealias.
+* The annotation cannot be applied to a local declaration or a constructor.
+* Two equal string values denote the same discriminator.
+  The value has no additional structure and is compared exactly.
+* Changing, adding, or removing the discriminator of a published declaration is a binary-incompatible change, unless the old declaration is retained under its old identity.
```
InversionSpaces: nit: This entry feels a little bit off, as points above explain formal application rules of the annotation, but this one is about its effect. It also has other effects, but they are not listed here

proposals/KEEP-xxxx-signature-discriminator.md
InversionSpaces: On the second thought, if we include properties, then declaration-conflict is the right common term. Did I understand right?

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,359 @@
...
+On other platforms the backend encodes it in the declaration's binary identity.
+
+For example, on the JVM, the method descriptor has no field for a discriminator, so the compiler encodes it in the JVM method name.
+Conceptually, the emitted name consists of the Kotlin source name followed by a stable encoding of the discriminator, while Kotlin metadata continues to expose the original source name.
+Kotlin call sites and overrides use the encoded JVM name in bytecode.
+
+This encoding must be deterministic, collision-free, stable across recompilations and compiler versions.
+The exact encoding will be specified before stabilization.
+
+### Conflicting overloads
```
InversionSpaces, line 185: "overload signature" was not included in the list above on purpose?

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,359 @@
...
+    override fun value(): String = "new"
+}
+```
+
+As this annotation is an important part of the declaration signature, it does not propagate automatically; one needs to write it explicitly on an override of a discriminated declaration.
+
+This is intentionally verbose:
+
+* it keeps declaration identity visible at every source declaration;
+* it avoids deriving the effective signature of an override from non-local supertypes;
```
InversionSpaces: What are non-local supertypes?

InversionSpaces (commented):
What is our stance on non-migration usages of `@SignatureDiscriminator`?
The following should be allowed and not generate a warning?
```kotlin
private fun foo(a: String) {}
@SignatureDiscriminator("foo-b")
private fun foo(b: String) {}
```
I guess it could be useful for swift import possibly.
EDIT: Oh, I see, just started to read the interop KEEP.

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,532 @@
...
+interface API {
+    @Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
+    fun value(): CharSequence
+
+    @SignatureDiscriminator("string-result")
+    fun value(): String // compiled to a mangled `fun value-<c0ffee>` on the JVM
+}
+```
+
+To be usable, Java callers need a valid Java name such as `stringValue()`.
```
InversionSpaces: This reads like "Java callers need ... to be usable" which is not the right meaning I believe.

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,532 @@
...
+## Alternatives
+
+### Use the discriminator value as the export name by default
+
+This would make `@SignatureDiscriminator("string-result")` automatically behave like every platform naming annotation.
+
+It is rejected because:
+
+* a string that is good opaque identity is not necessarily a good or legal Java, JavaScript, Objective-C, Swift, or WebAssembly name;
+* changing a foreign-language name should not change Kotlin-to-Kotlin linkage;
```
InversionSpaces: But it would not? I mean, if user wants to change the export name of a declaration with already applied `SignatureDiscriminator`, they would have to use explicit `ExportName`. Changing `SignatureDiscriminator` remains binary incompatible.

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,532 @@
...
+```kotlin
+interface API {
+    @SignatureDiscriminator("string-result")
+    @JvmName("stringValue")
+    fun value(): String
+}
+
+class APIImpl : API {
+    @SignatureDiscriminator("string-result")
+    override fun value(): String = "new" // generates both a Kotlin-mangled `value-<...>` and a `stringValue` bridge on the JVM
```
InversionSpaces, line 279: I feel like alternatives, their pros and cons are not conveyed clearly for this idea of "primary exported callable". 
If an interface sets export name, inheritors generate methods with the same name anyway as bridges, why not make their primary methods have the same name by default and generate no bridges? (I guess for uniformity with the multiple inheritance case, where it is not so clear which one is the primary name, but the reader has to connect the dots)
In the same direction, why not "inherit" `useAsExportName = true` by default?
You would have to set `useAsExportName = true` through your hierarchy if you want to inherit your interfaces comfortably on the platform, otherwise you would have to override mangled names sometimes.

proposals/KEEP-xxxx-signature-discriminator.md
ice-phoenix: Thanks, that's a very good suggestion!

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,371 @@
...
+interface API {
+    fun value(): String
+}
+class APIImpl : API {
+    override fun value(): String = "new"
+}
+```
+
+Replacing the old declaration is binary incompatible: previously compiled consumers still link against the old binary signature.
+The usual migration recipe is to keep the old declaration hidden and add the new one, which usually works until it does not:
```
serras, line 72:
```suggestion
It is sometimes possible to keep the old declaration as "hidden" (meaning it appears in the binary, but does not exist for any other Kotlin purpose), but this is not a silver bullet.
```

The current text reads too much as "things you should know", but I think it's better to assume that the reader isn't aware of all intricacies of Kotlin.

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,371 @@
...
+    fun value(): String
+}
+```
+
+This does not compile because Kotlin does not allow overloads that differ only in return type.
+Similar problems arise also for the following cases:
+
+* changing a property type;
+* changing type parameters and arguments when they are erased on the platform;
+* changing suspendness of a Kotlin callable.
```
serras, line 88:
```suggestion
* adding `suspend` to or removing it from a Kotlin function.
```

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,371 @@
...
+
+`@SignatureDiscriminator` makes the new declaration a distinct callable for conflicting overloads, override matching and binary linkage while keeping its Kotlin source name unchanged:
```
serras, line 90: Although it's explained later, the sentence "conflicting overloads, override matching and binary linkage" requires a lot from the reader. May I suggest that you change this into a list of goals, and why each of them are important? I think that would also make the next section ("Why existing ... not enough") easier to digest.

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,371 @@
...
+
+* `value` must be a non-empty string literal.
+  Constant expressions, concatenations, and references to `const val` declarations are not accepted.
+  The compiler must be able to read the discriminator before ordinary expression resolution.
+* The annotation cannot be applied through a typealias.
+* The annotation cannot be applied to a local declaration or a constructor.
+* Two equal string values denote the same discriminator.
+  The value has no additional structure and is compared exactly.
+
+### The signatures of a Kotlin callable
```
serras, line 143: A table would make wonders here.

```markdown
|                   | Source name | Type parameters | Value parameter types | Value parameter names    | Return type | Other attributes | Signature discriminator 🆕 |
|-------------------|-------------|-----------------|-----------------------|--------------------------|-------------|------------------|---------------------------|
| Call resolution   | ☑️           | ☑️               | ☑️                     | If using named arguments | ❌           | ❌                | ❌                         |
| Overload          | ☑️           | ☑️               | ☑️                     | ❌                        | ❌           | ❌                | ☑️                         |
| Override matching | ☑️           | ☑️               | ☑️                     | ❌                        | ❌           | ☑️                | ☑️                         |
```

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,537 @@
...
+    @SignatureDiscriminator("string-result")
+    @JvmName("stringValue")
+    fun value(): String
+}
+```
+
+### Platform names and inheritance
+
+When a callable `C` with platform name `N_C` overrides another callable `Q` with platform name `N_Q`, it creates *two* platform entities: the `N_C`-named primary callable and the `N_Q`-named secondary bridge.
+To avoid this, one should prefer to also set the platform name via `@ExportName` to `N_Q` when inheriting from such callables.
```
serras, line 268: This is really hard to understand.

serras:
As a general comment, the fact that _Export model_ is separate from the "core proposal" feels very weird to me. In particular, I would expect the problem of open members to play a much more important role. Currently the text only states:

> As `@SignatureDiscriminator` gives us a way to disambiguate callables in complicated cases (e.g., with inheritance), we allow changing platform name via `@ExportName` for open or abstract members.

But I find it very hard to understand why this is the case.

serras:
It would be nice if the proposal also outlined which restrictions could be dropped from the language once `@SignatureDiscriminator` lands.

serras:
Am I correct that apart from `ExportName` in open/abstract members, this proposal could also allow `internal` member on interfaces if they come with a signature discriminator?

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,371 @@
...
+
+Kotlin uses different notions of a callable's signature for different purposes.
+The proposal changes some of them and intentionally leaves others unchanged.
+
+**Call-resolution signature.**
+This is used to decide which declaration a source call resolves to.
+It includes the Kotlin source name, value parameter types and type parameters, and may use value parameter names for calls with named arguments.
+It excludes the return type.
```
kunyavskiy, line 151: That's a bit bold statement:
```
package a
fun foo(): Int = TODO()
package b
fun foo(): String {}
package c
import a.*
import b.*

fun main() {
    val x: () -> Int = ::foo
    val y: () -> String = ::foo
}
```

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,371 @@
...
+**Expect/actual-matching signature.**
+For all intents and purposes, it works the same way as override-matching signature.
+
+**Binary signature.**
+This identifies a declaration in a compiled artifact.
+On klib-based platforms the discriminator is stored in the linkage signature.
+On other platforms the backend encodes it in the declaration's binary identity.
+
+For example, on the JVM, the method descriptor has no field for a discriminator, so the compiler encodes it in the JVM method name.
+Conceptually, the emitted name consists of the Kotlin source name followed by a stable encoding of the discriminator, while Kotlin metadata continues to expose the original source name.
```
kunyavskiy, line 179: So, this function would be not callable from java? Or how exactly should it looks like?

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,371 @@
...
+
+**Overload signature.**
+This is used for the conflicting overloads checking.
+It includes the Kotlin source name, value parameter types and type parameters.
+It excludes the return type and the value parameter names.
+The proposal adds the presence and exact value of `@SignatureDiscriminator` to this signature.
+
+**Override-matching signature.**
+This is used to match an `override` with inherited members.
+It includes the Kotlin source name, value parameter types and type parameters.
```
kunyavskiy, line 165: Btw, also package / containing class. Almost everywhere when you reference name.

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,371 @@
...
+
+Consequently, the intended migration from the [motivation](#motivation) section compiles:
+
+```kotlin
+interface API {
+    @Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
+    fun value(): CharSequence
+
+    @SignatureDiscriminator("string-return-type")
+    fun value(): String
```
kunyavskiy, line 199: Do we in fact need to provide default value here, calling the old function? It's also broken with super.value() calls btw. 

I still not getting the whole migration story for open methods. Just doing this wouldn't work, because it would break clients, that has new interface but old implementations. 

While this feature definetely opens the migration path, it would be nice to explicitly show the whole migration story. Otherwise we can end-up with making a feature, that's not really usable, because of other problems nearby.

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,371 @@
...
+interface API {
+    @Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
+    var item: CharSequence
+
+    @SignatureDiscriminator("string-item")
+    var item: String
+}
+```
+
+but is applied to all derived entities: property itself, backing field, getter and setter, etc.
```
kunyavskiy, line 320: But we can apply it specifically to the getter/setter (but not field btw). What does it mean?

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,537 @@
...
+* **Depends on**: [Signature discriminators for Kotlin callables](KEEP-xxxx-signature-discriminator.md)
+* **Related YouTrack issues**: [KT-31420](https://youtrack.jetbrains.com/issue/KT-31420), [KT-20068](https://youtrack.jetbrains.com/issue/KT-20068), [KT-61323](https://youtrack.jetbrains.com/issue/KT-61323)
+
+## Abstract
+
+This proposal extends signature discriminators to platform import and export.
+The core proposal gives a Kotlin callable an additional declaration identity without treating the discriminator as a foreign-language name.
+This proposal defines the convenience mapping layer between that Kotlin identity and Java, JavaScript, Objective-C, Swift, and WebAssembly identities.
+
+For export, existing platform naming annotations (`@JvmName`, `@ObjCName`, etc.) may be now applied to all members, including open and abstract ones.
```
kunyavskiy, line 17: Is it true? I think we don't allow JvmName on open?

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,537 @@
...
+The mental model is that such a declaration has the corresponding platform-specific name annotation applied with the discriminator value as the selected name.
+
+For import, a platform importer does not by default use signature discriminators, and the signatures work according to the default Kotlin rules.
+The convenience layer here is that we can begin to synthesize a discriminator when the platform identity contains information that ordinary Kotlin signatures lose.
+The main concrete application is Objective-C selectors.
+
+For example, these Objective-C selectors could be imported with distinct synthesized discriminators even though their Kotlin parameter types are identical:
+
+```kotlin
+@SignatureDiscriminator("objc:tableView:titleForHeaderInSection:") // synthesized
```
kunyavskiy, line 40: It's a bit confusing in example, that tableView is both function and first parameter name. not really clear which of two is used in the string. 
And as far as I remember it replaces name, not adds to it, so I'm not sure if we can compatibly switch to new behavior.

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,537 @@
...
+interface API {
+    @SignatureDiscriminator("string-result")
+    @JvmName("stringValue")
+    fun value(): String
+}
+```
+
+### Platform names and inheritance
+
+When a callable `C` with platform name `N_C` overrides another callable `Q` with platform name `N_Q`, it creates *two* platform entities: the `N_C`-named primary callable and the `N_Q`-named secondary bridge.
```
kunyavskiy, line 267: Do we actually have a use-case for that? Let's maybe just prohibit it on the first version of the feature?

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,537 @@
...
+
+```kotlin
+@SignatureDiscriminator("stringValue", useAsExportName = true)
+fun value(index: Int): String = "value"
+// Objective-C selector: stringValueIndex:
+```
+
+The Swift-facing name remains derived from the Kotlin source name, so the same declaration is presented as `value(index:)` through `NS_SWIFT_NAME`.
+This keeps the discriminator responsible for the Objective-C runtime identity while retaining an idiomatic Swift source API.
+An explicit `@ObjCName(name = ...)` overrides the selector base, and `@ObjCName(swiftName = ...)` independently overrides the Swift-facing name.
```
kunyavskiy, line 416: Please check this part explicitly with Swift export people on that. It looks quite inconsistent with eveything else.

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,537 @@
...
+
+## Compatibility and migration
+
+### `@ObjCSignatureOverride`
+
+`@ObjCSignatureOverride` has shipped in Kotlin 2.0 with source retention and deliberately narrow applicability.
+This proposal does not deprecate it.
+Existing annotated overrides continue to compile and acquire the selector-derived identity described above.
+
+The compiler may offer a quick-fix to replace it with an explicit inherited discriminator only in rare ambiguity cases; the `@ObjCSignatureOverride` marker annotation remains the preferred spelling otherwise.
```
kunyavskiy, line 506: The IDE?

proposals/KEEP-xxxx-signature-discriminator-interop.md
ice-phoenix: That's the state with this proposal in place, one of its goals is to allow `@JvmName` on open members.

proposals/KEEP-xxxx-signature-discriminator-interop.md
ice-phoenix: Good point, I'll update to the `tableView(of:..., title...)` shape to make the generated selectors unambiguous (aka the first param is ignored).

> And as far as I remember it replaces name, not adds to it, so I'm not sure if we can compatibly switch to new behavior.

Didn't fully get that part, what replaces name?

proposals/KEEP-xxxx-signature-discriminator-interop.md
ice-phoenix: That makes sense. So the rule is: if you unambiguously inherit only one platform name, you must repeat it.

proposals/KEEP-xxxx-signature-discriminator-interop.md
ice-phoenix: I'll try to rephrase that to be more approachable. Thanks!

proposals/KEEP-xxxx-signature-discriminator-interop.md
glukianets: The actual selector of the method declared `- (NSString *)tableView:(UITableView *)tv titleForHeaderInSection:(NSUInteger)sn;` would be `tableView:titleForHeaderInSection:`. The discriminator suggests a three-argument function with the leading selector part `objc:`. That's confusing.  Or at least I am confused.

proposals/KEEP-xxxx-signature-discriminator.md
ice-phoenix: I would argue whether callable reference access is a proper Kotlin callable access, it is **very** special. To avoid the potential confusion, I'm going to refine the definition of what the considered cases are.

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,537 @@
...
+Parameter labels continue to follow the existing exporter rules and parameter-level `@ObjCName` annotations:
+
+```kotlin
+@SignatureDiscriminator("stringValue", useAsExportName = true)
+fun value(index: Int): String = "value"
+// Objective-C selector: stringValueIndex:
+```
+
+The Swift-facing name remains derived from the Kotlin source name, so the same declaration is presented as `value(index:)` through `NS_SWIFT_NAME`.
+This keeps the discriminator responsible for the Objective-C runtime identity while retaining an idiomatic Swift source API.
```
glukianets: Why wouldn't doing the same thing for swift export be idiomatic? I.e. exporting as `stringValue(index:)`

proposals/KEEP-xxxx-signature-discriminator.md
ice-phoenix: Yes, if you want the function which uses a Kotlin-specific discriminator to be nicely callable from Java, you'd need to use the proper export tooling (or the `useAsExportName` shortcut if it stays in the final design)

proposals/KEEP-xxxx-signature-discriminator.md
ice-phoenix: I didn't really intend for `@SignatureDiscriminator` to be applicable for getters/setters separately, considering how tightly coupled our resolution is for properties (that's why it does not have these targets as available). Do you see a use-case for that? Or am I simply missing this restriction being explicitly stated?

ice-phoenix:
TL;DR version of the update

- Added a deeper discussion around migration stories for different cases (all implementations are under our control vs there are some external implementations).
- Fixed the Objective-C-selector-to-Kotlin-signature description. Also explained why unfortunately special treatment of `@ObjCSignatureOverride` would still be needed even with `@SignatureDiscriminator`, as Native views and links ObjC functions in a very special way.
- Wording fixes here and there.

ice-phoenix:
> Am I correct that apart from `ExportName` in open/abstract members, this proposal could also allow internal member on interfaces if they come with a signature discriminator?

@serras not really, IIRC the main problem with internal members on interfaces is that they are weird. A public interface with an abstract internal member cannot be implemented in another module, it can only be used there. Which makes it a strange flavor of `sealed-via-internal interface`, which is more flexible that regular sealed interfaces, and that's strange.

Signature discriminators do nothing here.

serras:
@ice-phoenix How to implement them abstract internal members is part of the problem. But I think we also do some mangling of non-abstract members, so that they don't clash with potential future implementors outside that module. This is the part where I though signature discriminator could somehow help.

proposals/KEEP-xxxx-signature-discriminator.md
```diff
@@ -0,0 +1,715 @@
...
+
+IDE support should include completion and quick-fixes that copy the inherited discriminator into generated override and actual stubs.
+
+## Alternatives
+
+### Unconditionally use the discriminator as the platform name
+
+The discriminator value could become the JVM name, JavaScript export name, Objective-C selector base, and WebAssembly export name.
+
+> Note: in essence, this is the previously proposed `@BinarySignatureName` design.
```
SvyatoslavScherbina: Is it?

The previously proposed [design](https://github.com/Kotlin/KEEP/blob/binary-signature/proposals/multiplatform/binary-signature.md) that I know of says:

> The platform interoperability uses different mechanisms and is not affected by the specified signature name. The exported name of declaration for JavaScript, C/C++, Objective-C, and Swift is still derived from the source name of declaration and could be customized, if needed, by separate platform-specific annotations.

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,563 @@
...
+* **Depends on**: [Signature discriminators for Kotlin callables](KEEP-xxxx-signature-discriminator.md)
+* **Related YouTrack issues**: [KT-31420](https://youtrack.jetbrains.com/issue/KT-31420), [KT-20068](https://youtrack.jetbrains.com/issue/KT-20068), [KT-61323](https://youtrack.jetbrains.com/issue/KT-61323)
+
+## Abstract
+
+This proposal extends signature discriminators to platform import and export.
+The core proposal gives a Kotlin callable an additional declaration identity without treating the discriminator as a foreign-language name.
+This proposal defines the convenience mapping layer between that Kotlin identity and Java, JavaScript, Objective-C, Swift, and WebAssembly identities.
+
+For export, existing platform naming annotations (`@JvmName`, `@ObjCName`, etc.) may be now applied to all members, including open and abstract ones.
```
SvyatoslavScherbina: TODO: how is this achievable?

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,563 @@
...
+The applied platform name does not participate in Kotlin override or expect/actual matching; the signature discriminator continues to do that job.
+The convenience layer here is that the discriminator value can become the platform name, if the user so chooses.
```
SvyatoslavScherbina: Seems too limited.

proposals/KEEP-xxxx-signature-discriminator-interop.md
SvyatoslavScherbina: > It's a bit confusing in example, that tableView is both function and first parameter name. 

In fact, the real platform libraries have exactly this: `tableView` functions with `tableView` as the first parameter name, e.g.
```text
public open fun tableView(tableView: platform/UIKit/UITableView, titleForHeaderInSection: kotlin/Long /* = platform/darwin/NSInteger^ */): kotlin/String?
```

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,563 @@
...
+}
+
+class APIImpl : API {
+    @SignatureDiscriminator("string-result")
+    @JvmName("stringValue")
+    override fun value(): String = "new" // repeats the inherited platform name and generates only `stringValue` on the JVM
+}
+```
+
+#### Why not implicitly inherit the platform name?
```
SvyatoslavScherbina: I think `@ObjCName` is inherited implicitly at the moment and can't even be used on an `override`.

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,563 @@
...
+It must satisfy the existing `@JsName` identifier restrictions.
+
+This option does not imply `@JsExport`; it only selects the name if JavaScript code is generated or exported for the declaration.
+An explicit `@JsName` takes precedence.
+Klib linkage continues to use the discriminator as Kotlin identity independently of the JavaScript name.
+
+With `useAsExportName = false`, no implicit `@JsName` is applied.
+The core proposal scheme is used then.
+
+#### Kotlin/Native: Objective-C and Swift export
```
SvyatoslavScherbina: What about C Export with its [@CName](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.native/-c-name/)?

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,563 @@
...
+For Objective-C export, the discriminator value becomes the selector base.
+Parameter labels continue to follow the existing exporter rules and parameter-level `@ObjCName` annotations:
+
+```kotlin
+@SignatureDiscriminator("stringValue", useAsExportName = true)
+fun value(index: Int): String = "value"
+// Objective-C selector: stringValueIndex:
+```
+
+The Swift-facing name remains derived from the Kotlin source name, so the same declaration is presented as `value(index:)` through `NS_SWIFT_NAME`.
```
SvyatoslavScherbina: Why?

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,563 @@
...
+
+#### Kotlin/Wasm
+
+For all intents and purposes, Kotlin/Wasm works the same way as [Kotlin/JS](#kotlinjs).
+
+## Import model
+
+### Synthesized discriminators
+
+A platform import may attach a synthesized discriminator when two native declarations would otherwise merge to the same Kotlin declaration signature.
```
SvyatoslavScherbina: Why not always?

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,563 @@
...
+* preserved as if `@SignatureDiscriminator` was applied to the corresponding declaration; and
+* namespaced so it cannot accidentally equal a user-authored discriminator.
+
+Ordinary imported declarations that are already distinguishable by Kotlin rules do not acquire a separate discriminator value merely because they came from another language.
+However, if a platform import decides to also generate a discriminator for such cases, it could do so.
+
+### Objective-C import
+
+This is most useful for Objective-C import, where parameter names (selector) are an important part of a callable signature.
+The importer records a selector-derived discriminator for imported functions that require one.
```
SvyatoslavScherbina: > that require one.

What does that mean?

proposals/KEEP-xxxx-signature-discriminator-interop.md
```diff
@@ -0,0 +1,563 @@
...
+fun tableView(of: UITableView, titleForFooterInSection: NSInteger): String?
+```
+
+These declarations are handled under the general discriminator-aware rules.
+
+### Overriding imported Objective-C members
+
+Existing source compatibility is preserved as follows.
+
+When an override has exactly one compatible imported Objective-C member, the compiler adopts the inherited selector-derived discriminator implicitly.
```
SvyatoslavScherbina: > compatible

Important to note here is that this compatibility doesn't follow Kotlin rules. Instead, it always checks parameter names, even when there is no clash.

